PR #17, opened 2021-10-19
Add clients library

### LINKED ISSUE #5: Add client library shared module

voltrevo:
Resolves #5.

Creates npm module [bls-wallet-clients](https://www.npmjs.com/package/bls-wallet-clients). More details in the added [README.md](https://github.com/jzaki/bls-wallet/blob/bw-5-clients-library/clients/README.md).


jacque006 (approved):
One thing we may want to consider for the future is using https://github.com/dethcrypto/TypeChain instead of manually generating and committing contract abi's as TS files.

Edit:

Thinking about this some more, we would need to export those typechain/abi defs for consumption by this and others (aggregator, maybe extension?). This likely could just be an artifact(s) that the contracts dir creates on merges/releases via npm or deno package.

jzaki:
Agreed @jacque006 , I thought I had a ticket for it, will create one now.

voltrevo:
> One thing we may want to consider for the future is using https://github.com/dethcrypto/TypeChain instead of manually generating and committing contract abi's as TS files.

Yeah, I looked into this briefly a while ago. I was trying to make it work in deno though and it didn't seem worth it. Worth trying again.

jzaki:
Will do from hardhat #23

PR #26, opened 2021-10-21
Monorepo setup

### LINKED ISSUE #4: Setup Monorepo, add README, scripting
- Add setup README with setup instructions for monorepo
- Add scripts to aid in initial setup for first time devs

jacque006:
## What is this PR doing?
This PR adds a streamlined setup process and documentation for the monorepo of `bls-wallet` components. It does this by:
- Adding a `setup.ts` script for initial repo setup.
- Consolidating contract addresses and deployment metadata into a `network.json` file, which can be loaded/parsed with `NetworkConfig` type and functions in `clients`.

- Clone a fresh copy of the repo and checkout this branch OR (WARNING this will clear all untracked files) checkout this branch and run `git clean -xdf`.
- Follow the setup instructions in the top level `README.md`
- Ensure tests run and pass for `./aggregator`.
- Ensure tests run and pass for `./contracts`.
- With contracts deployed, aggregator running, and extension added to a browser, ensure a new BLS wallet can be created and a txn sent.

Resolves https://github.com/jzaki/bls-wallet/issues/4

## Checklist

## Guidelines

### TODO
- [x] Move contract addresses and other deployments to network JSON file (if agreed that it is a good idea)
- [x] `docker-compose` file ~to easily start all services~
- [x] Setup script
~- [ ] Replace remaining usages of `VERIFICATION_GATEWAY_ADDRESS` in `./extension`~ Ran into some config issues getting webpack to properly read in deployed contracts config at build time, will need to revisit.
- [x] Complete main `README`s
- [x] Update sub `README`s
- [x] Add local depoy setup and documentation.


aggregator/README.md
```diff
@@ -31,6 +31,14 @@ commands.
 
 ### PostgreSQL
 
+#### With docker-compose
+
+```sh
+docker-composer up -d postgres
+```
```
jacque006, line 39: TODO Move this up a level

contracts/.env.example
```diff
@@ -1,7 +1,7 @@
 PRIVATE_KEY_AGG=0000000000000000000000000000000000000000000000000000000000000a99
-PRIVATE_KEY_AGG_OKOV=
-PRIVATE_KEY_AGG_RINKARBY=
-PRIVATE_KEY_AGG_ARB1=
+PRIVATE_KEY_AGG_OKOV=0000000000000000000000000000000000000000000000000000000000000001
+PRIVATE_KEY_AGG_RINKARBY=0000000000000000000000000000000000000000000000000000000000000001
+PRIVATE_KEY_AGG_ARB1=0000000000000000000000000000000000000000000000000000000000000001
```
jacque006, line 23: This was preventing contracts repo's tests from running with a fresh install due to hardhat env imports, added dummy values

extension/webpack.config.js
```diff
@@ -67,7 +67,7 @@ module.exports = {
     contentScript: path.join(sourcePath, 'ContentScript', 'index.ts'),
     pageContentScript: path.join(sourcePath, 'PageContentScript', 'index.ts'),
     popup: path.join(sourcePath, 'Popup', 'index.tsx'),
-    createTransaction: path.join(sourcePath, 'createTransaction', 'index.tsx'),
+    createTransaction: path.join(sourcePath, 'CreateTransaction', 'index.tsx'),
```
jacque006: TODO Is this fix still needed?

contracts/networks/rinkarby.json
```diff
@@ -0,0 +1,17 @@
...
+        "rewardToken": "0x8CD755F93E56C90E3d5803c036771CFCc12B7c53"
+    },
+    "auxiliary": {
+        "chainid": 421611,
+        "domain": "TODO keccak256(abi.encodePacked(uint32(0xfeedbee5)))",
+        "genesisEth1Block": 123456,
+        "deployedBy": "",
+        "version": "c1d1f197ba9b68284cbc4ec145e286a4e84c0eaf"
+    }
+}
```
jacque006: @jzaki @voltrevo I would like to move the contract addresses and other deployment metadata out of the per directory env files and into a file modeled after `genesis.json` deployment metadata file we used in Hubble (https://github.com/thehubbleproject/hubble-contracts/blob/master/docs/setup/LOCAL_DEVELOPMENT.md#deploy).

The number one issue me and @kautukkundan ran into while getting setup was configuring these correctly across the aggregator and extension. Imagine others will as well. This has the advantage of having all the contract and onchain deployment info in one spot, at the cost of having to pass in a path to this config file when running the aggregator or extension. I think this change is worth making now, but open to other considerations/ideas.

extension/webpack.config.js
voltrevo: Oh yeah I did that in my branch too

voltrevo: Doesn't show up on macOS because of the case insensitive filesystem thing

contracts/networks/rinkarby.json
voltrevo: Sounds good to me. Definitely need some kind of combined solution, and I like moving to a json file too.

aggregator/README.md
```diff
@@ -95,7 +103,7 @@ Can be run locally or hosted.
 
 ## Testing
 
-- launch optimism
+- launch arbitrum
```
voltrevo: Worth considering just using geth for development rather than arbitrum, since the differences shouldn't affect us. Nice to have a local arbitrum for the added accuracy if it doesn't add too much complexity though.

aggregator/src/chain/BlsWallet.ts
```diff
@@ -187,9 +191,11 @@ export default class BlsWallet {
   static async Nonce(
     publicKey: string,
     signerOrProvider: SignerOrProvider,
+    verificationGatewayAddress: string,
   ): Promise<BigNumber> {
     const verificationGateway = await this.#VerificationGateway(
       signerOrProvider,
+      verificationGatewayAddress
```
voltrevo: Not too concerned about the missing trailing comma here, but it suggests to me that you don't have the deno extension configured. Let me know if you'd like some help with that.

aggregator/test/helpers/Fixture.ts
```diff
@@ -89,6 +89,7 @@ export default class Fixture {
     public chainId: number,
     public walletService: WalletService,
   ) {
+    // TODO How do we want to read in these addresses from a file? What defualt should we use?
```
voltrevo: > What defualt should we use?

My preference is for no default. Startup should fail if the config is missing a required address.

extension/webpack.config.js
jacque006: Yeah, imagine just me and anyone else on Ubuntu/Linux and likely Windows

voltrevo (commented):
(Inline comments added)

contracts/networks/rinkarby.json
jacque006: One other thing I should mention, I think we should do this for contracts but leave separate env configs for per app/service configs such as port, rpc url, etc. This is mainly just to make it easier to point to on-chain deps.

aggregator/README.md
jacque006: I will take a look into local geth as well then. We may even be able to just use `hardhat node`, but think there may be some issues with the gas estimates in the BLS precompile contract.

aggregator/src/chain/BlsWallet.ts
jacque006: Good catch, was waiting till I moved a few more things over before running the `permerge` script to check. Will look into deno setup and reach out if I find something I can't solve.

aggregator/.env.example
```diff
@@ -1,21 +1,15 @@
-RPC_URL=http://localhost:8545
+RPC_URL=https://rinkeby.arbitrum.io/rpc
```
jacque006: I switched all of the `.env.example`'s to Rinkarby as I found this was the easiest/fastest way to get setup. We can change back if desired.

aggregator/README.md
jacque006: After some investigation into geth setup, ran into some issues with contract deploys mainly around address of PreCompileEstimator. @jzaki Are there are any scripts for deploying that or should I wait on the `CREATE2` work? Looks like it is a bit more fleshed out on that branch.

extension/webpack.config.js
```diff
@@ -187,7 +187,11 @@ module.exports = {
     // plugin to enable browser reloading in development mode
     extensionReloaderPlugin,
     new DotenvWebpackPlugin({
-      path: `./.env.${requireEnv(process.env.ENV)}`,
+      path: `./.env${
+        process.env.ENV === undefined ?
+        '' :
+        `.${process.env.ENV}`
+      }`,
```
jacque006, line 201: Swapped this to `.env` to keep consistent with other components.

aggregator/src/helpers/networkConfig.ts
```diff
@@ -0,0 +1,9 @@
+// TODO (merge-ok) Replace this with client package import
+import { NetworkConfig, getConfig } from "../../../clients/src/NetworkConfig.ts";
```
jacque006: This and similar TODOs can be done once a new version of the client package is published.

We can also leave this as is and do local linking do keep things simple.

contracts/.env.example
```diff
@@ -1,42 +1,12 @@
...
-OKOVAN_BLS_LIBRARY
-OKOVAN_VERIFICATION_GATEWAY_ADDRESS=
-OKOVAN_BLS_EXPANDER_ADDRESS=
-OKOVAN_ERC20_CONTRACT_ADDRESS=
-
-OETH_PRECOMPILE_COST_ESTIMATOR_ADDRESS=
-OETH_BLS_LIBRARY
-OETH_VERIFICATION_GATEWAY_ADDRESS=
-OETH_BLS_EXPANDER_ADDRESS=
-OETH_ERC20_CONTRACT_ADDRESS=
```
jacque006: These were moved to `./contracts/networks/local.json` and `rinkarby.json`

contracts/networks/local.json
```diff
@@ -0,0 +1,18 @@
...
+        "blsExpander": "0x5FC8d32690cc91D4c39d9d3abcBD16989F875707",
+        "testToken": "",
+        "rewardToken": ""
+    },
+    "auxiliary": {
+        "chainid": 1337,
+        "domain": "0x0054159611832e24cdd64c6a133e71d373c5f8553dde6c762e6bffe707ad83cc",
+        "genesisBlock": 0,
+        "deployedBy": "TODO fill in",
+        "version": "c1d1f197ba9b68284cbc4ec145e286a4e84c0eaf"
```
jacque006: When the local deploy script is ironed out when can have these all be generated and treat this like `.env` files.

README.md
```diff
@@ -0,0 +1,88 @@
...
+
+TS/JS Client libraries for web apps and services.
+
+### Contracts
+
+`bls-wallet` Solidity contracts.
+
+### Extension
+
+Quill browser extension used to manager BLS Wallets and sign transactions.
```
voltrevo:
```suggestion
Quill browser extension used to manage BLS Wallets and sign transactions.
```

README.md
```diff
@@ -0,0 +1,88 @@
...
+PRIVATE_KEY_AGG=PK0
+PRIVATE_KEY_ADMIN=PK1
+```
+
+From `./contracts/networks/rinkarby.json` (JSON key path):
+
+`./extension/.env`
+```
+VERIFICATION_GATEWAY_ADDRESS=addresses.verificationGateway
+```
```
voltrevo, line 115: I feel this section could use some more clarity. It sounds like you're just saying the user needs to find appropriate values for those three fields? Are those the only ones?

README.md
```diff
@@ -0,0 +1,88 @@
...
+cd ./aggregator
+./programs/aggregator.ts
+```
+
+In a seperate terminal/shell instance
+```sh
+cd ./extension
+yarn run dev:chrome # or dev:firefox, dev:opera
+```
+Then install `./extension/{browser}/manifest.json` in your browser of choice.
```
voltrevo: Let's add some more detail / link to guide.

E.g. in chrome, you need to drag the `./extension/chrome` *folder* into the extensions page, giving it the manifest.json file doesn't work.

aggregator/src/helpers/dotEnvPath.ts
```diff
@@ -3,7 +3,8 @@ import { exists, parseArgs } from "../../deps.ts";
 const args = parseArgs(Deno.args);
 
 export const envName = args.env;
-const dotEnvPath = envName ? `.env.${envName}` : ".env";
+const envNameExists = envName && envName !== "undefined";
```
voltrevo: How does the `"undefined"` case occur?

README.md
jacque006: From what I experienced of the rinkarby setup yes. Defaulting the examples to rinkarby settings and consolidating contract addresses into one file took care of most things. This might just be worth scripting away and letting a dev choose between networks and just string substituting them in.

aggregator/src/helpers/dotEnvPath.ts
jacque006: When I did not set an `--env` arg, and a string of > 0 length (`"undefined"`) is truthy. Could be a difference in our deno versions?

```
deno 1.15.1 (release, x86_64-unknown-linux-gnu)
v8 9.5.172.19
typescript 4.4.2
```

aggregator/src/helpers/networkConfig.ts
```diff
@@ -0,0 +1,9 @@
...
+import * as env from "../env.ts";
+
+export type { NetworkConfig };
+
+export default function getNetworkConfig(): Promise<NetworkConfig> {
+    return getConfig(env.NETWORK_CONFIG_PATH, Deno.readTextFile);   
+}
```
voltrevo: I don't think it's ok to reach into `../../../clients`, even temporarily. It's confusing and lowers the bar for others to add more imports like these, which would create a mess that might never be untangled.

Especially since it isn't providing much - the `NetworkConfig` type is simply cast onto the JSON file without checking, and the only fields the aggregator is actually interested in are the addresses for `VerificationGateway` and the reward token.

Given that, how about this instead?

```ts
import * as env from "../env.ts";

const uncheckedNetworkConfig = JSON.parse(
  await Deno.readTextFile(env.NETWORK_CONFIG_PATH),
);

// Exporting allows for requireAddress('testToken') in tests
export function requireAddress(name: string): string {
  const value = uncheckedNetworkConfig.addresses[name];

  if (typeof value !== "string") {
    throw new Error(`Invalid ${name} address in network config`);
  }

  return value;
}

export const verificationGatewayAddress = requireAddress("verificationGateway");
export const rewardTokenAddress = requireAddress("rewardToken");
```

voltrevo: Also, although this is moot if you decide to accept the suggestion above, I feel rather strongly that if a module has a default export, then the filename of that module should match that name exactly. In this case I'd probably resolve that by making `getNetworkConfig` a named export only, although renaming the file to `getNetworkConfig.ts` would also avoid breaking the rule.

While I'm at it 😅, there's another relevant convention I'd like to raise here. I haven't seen it much from other developers, but I believe that when a function returns a variable that is naturally named `someValue`, that function should be called `SomeValue`. This avoids the proliferation of naming prefixes like `getX`, `makeX`, `generateX`, etc., although it could also just suffer from the [one-universal-standard](https://xkcd.com/927/) problem. What are your thoughts? If pascal case functions aren't preferred, we should probably work towards removing the existing uses of it.

voltrevo: Also, incorrect indentation size in this file.

clients/src/NetworkConfig.ts
```diff
@@ -0,0 +1,54 @@
+/**
```
voltrevo: Needs indentation fix (2 spaces instead of 4).

clients/src/NetworkConfig.ts
```diff
@@ -0,0 +1,54 @@
...
+type ReadFileFunc = (filePath: string) => Promise<string>;
+
+/**
+ * Rerieves Network config
+ * 
+ * @param networkConfigPath Path to config
+ * @param readFileFunc Callback to retrieve the config. This could be via fetch, fs.readFile, etc.
+ */
+export async function getConfig(networkConfigPath: string, readFileFunc: ReadFileFunc): Promise<NetworkConfig> {
+    return JSON.parse(await readFileFunc(networkConfigPath));
```
voltrevo: I'm against assigning types optimistically to functions that return `any`. If the json file doesn't match the type, it can break in confusing ways downstream, especially if there's type logic going on.

Have you ever used `io-ts`? That would be my go-to solution here.

contracts/scripts/deploy_all.ts
```diff
@@ -0,0 +1,79 @@
...
+import util from "util";
+import { exec } from "child_process";
+import { writeFile } from "fs";
+import { ethers } from "hardhat";
+import { NetworkConfig } from "../../clients/src";
+import deployAndRunPrecompileCostEstimator from "../shared/helpers/deployAndRunPrecompileCostEstimator";
+import Fixture, { DOMAIN_HEX } from "../shared/helpers/Fixture";
+
+const execAsync = util.promisify(exec);
+const writeFileAsync = util.promisify(writeFile);
```
voltrevo: Nitpick: I'd prefer the callback versions get a -Cb prefix and the promisified versions get the unmodified name.

Also for `fs` you can just import from `fs/promises` which is cleaner and the simpler underlying code may produce less weirdness (e.g. error messages).

contracts/scripts/deploy_all.ts
```diff
@@ -0,0 +1,79 @@
+require("dotenv").config();
```
voltrevo: File needs indentation fix

voltrevo: Maybe add `.vscode/settings.json` with:

```json
{
  "editor.tabSize": 2,
  "editor.insertSpaces": true
}
```

contracts/shared/helpers/networkConfig.ts
```diff
@@ -0,0 +1,14 @@
+import { readFile } from "fs";
```
voltrevo: Indentation fix for this file.

setup.ts
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env -S deno run --unstable --allow-run --allow-read --allow-write
```
voltrevo, line 1: Indentation

setup.ts
```diff
@@ -0,0 +1,41 @@
...
+console.log(`setting up components (${components.join(", ")})...`);
+await Promise.all(components.map(async ({ name, skipYarn }) => {
+    if (!skipYarn) {
+        console.log(`yarn installing in ${name}...`);
+        await runYarn(name);
+    }
+
+    const envFilePath = `./${name}/.env`;
+    const envExampleFilePath = `${envFilePath}.example`;
+    console.log(`copying ${envExampleFilePath} to ${envFilePath}...`);
```
voltrevo, line 46: Maybe move this inside the `if`, otherwise if you had any existing `.env` files it would suggest they had been overwritten.

README.md
voltrevo: Yeah hmm. Once we get something semi-final I'd like to spin up a vm and have a go at a from-scratch setup to get a clearer picture.

aggregator/src/helpers/dotEnvPath.ts
voltrevo: Hmm nah don't think it's a version thing. Just did some experimenting with 1.15.3 and the only way I can get `args.env` to be the *string* `"undefined"` is by running:

```sh
deno run testEnv.ts --env undefined
```

Which, I mean, if you do that, then I suppose resolving `.env.undefined` is what you explicitly asked for.

```ts
// testEnv.ts

import { parse } from "https://deno.land/std@0.103.0/flags/mod.ts";

const args = parse(Deno.args);

console.log(args.env);
```

voltrevo (changes requested):
Good stuff!

Finished going through here, I think the only things with significant uncertainty are:
1. Reaching into other projects (i.e. things like `import "../../../clients";`)
2. Using `JSON.parse` for `networkConfig` and not type checking it

I suggested looking at `io-ts` for (2)... watch out for deno incompatibility with that, I have a feeling I tried it before 😕. Let's chat about it if you don't see an easy way forward.

aggregator/src/helpers/dotEnvPath.ts
jacque006: Confirmed, don't know how I was seeing that issue. Reverting.

aggregator/src/helpers/networkConfig.ts
jacque006: I used the clients import to get things working, and am fine switching it to the npm import. That may be worth splitting off the config work into a separate PR that is merged before this one. The idea behind having the common config is that anyone who wants to write an app/service can just point at a file/url and get all of the needed deployed contract addresses and metadata. They just need to provide the the i/o path to the file. This will be more info than is needed than by the aggregator or extension, but may be useful in other use cases.

Makes sense to validate the fields to ensure everything is provided correctly, I can do that right after it is parsed.

~As far as naming goes, I generally like to have my types (structs) separate from the functions that create them. Can you provide an example of what the `NetworkConfig` func and type would look like in your naming scheme?~ Switched to `getNetworkConfig.ts` to match similar files.

aggregator/src/helpers/networkConfig.ts
jacque006: Moved these to `bls-wallet-clients` imports, can just use `yarn link` in other spots. Unsure what to do for deno, may need to `-alpha` tag npm releases to test out?

jacque006:
Re:

1. Good call on not importing directly from that package, forgot that `yarn link` is a thing. Moved to `bls-wallet-clients` imports and added a section to top level README. Could make more sense in `./clients` but seemed better aligned top level.
2. I was successfully able to bring in and use `io-ts` to validate the config, and it worked in Deno! I was not able to to test the full `esm.sh` linkage path, we likely need to publish an early version of `bls-wallet-clients` (`@x.y.z-alpha`) to ensure it works.

Also and unfortunately, when `esm.sh` loads in `io-ts`, it also does not appear to cache its peer dep `fp-ts`:
Screenshot from 2021-11-05 13-59-38

~Still sorting through indentation issues on my end.~ Added top level `.vscode/settings.json`. Didn't help in some cases so just manually adjusted.

setup.ts
```diff
@@ -0,0 +1,52 @@
...
+    console.log(`yarn installing in ${name}...`);
+    await runYarn(name);
+  }
+
+  const envFilePath = `./${name}/.env`;
+  const envExampleFilePath = `${envFilePath}.example`;
+
+  const [envExists, envExampleExists] = await Promise.all([
+    exists(envFilePath),
+    sexists(envExampleFilePath)
```
voltrevo:
```suggestion
    exists(envExampleFilePath)
```

aggregator/src/helpers/getNetworkConfig.ts
```diff
@@ -0,0 +1,6 @@
+import { NetworkConfig, getConfig } from "../../deps";
```
voltrevo:
```suggestion
import { NetworkConfig, getConfig } from "../../deps.ts";
```

aggregator/src/app/app.ts
```diff
@@ -12,10 +12,13 @@ import TxTable from "./TxTable.ts";
 import createQueryClient from "./createQueryClient.ts";
 import Mutex from "../helpers/Mutex.ts";
 import Clock from "../helpers/Clock.ts";
+import getNetworkConfig from "../helpers/networkConfig.ts";
```
voltrevo: This file does not exist. Looks like you meant to do a rename here.

README.md
```diff
@@ -0,0 +1,110 @@
...
+
+`./extension/.env`
+```
+VERIFICATION_GATEWAY_ADDRESS=addresses.verificationGateway
+```
+
+## Run
+
+```sh
+docker-composer up -d postgres # Or see local postgres instructions in ./aggregator/README.md#PostgreSQL
```
voltrevo:
```suggestion
docker-compose up -d postgres # Or see local postgres instructions in ./aggregator/README.md#PostgreSQL
```

aggregator/README.md
```diff
@@ -31,6 +31,15 @@ commands.
 
 ### PostgreSQL
 
+#### With docker-compose
+
+```sh
+cd .. # root of repo
+docker-composer up -d postgres
```
voltrevo:
```suggestion
docker-compose up -d postgres
```

voltrevo (changes requested):
Had a go at testing these changes this morning but it looks like there are several loose ends. I'm not sure whether this was meant to be considered ready? If not, could you mark the PR as a draft?

> Moved these to bls-wallet-clients imports, can just use yarn link in other spots. Unsure what to do for deno, may need to -alpha tag npm releases to test out?

Yeah it's a bit awkward having to publish for testing hey. Anyway, I've published 0.2.5-rc.1 from this branch, let me know if you need any additional versions. Maybe we should create a dedicated test module on npm.

aggregator/src/app/app.ts
jacque006: Fixed, missed in rename.

README.md
jacque006: Step removed and local and rinkarby setup instructions clarified.

contracts/scripts/deploy_all.ts
```diff
@@ -0,0 +1,85 @@
...
+    parameters: {},
+    addresses: {
+      create2Deployer: create2Deployer.address,
+      precompileCostEstimator: precompileCostEstimatorAddress,
+      blsLibrary: fx.blsLibrary.address,
+      verificationGateway: fx.verificationGateway.address,
+      blsExpander: fx.blsExpander.address,
+      testToken,
+      // TODO Is this correct for current compatibility?
+      rewardToken: testToken,
```
jacque006: Thoughts? Still being used by one test case.

jacque006:
This should now be up to date with latest changes from `main`. I've retested what I can but there appear to be a couple points that have broken with recent updates:
- `extension` has not been updated with latest changes from `contracts/clients`. TS typechecking errors.
- `clients` and `aggregator` may not have been fully updated with `CREATE2` changes in `contracts`. Seeing this in some errors running tests on `aggregator`.

That being said, I feel pretty comfortable/safe merging this in its current state and tackling the above moving forward.

@voltrevo your re-review would be appreciated when you're back in.

voltrevo (approved):

PR #34, opened 2021-11-03
Use clients+signer modules in contracts

### LINKED ISSUE #21: Contracts: Refactor to make use of BlsWallet and VerificationGateway from bls-wallet-clients.

voltrevo:
## What is this PR doing?
Updates the contracts project to use the new clients and signer modules.

Also builds on top of #50 and #52, which are the latest updates to those modules, so they are included in the diff here.

I'd advise just reviewing here, but happy to talk about getting #50 and #52 through separately. If we merge this directly, we can simply close (ie cancel) those PRs.

Run the included tests, also @jzaki try doing any other tasks that ought to work in the contracts project.

Resolves #21

## Checklist

## Guidelines


clients/src/BlsWallet.ts
```diff
@@ -1,19 +1,33 @@
...
+  }
+  | {
+    ethValue?: BigNumber;
+    contract: ethers.Contract,
+    method: string;
+    args: string[];
+  }
+);
+
 export default class BlsWallet {
```
jzaki, line 31: Will files like BLSWallet.ts and VerificationGateway.ts be replaced by the typechain types, or will they use the typechain types (like BLSWallet and VerificationGateway) inside? If the latter, should these .ts files have different names, like BLSWalletWrapper.ts / BLSWalletHelper.ts?

clients/src/contractAbis/BlsWalletAbi.ts
```diff
@@ -1,43 +1,49 @@
 export default [
```
jzaki, line 1: Can files like `...Abi.ts` (and `...Bytecode.ts`) be removed and references updated to point to the compiler generated ones? Okay to be a separate issue/pr, or not if we're keeping it decoupled for a reason.

contracts/shared/helpers/Fixture.ts
```diff
@@ -75,115 +40,89 @@ export default class Fixture {
...
-    let blsWalletImpl = await create2Fixture.create2Contract("BLSWallet") as BLSWallet;
+    let blsWalletImpl = await create2Fixture.create2Contract("BLSWallet");
     try {
       await (await blsWalletImpl.initialize(
         ethers.constants.AddressZero
       )).wait();
     } catch (e) {}
 
     // deploy Verification Gateway
-    let verificationGateway = await create2Fixture.create2Contract("VerificationGateway") as VerificationGateway;
```
jzaki, line 105: VerificationGateway was the typechain type, and naming with lowercase contract name the convention for Contract instances.

contracts/shared/helpers/Fixture.ts
```diff
@@ -75,115 +40,89 @@ export default class Fixture {
...
 
     // deploy BLSExpander Gateway
-    let blsExpander = await create2Fixture.create2Contract("BLSExpander") as BLSExpander;
+    let blsExpander = await create2Fixture.create2Contract("BLSExpander");
     try {
-      await (await blsExpander.initialize(verificationGateway.address)).wait();
+      await (await blsExpander.initialize(vgContract.address)).wait();
     } catch (e) {}
 
+    const verificationGateway = new VerificationGateway(vgContract.address, vgContract.signer);
```
jzaki, line 75: This naming will be less conventional since it is not the verification gateway contract (ethers/typechain), but rather a wrapper to interact with it.

clients/src/BlsWallet.ts
jzaki: As discussed will create a ticket for explicit typechain integration post-merge.

clients/src/contractAbis/BlsWalletAbi.ts
jzaki: As discussed will create a ticket for explicit linking post-merge (and when signer is within /contracts).

contracts/shared/helpers/Fixture.ts
jzaki: As discussed "VerificationGateway.ts" will be replaced with typechain type.

contracts/shared/helpers/Fixture.ts
jzaki: As discussed "VerificationGateway.ts" will be replaced with typechain type. And BLSWallet.ts can have a more appropriate name leaving BLSWallet as the typechain type.

PR #40, opened 2021-11-08
Support for general transactions in extension

### LINKED ISSUE #6: Expand supported dapp transactions from basic transfers to general functions

### LINKED ISSUE #30: Represent general transactions in extension confirm dialog

kautukkundan:
closes #30

- added a new field in the create transaction window to accept encoded function data
- passing the function data to confirmation dialog
- showing different prompt message based on the function data

I had to include #6 as it contains base for #30

### DIFF
--- a/extension/source/Background/RequestHandler.ts
+++ b/extension/source/Background/RequestHandler.ts
@@ -51,15 +51,27 @@ export default function RequestHandler(
           throw new Error('Not implemented: "value" field missing');
         }
 
+        if (tx.data === undefined) {
+          throw new Error('Not implemented: "function data" field missing');
+        }
+
         if (app.wallet === undefined) {
           throw new Error('No wallet available');
         }
 
+        let promptText: string;
+        if (tx.data === '0x') {
+          promptText = `ETH Transfer 
+          ${formatBalance(tx.value, 'ETH')} 
+          to ${formatCompactAddress(tx.to)}`;
+        } else {
+          promptText = `Contract Interaction ${formatCompactAddress(tx.to)} 
+          value ${formatBalance(tx.value, 'ETH')} 
+          data  ${tx.data}`;
+        }
+
         const promptResult = await promptUser({
-          promptText: `Send ${formatBalance(
-            tx.value,
-            'ETH',
-          )} to ${formatCompactAddress(tx.to)}?`,
+          promptText,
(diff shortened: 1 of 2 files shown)


extension/source/Background/RequestHandler.ts
```diff
@@ -51,15 +51,27 @@ export default function RequestHandler(
...
+        let promptText: string;
+        if (tx.data === '0x') {
+          promptText = `ETH Transfer 
+          ${formatBalance(tx.value, 'ETH')} 
+          to ${formatCompactAddress(tx.to)}`;
+        } else {
+          promptText = `Contract Interaction ${formatCompactAddress(tx.to)} 
+          value ${formatBalance(tx.value, 'ETH')} 
+          data  ${tx.data}`;
+        }
```
kautukkundan, line 71: I am still not sure if passing data via URL search params is the best way forward but it gets the job done for now. I guess `window.postMessage` can be useful here.

sending data as an object will also help in formatting + better UI design in the confirm dialog box

Happy to hear your thoughts

kautukkundan: TODO - 
use the `tx.data` to show more details about the function (method name etc) similar to metamask. Right now it will only show the hex data.

Need to study how metamask does it.

jacque006: This may be useful https://docs.ethers.io/v5/api/utils/abi/interface/#Interface--parsing . You would need to get the contract ABI from the address and create a `Contract` object, but then you can display more details about the tx.

jacque006 (approved):
LGTM :+1:

extension/source/Background/RequestHandler.ts
kautukkundan: Yeah, This works at web app level but I don't think that the entire ABI gets passed over when metamask receives the tx (i.e at extension level)

jacque006: Right, you would need to use another source. For example, Etherscan allows you to retrieve the ABI for a given contract https://docs.etherscan.io/tutorials/read-write-contract-using-abis#2.-obtaining-a-contracts-abi . When the `To` address is a contract, you would do that ABI lookup. I imagine there are other sources for ABIs as well.

PR #42, opened 2021-11-10
Added inpage provider

chaitanyapotti:
This provider meets the EIP-1193 spec completely and is usable with all existing ethereum web3 libraries.


kautukkundan (approved):
LGTM
would be great if @voltrevo could also go over it once before merging

jacque006:
@chaitanyapotti Will we have any conflicts if someone has MetaMask installed alongside with Quill, such as one overriding the other at `window.ethereum` (https://github.com/MetaMask/providers#usage)?

chaitanyapotti:
I haven't added that part yet. We can decide on whether we should use window.ethereum or window.quill in this extension at a later date

voltrevo:
> @chaitanyapotti Will we have any conflicts if someone has MetaMask installed alongside with Quill, such as one overriding the other at `window.ethereum` (https://github.com/MetaMask/providers#usage)?

Yeah that's going to be an issue. Currently I think we overwrite and also set `window.quillEthereumProvider` so you can be explicit if you want to. We should probably look into what else we can do here like notifying the user if a conflict is detected.

voltrevo:
@chaitanyapotti I marked this as a draft because (correct me if I'm wrong) this is currently just adding unused code. Let us know if I'm mistaken or when it becomes ready.

voltrevo:
It looks like this design is heavily copied (from torus?). I'm concerned that we're bringing in more than we need by doing that. E.g. Why is there a `BaseProvider`? Why have streams been introduced? Also need to make sure the licence is being complied with if it's not public domain - getting tangled up in licences via copy+modify makes me nervous (more complicated than when you have an exact copy by using a dependency).

voltrevo:
Also, this appears to be aimed at accurately implementing the provider spec (e.g. including deprecated APIs), which is quite different from adding support for arbitrary transactions. Wouldn't it be easier to do that first? It would more naturally build on top of the current system and ties in with the work @kautukkundan is doing to represent those transactions in the confirm dialog.

(Not to say that we don't need to implement the spec accurately, I assume we will and these contributions can still be useful.)

voltrevo:
(Sorry if this is raining on your parade a bit... maybe we should have a chat about it?)

extension/source/Controllers/Account/AccountTrackerController.ts
```diff
@@ -0,0 +1,155 @@
...
+      const currentBlock = this.config._currentBlock;
+      if (!currentBlock) return;
+      const addresses = Object.keys(accounts);
+      if (addresses.length > 0) {
+        await Promise.all(
+          addresses.map((x) => this._updateAccount(x, currentBlock)),
+        );
+      }
+    } catch (error) {
+      // ignore error
```
voltrevo:
```suggestion
      console.error(error);
```

extension/source/Controllers/Account/IAccountTrackerController.ts
```diff
@@ -0,0 +1,37 @@
...
+  syncAccounts(): void;
+
+  /**
+   * Refreshes the balances of all accounts
+   */
+  refresh(): Promise<void>;
+}
+
+export interface AccountTrackerConfig extends BaseConfig {
+  _currentBlock?: string;
```
voltrevo, line 16: Why is the current block in config? Isn't that state?

extension/source/Controllers/Account/AccountTrackerController.ts
```diff
@@ -0,0 +1,155 @@
...
+    try {
+      const { accounts } = this.state;
+      const currentBlock = this.config._currentBlock;
+      if (!currentBlock) return;
+      const addresses = Object.keys(accounts);
+      if (addresses.length > 0) {
+        await Promise.all(
+          addresses.map((x) => this._updateAccount(x, currentBlock)),
+        );
+      }
```
voltrevo:
```suggestion
      await Promise.all(
        addresses.map((x) => this._updateAccount(x, currentBlock)),
      );
```

extension/source/Controllers/Account/AccountTrackerController.ts
```diff
@@ -0,0 +1,155 @@
...
+    address: string,
+    currentBlock: string,
+  ): Promise<void> {
+    const currentChainId = this.getCurrentChainId();
+    if (currentChainId === 'loading') {
+      return;
+    }
+    return new Promise((resolve, reject) => {
+      // query balance
+      this.ethQuery.sendAsync(
```
voltrevo: What's the idea behind `ethQuery`? Can't we use ethers directly?
https://docs.ethers.io/v5/api/providers/provider/#Provider-getBalance

extension/source/Controllers/BaseController.ts
```diff
@@ -0,0 +1,135 @@
...
+   * @param overwrite - Overwrite config instead of merging
+   * @param fullUpdate - Boolean that defines if the update is partial or not
+   */
+  configure(config: Partial<C>, overwrite = false, fullUpdate = true): void {
+    if (fullUpdate) {
+      this.internalConfig = overwrite
+        ? (config as C)
+        : Object.assign(this.internalConfig, config);
+
+      Object.keys(this.internalConfig).forEach((key) => {
```
voltrevo, line 91: Is there a reason you prefer `items.forEach((item) => { ... })` over `for (const item of items) { ... }`?

extension/source/Controllers/Block/PollingBlockTracker.ts
```diff
@@ -0,0 +1,98 @@
...
+import { createRandomId, timeout } from '../utils';
+import {
+  PollingBlockTrackerConfig,
+  PollingBlockTrackerState,
+} from './IBlockTrackerController';
+import { ExtendedJsonRpcRequest } from '../Network/INetworkController';
+
+const sec = 1000;
+
+class PollingBlockTracker extends BaseBlockTracker<
```
voltrevo, line 11: I feel like this base+polling split class design is doing too much. Do you think we could simplify things with a single controller that uses `provider.on('block', ...)` instead of `eth_blockNumber`?

voltrevo: (In general I'm confused about the absence of ethers from this design. Why would that be? Is it also absent from the metamask/torus code you're drawing from?)

extension/source/Controllers/Network/NetworkController.ts
```diff
@@ -0,0 +1,249 @@
...
+      );
+    });
+  }
+
+  async getEIP1559Compatibility(): Promise<boolean> {
+    const { EIPS } = this.state.properties;
+    // log.info('checking eip 1559 compatibility', EIPS[1559])
+    if ((EIPS as any)[1559] !== undefined) {
+      return (EIPS as any)[1559];
+    }
```
voltrevo:
```suggestion
    const { EIPS } = this.state.properties as any;
    // log.info('checking eip 1559 compatibility', EIPS[1559])
    if (EIPS[1559] !== undefined) {
      return Boolean(EIPS[1559]);
    }
```

voltrevo: If `Record<string, boolean | undefined>` is an accurate type for `EIPS` then it would be nice to add that typing to `this.state.properties` rather than do casts and conversions here.

extension/source/Controllers/Network/NetworkController.ts
```diff
@@ -0,0 +1,249 @@
...
+            (latestBlock as Block).baseFeePerGas !== undefined;
+          this.update({
+            properties: {
+              EIPS: { 1559: supportsEIP1559 },
+            },
+          });
+          resolve(supportsEIP1559);
+        },
+      );
+    });
```
voltrevo: Haven't been commenting on all these but there are a lot of examples of code like this:

```ts
return new Promise((resolve, reject) => {
  someAsyncFunction((err, value) => {
    if (err) return reject(err);
    // do some processing on `value`
    resolve();
  });
});
```

Which should be like this:

```ts
const value = await new Promise((resolve, reject) => {
  someAsyncFunction((err, value) => {
    if (err) { reject(err); }
    else { resolve(value); }
  });
});

// do some processing on `value`
```

Also can be further improved by promisifying `someAsyncFunction` so you can simply write:

```ts
const value = await promisifedFunction();
// do some processing on `value`
```

Or even just promisified inline:

```ts
const value = await promisify(someAsyncFunction)();
// do some processing on `value`
```

extension/source/Controllers/Account/IAccountTrackerController.ts
chaitanyapotti: Added as a comment.
Config represents variables which subscriber doesn't care about.
State represents variables which subscriber is interested in

extension/source/Controllers/Account/AccountTrackerController.ts
chaitanyapotti: It's by design. ethers is a bit heavy handed in it's approach to blockchain calls.
we use ethQuery within controllers for 3 reasons:
1 - light weight processing (No listeners added to provider)
2- goes through our rpc engine. (can do additional processing inside our middlewares)
3 - uses fetch middleware at the end to communicate with rpc endpoint (which has retry logic built-in)

extension/source/Controllers/BaseController.ts
chaitanyapotti: Eslint recommended.

extension/source/Controllers/Block/PollingBlockTracker.ts
chaitanyapotti: Answered comment about ethers.
polling is split into separate class so that in the future we can use websocket block tracker if necessary.
that wouldn't have an infinite loop and would be notified of any new blocks.

extension/source/Controllers/Network/NetworkController.ts
chaitanyapotti: addressed by adding full types

extension/source/Controllers/Network/NetworkController.ts
chaitanyapotti: Added promise based augmentation method `request`

voltrevo (changes requested):
It's not really feasible to review this in depth, but I've skimmed through a couple times now and I'm keen to get this merged when we're able to experiment with it and manually verify it does what we expect.

A few code-specific things are needed though:
- Remove commented code
- Fix imports so that things compile
- Fix unused variables (seen when building)
- Rename `TorusController` -> `QuillController`

It looks like `QuillController` is intended to run in the background script. I don't see anything attaching to `runtime.onMessage` to facilitate communicating with it from the ui though. Am I missing something?

chaitanyapotti:
Addressed the comments in jitsi call.
Here's how metamask connects ui to background:
https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/ui.js
https://github.com/MetaMask/metamask-extension/blob/develop/ui/index.js

voltrevo (changes requested):
So close... unfortunately I can't find that function you (@chaitanyapotti) mentioned that's a placeholder for getting the user permission.

@chaitanyapotti Can you please add this or point out where I've missed it?

voltrevo:
🤞

PR #46, opened 2021-11-15
Added auto-labeller workflow

kautukkundan:
automatically adds label to PR based on the part of the codebase which changed


.github/workflows/labeler.yml
```diff
@@ -0,0 +1,11 @@
...
+on:
+- pull_request_target
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/labeler@main
+      with:
+        repo-token: "${{ secrets.GITHUB_TOKEN }}"
```
jacque006:
```suggestion
        repo-token: "${{ secrets.GITHUB_TOKEN }}"

```

.github/labeler.yml
```diff
@@ -0,0 +1,8 @@
+aggregator:
+  - aggregator/**/*
+extension:
+  - extension/**/*
+contracts:
+  - contracts/**/*
+clients:
+  - clients/**/*
```
jacque006, line 8:
```suggestion
clients:
  - clients/**/*
signer:
  - signer/**/*
```

jacque006 (changes requested):
Newline nitpick, and we should add signer as well. Once those are in, :shipit:

jacque006 (approved):

PR #55, opened 2021-11-19
Merge signer into clients and move inside contracts

### LINKED ISSUE #54: Merge signer into clients and put clients inside contracts

voltrevo:
## What is this PR doing?

This is a simplification that's important for development inside `contracts`. Basically, this functionality is integral to how `contracts` is developed, because they both depend on each other. It wasn't obvious until now because `contracts` basically had its own rough version of doing the same thing.

So we now just have one module, bls-wallet-clients that contains what was in bls-wallet-signer, and that is inside `contracts`. This way, `contracts` is able to import the typescript code directly, and `bls-wallet-clients` is also somewhat independent inside, so we can build and publish it in isolation.

This is still not the final form of what we want. In particular, type simplifications are coming by changing the representation inside the contracts to be in sync with what the aggregator/extension need. Also, you still need to manually copy the abis/bytecode, and we'll make that part of the build process later.

`yarn`, `yarn hardhat test`.

Resolves #54.

## Checklist

## Guidelines


contracts/clients/package.json
```diff
@@ -1,33 +1,20 @@
...
-    "bls",
-    "signature",
-    "aggregation"
-  ],
-  "author": "Andrew Morris",
-  "license": "MIT",
-  "bugs": {
-    "url": "https://github.com/jzaki/bls-wallet-signer/issues"
-  },
-  "homepage": "https://github.com/jzaki/bls-wallet-signer#readme",
```
jacque006, line 29: Some of these might be worth keeping to make package more discoverable, but can always add back in later.

contracts/package.json
```diff
@@ -7,9 +7,7 @@
   "author": "James Zaki",
   "license": "MIT",
   "dependencies": {
-    "@openzeppelin/contracts": "^4.3.2",
-    "bls-wallet-clients": "0.3.0-9bdca2f",
-    "bls-wallet-signer": "^0.8.0-rc.1"
```
jacque006, line 12: If you want to keep the imports the same as elsewhere, you should be able to use a local path to install here for `bls-wallet-clients`. https://docs.npmjs.com/cli/v6/configuring-npm/package-json#local-paths

jacque006 (approved):

PR #56, opened 2021-11-19
Decode method name from encoded data when sending Tx

kautukkundan:
## What is this PR doing?
- [x] changed Confirm class component to functional component 
- [x] changed format of prompt text
- [x] fetch method from parity registry 
- [x] fetch method from etherscan api
- [x] added useInputDecode hook to decode encoded method data
- [x] added decoded function to UI

1. install new packages using `yarn`
2. add env variable `PROVIDER_URL` = alchemy/infura endpoint for mainnet and `ETHERSCAN_KEY` = currently unused
3. run `ENV=local yarn run dev:chrome` and import extension as a package inside chrome
4. create a tx by inputting contract addr and function data - [can use this tx](https://etherscan.io/tx/0xdad712c90fb9e823bdd01b1c105a7a0ec4f7a63ddf7e2e240facb4318661fd80) or any other from etherscan

## Checklist

## Guidelines


kautukkundan:
confirmation box now displays actual method name

Screenshot 2021-11-22 at 3 18 24 AM

extension/source/hooks/useInputDecode.tsx
```diff
@@ -0,0 +1,84 @@
...
+        if (!method) {
+          method = await getMethodFromEtherscan(to, data);
+        }
+      } catch (error) {
+        console.log({ error });
+      }
+
+      if (method) {
+        setMethod(formatMethod(method));
+      }
```
jacque006: These can be run in parallel to reduce waiting time if `getMethodFromOnChainRegistry` does not find anything.
```suggestion
      try {
        const registryPromise = getMethodFromOnChainRegistry(data);
        const etherScanPromise = getMethodFromEtherscan(to, data);
        const method = await registryPromise ?? await etherScanPromise;
        if (method) {
          setMethod(formatMethod(method));
        }
      } catch (error) {
        console.log({ error });
      }
```

jacque006 (commented):
One optimization for address lookup, other changes look good.

One other note as we look at using functional React components. There are a number of optimizations you can do using `useMemo` & `useCallback` to reduce re-renders. https://www.backtofrontend.com/react-optimizations-memo-usecallback-usememo/ does a very good job of highlighting some basic cases. There would be some good uses for these especially for the `useInputDecode` hook, but might be overkill at this point. If you're interested in trying those out, I'd be happy to show you offline :)

extension/source/hooks/useInputDecode.tsx
kautukkundan: sweet!

kautukkundan:
> There are a number of optimizations you can do using `useMemo` & `useCallback` to reduce re-renders. 

Yeah Agreed, however each trigger creates a new self-contained window (with independent states) and the state is set once (default -> final) via url params. Since the state does not depend on any other interaction it wont change at all and the window will only be rendered once. So I guess for now memoization might not be needed here. WDYT?

jacque006:
In that case the memoization is likely a premature optimization. Can revisit when/if that hook is used in other places in the future.

extension/source/hooks/useInputDecode.tsx
kautukkundan: done 👍🏼

jacque006 (approved):

PR #69, opened 2021-12-14
Gateway upgrade path

### LINKED ISSUE #44: Wallet address lookup
VGs to use a local override mapping of wallets:
- [ ] access wallet address via public key hash (before using create2 determined address)
- ~~[ ] gateways access wallet's desired public key via wallet contract~~

Background: VG previously used a mapping (hash=>wallet), now deterministic create2. With upgradable contracts (and ability to change gateway), need to ensure walletFromHash function will endure, and consider multi-network scenarios.

jzaki:
## What is this PR doing?
Adds upgradability to VG and BLSWallet contracts.

```
cd contracts
yarn hardhat test
```
See `tests/upgrade-tests.ts` results.

Closes #44

## Checklist

## Guidelines

### DIFF
--- a/contracts/contracts/BLSWallet.sol
+++ b/contracts/contracts/BLSWallet.sol
@@ -52,7 +52,14 @@ contract BLSWallet is Initializable, IBLSWallet
     /**
     Wallet can migrate to a new gateway, eg additional signature support
      */
-    function setTrustedBLSGateway(address blsGateway) public onlyTrustedGateway {
+    function setTrustedBLSGateway(address blsGateway) public onlyThis {
+        uint256 size;
+        // solhint-disable-next-line no-inline-assembly
+        assembly { size := extcodesize(blsGateway) }
+        require(
+            (blsGateway != address(0)) && (size > 0),
+            "BLSWallet: gateway address param not valid"
+        );
         trustedBLSGateway = blsGateway;
     }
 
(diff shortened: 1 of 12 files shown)


contracts/contracts/BLSWallet.sol
```diff
@@ -55,8 +69,32 @@ contract BLSWallet is Initializable
...
+            bool _success,
+            bytes[] memory _results
+        ) {
+            success = _success;
+            results = _results;
+        }
+        catch {
+            success = false;
+        }
+        incrementNonce(); // regardless of outcome of operation
```
jacque006, line 86: :+1:

contracts/shared/helpers/Fixture.ts
```diff
@@ -64,29 +59,35 @@ export default class Fixture {
...
+      "BLSExpander",
+      ethers.utils.defaultAbiCoder.encode(
+        ["address"],
+        [verificationGateway.address],
+      ),
     );
 
+    // try {
+    //   await (await blsExpander.initialize(vgContract.address)).wait();
+    // } catch (e) {}
```
jacque006: Are these commented blocks still needed if new contracts are being deployed? Keeping them for reference?

jzaki: I'll get rid of those. It was was from when the gateway and expander were upgradable, but that is no longer the case.
Namely so that wallet holders needn't fear that the VG will get upgraded on them. The opt-in model is now the case where a wallet holder can change their VG.

jacque006 (approved):

PR #72, opened 2021-12-22
Wallet key recovery

### LINKED ISSUE #70: BLSWallet recovery
- [x] bls key holder able to set a recovery hash (with delay)
- [x] bls key holder able to set new bls public key (with delay)
- [x] isAdmin check using hash of msg.sender (+salt?) checked against recovery hash 
- [x] Admin able to instantly set bls public key, and also cancel pending updates

jzaki:
## What is this PR doing?
Adding recovery functionality to wallets

`yarn hardhat test` from `contracts`

Closes #70 

## Checklist

## Guidelines

### DIFF
--- a/contracts/contracts/BLSWallet.sol
+++ b/contracts/contracts/BLSWallet.sol
@@ -169,6 +169,22 @@ contract BLSWallet is Initializable, IBLSWallet
         }
     }
 
+    function recover(
+        uint256[4] calldata newBLSKey
+    ) public onlyTrustedGateway {
+        // set new bls key
+        blsPublicKey = newBLSKey;
+        // clear any pending operations
+        pendingRecoveryHashTime = type(uint256).max;
+        pendingRecoveryHash = bytes32(0);
+        pendingBLSPublicKeyTime = type(uint256).max;
+        pendingBLSPublicKey = [0,0,0,0];
+        pendingGatewayTime = type(uint256).max;
+        pendingBLSGateway = address(0);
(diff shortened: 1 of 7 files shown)


contracts/contracts/BLSWallet.sol
```diff
@@ -14,31 +14,75 @@ import "./interfaces/IWallet.sol";
...
+        uint256[4] oldBLSKey,
+        uint256[4] newBLSKey
+    );
+    event GatewayUpdated(
+        address oldGateway,
+        address newGateway
+    );
+    event ProxyAdminFunctionApproved(
+        bytes approvedProxyAdmin
+    );
```
jacque006, line 59: Something to consider for the future would be to break this down into separate abstract component contracts that manage the state/pending/recovery of these sensitive fields. i.e. `blsPublicKey` and its pending fields, events, mutators, and set pending calls, then have `BLSWallet` inherit them. This would reduce the code size/complexity of `BLSWallet.sol`, and make it easier to break out isolated unit test suites for them.

jacque006 (approved):
:+1: Solid test coverage as well

contracts/contracts/BLSWallet.sol
jzaki: Agreed, at that stage it would be good to factor in upgradability of wallets (bls and potentially non-bls). Which needs to consider the inheritance order since memory locations when upgrading are important. -> Created #73

PR #75, opened 2021-12-24
Update extension with bundle changes

### LINKED ISSUE #74: Update extension to use new bundle format
# Problem

After https://github.com/jzaki/bls-wallet/pull/68 , the extension no longer works with the updated aggregator or smart contracts and the new bundle format for sending transactions to be aggregated in `bls-wallet-clients`.

# Solution

Update extension to use new bundle format in `bls-wallet-clients`.

jacque006:
## What is this PR doing?

- Updating extension to use new `bls-wallet-clients` and new bundle payload for sending transactions to be aggregated.
- Add PublicKey and PublicKeyStr to BLSWalletWrapper.
- Avoid publishing non-definition typescript files from typechain.
- Fix aggregator bundle POST path in clients.
- Remove @ethersproject specific deps from clients.
- Remove wallet creation paths from aggregator and extension.
- Update main README with client -> aggregator local workflow.
- Add send_eth.ts test script.
- Update labeler CI. Remove signer, don't add contracts on client only changes.

Run extension and complete current onboarding workflow & successfully send a transaction to the aggregator.

Resolves https://github.com/jzaki/bls-wallet/issues/74

## Checklist

## Guidelines

~TODO~
- [x] Remove wallet creation functionality from extension and aggregator.
- [x] Create issue for wallet creation to evaluate in future.


aggregator/deps.ts
```diff
@@ -39,18 +39,18 @@ export type {
...
+} from "https://esm.sh/bls-wallet-clients@0.5.2";
 
 export {
   Aggregator as AggregatorClient,
   BlsWalletWrapper,
   getConfig,
   VerificationGateway,
   VerificationGateway__factory,
-} from "https://esm.sh/bls-wallet-clients@0.5.0-d6d7dd4";
+} from "https://esm.sh/bls-wallet-clients@0.5.2";
```
voltrevo: Is there a reason you published non-experimental 0.5.2?

contracts/clients/package.json
```diff
@@ -9,7 +9,7 @@
   "license": "MIT",
   "private": false,
   "scripts": {
-    "build": "rm -rf dist && mkdir dist && cp -rH typechain dist/typechain && tsc",
+    "build": "rm -rf dist && mkdir dist && cp -rH typechain dist/typechain && find ./dist/typechain -type f \\! -name '*.d.ts' -name '*.ts' -delete && tsc",
```
voltrevo, line 12: Avoids publishing non-definition typescript files from typechain? Nice. Could you add something about this to the PR description?

contracts/clients/tsconfig.json
```diff
@@ -68,5 +68,5 @@
     /* Advanced Options */
     "skipLibCheck": true,                           /* Skip type checking of declaration files. */
     "forceConsistentCasingInFileNames": true        /* Disallow inconsistently-cased references to the same file. */
-  }
+  },
```
voltrevo: Interesting trailing comma here. That's invalid json, but I suppose `tsc` is ok with it?

extension/source/App.ts
```diff
@@ -192,7 +190,7 @@ export default class App {
       return undefined;
     }
 
-    return this.blsWalletSigner.getPublicKey(this.state.privateKey);
+    return this.wallet?.blsWalletSigner.getPublicKeyHash(this.state.privateKey);
```
voltrevo: Should be `getPublicKey` right? Not `getPublicKeyHash`?

(If the public key hash is what we really want then we should rename the method.)

extension/source/App.ts
```diff
@@ -206,14 +204,9 @@ export default class App {
...
 
     try {
-      const creationTx = await BlsWallet.signCreation(
-        this.state.privateKey,
-        env.NETWORK_CONFIG.addresses.verificationGateway,
-        this.provider,
-      );
-
-      const createResult = await this.aggregator.createWallet(creationTx);
+      throw new Error('App: Wallet creation not implemented');
```
voltrevo: We're not doing an explicit create anymore, so the goal is to just remove this `createWallet` method entirely.

contracts/clients/package.json
jacque006: Added to description.

aggregator/deps.ts
jacque006: `0.5.0` failed webpack transpilation due to .ts files being included from typechain as you pointed out in https://github.com/jzaki/bls-wallet/pull/75#discussion_r776168168 . I thought I had a fix in `0.5.1` but realized shortly thereafter I had made a mistake. `0.5.2` has the .ts files properly removed.

Given that `0.4.x` -> `0.5.0` broke package level compatibility for the extension and potentially other similar consumers, it seemed prudent to push a prod tag resolving the issue.

extension/source/App.ts
jacque006: I added this as typing filler, `getPublicKey` now returns an `Array(4)` instead of a string. This will be changed in future commits.

contracts/clients/tsconfig.json
jacque006: Apparently haha, will remove.

extension/source/App.ts
jacque006: I still think there are cases where someone may want to non-lazily create a wallet, but I imagine those are not needed anytime soon.

I will add tasks to this PR to remove this, similar functionality in aggregator, and create an issue for manual wallet creation we can evaluate later.

extension/source/App.ts
jacque006: Added `getPublicKeyStr` to signer and `PublicKey` and `PublicKeyStr` to `BLSWalletWrapper`.

extension/source/App.ts
jacque006: Removed wallet creation paths and logic from agg and ext.

contracts/clients/package.json
```diff
@@ -9,21 +9,15 @@
...
     "publish-experimental": "node scripts/showVersion.js >.version && npm version $(node scripts/showBaseVersion.js)-$(git rev-parse HEAD | head -c7) --allow-same-version && npm publish --tag experimental && npm version $(cat .version) && rm .version",
     "publish-experimental-dry-run": "node scripts/showVersion.js >.version && npm version $(node scripts/showBaseVersion.js)-$(git rev-parse HEAD | head -c7) --allow-same-version && npm publish --tag experimental --dry-run && npm version $(cat .version) && rm .version"
   },
   "dependencies": {
-    "@ethersproject/abi": "^5.4.0",
-    "@ethersproject/bignumber": "^5.4.1",
-    "@ethersproject/bytes": "^5.4.0",
-    "@ethersproject/keccak256": "^5.4.0",
-    "@ethersproject/sha2": "^5.4.0",
-    "@ethersproject/solidity": "^5.4.0",
```
jacque006, line 25: @voltrevo Given discussion today and what I saw in my tests this seemed safe to do. Let me know if I should revert.

extension/source/Background/index.ts
```diff
@@ -1,18 +1,18 @@
...
 import RequestHandler from './RequestHandler';
 
-import { AGGREGATOR_URL, NETWORK_CONFIG, CHAIN_RPC_URL } from '../env';
+import { AGGREGATOR_URL /* , NETWORK_CONFIG */, CHAIN_RPC_URL } from '../env';
 
 (async () => {
   const app = new App(
-    await initBlsWalletSigner({ chainId: NETWORK_CONFIG.auxiliary.chainid }),
+    // TODO Is this still needed for something else?
+    // await initBlsWalletSigner({ chainId: NETWORK_CONFIG.auxiliary.chainid }),
```
jacque006: @voltrevo @jzaki Due to `BLSWalletWrapper` now wrapping the signer, this does not appear to be needed anymore. Safe to remove?

extension/source/Popup/index.tsx
```diff
@@ -1,19 +1,19 @@
...
+import { AGGREGATOR_URL /* , NETWORK_CONFIG */, CHAIN_RPC_URL } from '../env';
 import Popup from './Popup';
 
 import './styles.scss';
 
 const appPromise = (async () =>
   new App(
-    await initBlsWalletSigner({ chainId: NETWORK_CONFIG.auxiliary.chainid }),
+    // TODO Is this still needed somewhere else?
+    // await initBlsWalletSigner({ chainId: NETWORK_CONFIG.auxiliary.chainid }),
```
jacque006: @voltrevo @jzaki Same here.

jacque006:
@voltrevo @kautukkundan Feel free to push any changes to this if you see something that needs to be corrected.

extension/source/App.ts
voltrevo: Interesting. I want to aim for handling these things in a more uniform way but not sure whether it's suitable here or not. In any case, it's something I'd rather fix as a followup than have in this PR if it looks like a good idea later.

voltrevo: ("Handling in a uniform way" means using the string tuple everywhere instead of converting between formats for different projects.)

contracts/clients/package.json
voltrevo: Nah, looks good. Nice one.

extension/source/App.ts
voltrevo: > I still think there are cases where someone may want to non-lazily create a wallet, but I imagine those are not needed anytime soon.

True. Should be ok with just submitting an empty operation from the client for that though. Keeps the interface simple.

extension/source/Background/index.ts
voltrevo: Yeah that makes sense. Safe to remove 👍.

extension/source/Popup/index.tsx
voltrevo: Safe to remove 👍.

contracts/clients/src/signer/getPublicKeyStr.ts
```diff
@@ -0,0 +1,16 @@
...
+    signerFactory: hubbleBls.signer.BlsSignerFactory,
+    domain: Uint8Array,
+  ) =>
+  (privateKey: string): string => {
+    const [x0, x1, y0, y1] = getPublicKey(signerFactory, domain)(privateKey);
+    return hubbleBls.mcl.dumpG2([
+      x0.toString(),
+      x1.toString(),
+      y0.toString(),
+      y1.toString(),
```
voltrevo, line 14: The `.toString()`s are a bit odd here... not sure if you realized. I want to explain what's going on here.

The result of `getPublicKey` is understandably typed as `PublicKey`, which is the wide/permissive type we get from typechain, so the elements are `BigNumberish`.

`dumpG2` requires strings, and so `.toString()`, being available on all possibilities of `BigNumberish` satisfies the type checker. However, `dumpG2` actually expects _hex_ strings (right?), so this will only work if all the `BigNumberish` elements are actually already hex strings.

...which they are, because they come from `signer.pubkey`.

tl;dr - `.toString()` satisfies the type checker, but the potential issue flagged by the type checker isn't actually fixed by `.toString()`.

I'm nit picking. I think. It's unlikely this will ever be an issue, but it's a cause of confusion and could be the cause of a bug with subsequent changes that would look reasonable.

I think I'll submit a followup PR that replaces the use of `getPublicKey` with the underlying hubblebls call. Using `as string` would also be preferable, since at least it keeps the flag that something odd is going on (because it is), and avoids the no-op.

voltrevo: https://github.com/jzaki/bls-wallet/issues/83

voltrevo: (Btw it's very possible that I also used the same kind of `.toString()` in the past 😅.)

extension/source/Background/index.ts
voltrevo:
```suggestion
```

extension/source/Popup/index.tsx
```diff
@@ -1,19 +1,19 @@
-import { Aggregator } from 'bls-wallet-clients';
-import { initBlsWalletSigner } from 'bls-wallet-signer';
+import { Aggregator /* , initBlsWalletSigner */ } from 'bls-wallet-clients';
```
voltrevo:
```suggestion
import { Aggregator } from 'bls-wallet-clients';
```

extension/source/Popup/index.tsx
```diff
@@ -1,19 +1,19 @@
...
 import { ethers } from 'ethers';
 import * as React from 'react';
 import ReactDOM from 'react-dom';
 import { browser } from 'webextension-polyfill-ts';
 
 import App from '../App';
-import { AGGREGATOR_URL, NETWORK_CONFIG, CHAIN_RPC_URL } from '../env';
+import { AGGREGATOR_URL /* , NETWORK_CONFIG */, CHAIN_RPC_URL } from '../env';
```
voltrevo:
```suggestion
import { AGGREGATOR_URL, CHAIN_RPC_URL } from '../env';
```

extension/source/Popup/index.tsx
voltrevo:
```suggestion
```

voltrevo (approved):
Amazing stuff 👍

contracts/clients/src/signer/getPublicKeyStr.ts
jacque006: > However, dumpG2 actually expects hex strings (right?), so this will only work if all the BigNumberish elements are actually already hex strings. ...which they are, because they come from signer.pubkey.

(dumpG2)[https://github.com/jzaki/bls-wallet/blob/e366d0af4c339db3f380732550c9319ea4d6982b/contracts/clients/deps/hubble-bls/mcl.ts#L223-L226] explicitly needs the pubkey segments to be strings, and I imagine in hex format. Good eye, there are some cases where this will fail spectacularly if it is a `number` or `BigNumber`:
Screenshot from 2022-01-04 14-01-37

We might want to just have the type for our pubkey be `[string, string,string, string]`, and then we can have a function that converts from the typechain type to our needed type (I don't think `hubble-bls` exports `solG2`). We could also just make a seperate wrapper class for it and put needed utility functions in that class similar to https://github.com/thehubbleproject/hubble-contracts/blob/master/ts/pubkey.ts

jzaki: Perhaps relevant here, some conversions remove preceding 0's from hex, which is problematic when a smart contract call requires the right length. Eg a 20 byte address:
`0x0102030405060708091011121314151617181920`, becoming,
`0x102030405060708091011121314151617181920`

PR #82, opened 2021-12-30
Fix docker build

### LINKED ISSUE #80: Fix aggregator docker build to work with network config
Network config typically points to a json file outside the aggregator subproject. Therefore it won't be included in the docker build and won't work. Fix it.

voltrevo:
## What is this PR doing?

The docker build was broken because our .env now generally points to a networkConfig outside the aggregator subproject, and therefore isn't available inside the docker image.

This fixes that by copying the networkConfig to the top-level and updating this value in the built .env file.

Also refactors `build.ts` into separate functions. You might prefer to go commit-by-commit to inspect the changes.

1. Run the build `./programs/build.ts --sudo-docker`
2. Run the image and check that it starts up normally `sudo docker run --rm -it --net=host <insert image name from build log>`

Resolves #80

## Checklist

## Guidelines

### DIFF
--- a/aggregator/programs/build.ts
+++ b/aggregator/programs/build.ts
@@ -76,38 +76,33 @@ async function ensureFreshBuildDir() {
 }
 
 async function buildEnvironment() {
-  const originalDotEnv = await Deno.readTextFile(dotEnvPath);
+  const repoDotEnv = await Deno.readTextFile(dotEnvPath);
 
-  type NetworkConfigPaths = { repo: string; build: string } | nil;
-  let networkConfigPaths: NetworkConfigPaths = nil;
+  let networkConfigPaths: { repo: string; build: string } | nil = nil;
+  const buildDotEnvLines: string[] = [];
 
-  const dotEnv = originalDotEnv
-    .split("\n")
-    .map((line) => {
-      if (line.startsWith("NETWORK_CONFIG_PATH=")) {
-        const repoNetworkConfigPath = line.slice(
-          "NETWORK_CONFIG_PATH=".length,
-        );
+  for (const line of repoDotEnv.split("\n")) {
+    let buildLine = line;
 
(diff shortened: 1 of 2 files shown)


aggregator/programs/build.ts
```diff
@@ -5,103 +5,164 @@ import { dirname, parseArgs } from "../deps.ts";
...
-      "git",
-      "ls-files",
-      "--others",
-      "--exclude-standard",
-    ),
-  ];
+  await shell.run(
+    "gzip",
+    `${repoDir}/build/${dockerImageName}.tar`,
+  );
```
jacque006, line 167: Can consider pushing this to a container image registry in the future such as Docker Hub. https://github.com/thehubbleproject/hubble-contracts/blob/master/.github/workflows/docker.yml

aggregator/programs/build.ts
```diff
@@ -5,103 +5,164 @@ import { dirname, parseArgs } from "../deps.ts";
...
 
-  console.log("Processing", f);
-  await Deno.mkdir(dirname(`${buildDir}/ts/${f}`), { recursive: true });
-  await Deno.copyFile(f, `${buildDir}/ts/${f}`);
+  if (networkConfigPaths !== nil) {
+    await Deno.copyFile(
+      networkConfigPaths.repo,
+      `${buildDir}/${networkConfigPaths.build}`,
+    );
+  }
```
jacque006, line 110: Generally when running containerized services, the container image does not have configuration included in the image (asides from safe/needed defaults) and instead the user provides that via [a variety of options](https://docs.docker.com/engine/reference/run/#overriding-dockerfile-image-defaults), most common being command overrides, env vars, and file/volume mounts. `geth` for example uses command overrides and volume mounts for loading and persisting data https://github.com/jzaki/bls-wallet/blob/main/docker-compose.yml#L26-L34 . We can consider updating how we want to do that in the future.

jacque006 (approved):

PR #85, opened 2022-01-06
Updated Keyring Controller

kautukkundan:
## What is this PR doing?
  updated bls-wallet-clients version
  added methods create and import BLS accounts
  updated KeyringController interface
  added method to sign transactions

## Checklist

## Guidelines

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -30,7 +30,7 @@
     "@toruslabs/openlogin-jrpc": "^1.3.3",
     "advanced-css-reset": "^1.2.2",
     "async-mutex": "^0.3.2",
-    "bls-wallet-clients": "0.5.2",
+    "bls-wallet-clients": "0.5.3",
     "bls-wallet-signer": "0.6.1",
     "dotenv-webpack": "^7.0.3",
     "emoji-log": "^1.0.2",
--- a/extension/package.json
+++ b/extension/package.json
@@ -30,7 +30,7 @@
     "@toruslabs/openlogin-jrpc": "^1.3.3",
     "advanced-css-reset": "^1.2.2",
     "async-mutex": "^0.3.2",
-    "bls-wallet-clients": "0.2.0",
+    "bls-wallet-clients": "0.5.2",
     "bls-wallet-signer": "0.6.1",
     "dotenv-webpack": "^7.0.3",
     "emoji-log": "^1.0.2",
--- a/extension/source/Controllers/Keyring/KeyringController.ts
+++ b/extension/source/Controllers/Keyring/KeyringController.ts
@@ -1,8 +1,4 @@
-import {
-  BlsWalletWrapper,
-  initBlsWalletSigner,
-  Operation,
-} from 'bls-wallet-clients';
+import { BlsWalletWrapper, Operation } from 'bls-wallet-clients';
 import generateRandomHex from '../../helpers/generateRandomHex';
 import BaseController from '../BaseController';
 import {
@@ -10,7 +6,7 @@ import {
   KeyringControllerConfig,
   KeyringControllerState,
 } from './IKeyringController';
-import { NETWORK_CONFIG, PROVIDER_URL } from '../../env';
+import { NETWORK_CONFIG, CHAIN_RPC_URL } from '../../env';
 import { ethers } from 'ethers';
 
 export default class KeyringController
@@ -61,10 +57,10 @@ export default class KeyringController
   }
 
   async signTransactions(address: string, tx: Operation) {
-    const signer = await this._getBLSSinger();
     const privKey = this._getPrivateKeyFor(address);
+    const signer = await this._getBLSSinger(privKey);
 
-    return signer.sign(tx, privKey);
+    return signer.sign(tx);
   }
 
   async _createAccountAndUpdate(privateKey: string): Promise<string> {
@@ -92,7 +88,7 @@ export default class KeyringController
   }
 
   private _getContractWalletAddress(privateKey: string) {
-    const provider = new ethers.providers.JsonRpcProvider(PROVIDER_URL);
+    const provider = new ethers.providers.JsonRpcProvider(CHAIN_RPC_URL);
     return BlsWalletWrapper.Address(
       privateKey,
       NETWORK_CONFIG.addresses.verificationGateway,
@@ -100,9 +96,12 @@ export default class KeyringController
     );
   }
 
-  private async _getBLSSinger() {
-    return initBlsWalletSigner({
-      chainId: Number(this.state.chainId),
-    });
+  private async _getBLSSinger(privateKey: string) {
+    const provider = new ethers.providers.JsonRpcProvider(CHAIN_RPC_URL);
+    return BlsWalletWrapper.connect(
+      privateKey,
+      NETWORK_CONFIG.addresses.verificationGateway,
+      provider,
+    );
   }
 }
--- a/extension/source/Controllers/Keyring/KeyringController.ts
+++ b/extension/source/Controllers/Keyring/KeyringController.ts
@@ -1,22 +1,29 @@
-import { Wallet } from 'ethers';
+import {
+  BlsWalletWrapper,
+  initBlsWalletSigner,
+  Operation,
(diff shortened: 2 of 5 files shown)


extension/source/Controllers/Keyring/KeyringController.ts
```diff
@@ -68,5 +60,49 @@ export default class KeyringController
...
+  }
+
+  private _getPrivateKeyFor(address: string): string {
+    const keyPair = this.state.wallets.find((x) => x.address === address);
+    if (!keyPair) throw new Error('key does not exist');
+    return keyPair.privateKey;
+  }
+
+  private _getContractWalletAddress(privateKey: string) {
+    const provider = new ethers.providers.JsonRpcProvider(PROVIDER_URL);
```
kautukkundan: Should this provider come from other controllers?
the `SafeEventEmitterProvider` used by other controllers isn't compatible with this

extension/package.json
```diff
@@ -30,7 +30,7 @@
     "@toruslabs/openlogin-jrpc": "^1.3.3",
     "advanced-css-reset": "^1.2.2",
     "async-mutex": "^0.3.2",
-    "bls-wallet-clients": "0.2.0",
+    "bls-wallet-clients": "0.5.2",
```
jacque006: This should now be `0.5.3` in `main`, this diff will be removed after @chaitanyapotti updates `feat/inpage-provider` with `main`.

extension/source/Controllers/Keyring/KeyringController.ts
jacque006: It would be better to use one ethers.js provider and pass that down, as this may create additional socket connections/network request overhead. For now though to get it working should be okay.

Also, you use `CHAIN_RPC_URL` instead of adding a new env var. `PROVIDER_URL === CHAIN_RPC_URL`.
```suggestion
    const provider = new ethers.providers.JsonRpcProvider(CHAIN_RPC_URL);
```

extension/source/env.ts
```diff
@@ -9,6 +9,7 @@ export const PRIVATE_KEY_STORAGE_KEY = requireEnv(
 export const AGGREGATOR_URL = requireEnv(process.env.AGGREGATOR_URL);
 export const CHAIN_RPC_URL = requireEnv(process.env.CHAIN_RPC_URL);
 export const CREATE_TX_URL = requireEnv(process.env.CREATE_TX_URL);
+export const PROVIDER_URL = requireEnv(process.env.PROVIDER_URL);
```
jacque006:
```suggestion
```

extension/source/Controllers/Keyring/KeyringController.ts
```diff
@@ -68,5 +60,49 @@ export default class KeyringController
...
+      NETWORK_CONFIG.addresses.verificationGateway,
+      provider,
+    );
+  }
+
+  private async _getBLSSinger() {
+    return initBlsWalletSigner({
+      chainId: Number(this.state.chainId),
+    });
+  }
```
jacque006: Instead of initing and handling signing via the BLS signer directly, you may just want to use the `BlsWalletWrapper` abstraction instead and add it into the state. I am unsure if the state needs to have only primitively typed data in it, if so you may be able to use another controller/construct to manage `BlsWalletWrapper` wallet objects.

You then would only need to call `BlsWalletWrapper.connect` once and then store the wallet object it returns and use that for signing/address operations. See https://github.com/jzaki/bls-wallet/tree/main/contracts/clients#blswalletwrapper for details.

voltrevo:
@kautukkundan I see you checked "I have manually tested these changes", but I don't understand what manual testing you could have done and you haven't written anything for that section. Could you clarify?

voltrevo (changes requested):
Looks sensible to me. I think you need to implement @jacque006's point about a shared ethers provider though and use `CHAIN_RPC_URL` instead of creating a new environment variable.

I'd also like @chaitanyapotti to give this a once-over since the target is his wip branch. I also imagine he'd be your best bet for doing the shared provider if there's any complications with that.

Once those things are taken care of let's get this merged.

extension/source/Controllers/Keyring/KeyringController.ts
kautukkundan: semi-fixed in f30cd85

Yeah I agree putting it in state and then just reusing the same object as the provider doesn't change.  
As of now I am creating a new object with a new provider for simplicity sakes. Would like to take @chaitanyapotti's opinion on how to implement it

extension/package.json
kautukkundan: resolved in f30cd85

kautukkundan:
ah @voltrevo manually tested on local terminal by importing modules and checking the returned values.

extension/source/env.ts
kautukkundan: resolved in f30cd85

extension/source/Controllers/Keyring/KeyringController.ts
kautukkundan: semi-fixed in f30cd85

Yeah I agree. As the prev comment mentions, should use the provider coming from the state. Couldn't because of incompatibility. Would be nice to get @chaitanyapotti's opinion on how to pass the provider to each controller.

Also @jacque006, the wrapper currently has a private constructor and static connect and address methods. It wont be possible to create an object (with only provider and verification address), save it in state and just use different privKeys to do different ops. WDYT about adding these to the client library itself?

jacque006: `connect` is essentially the async constructor for `BLSWalletWrapper` as it returns an initialized `BLSWalletWrapper` object. You would need to `connect` for each private key and store the wrapper object for each wallet.

I think rolling with the static methods for now is fine per our offline conversation. But it may be easier in the future to have a common place the `BLSWalletWrapper` object(s) can be accessed from.

extension/source/Controllers/Keyring/KeyringController.ts
```diff
@@ -68,5 +56,52 @@ export default class KeyringController
...
+  private _getContractWalletAddress(privateKey: string) {
+    const provider = new ethers.providers.JsonRpcProvider(CHAIN_RPC_URL);
+    return BlsWalletWrapper.Address(
+      privateKey,
+      NETWORK_CONFIG.addresses.verificationGateway,
+      provider,
+    );
+  }
+
+  private async _getBLSSinger(privateKey: string) {
```
jacque006:
```suggestion
  private async _getBLSWallet(privateKey: string): Promise<BLSWalletWrapper> {
```

extension/source/Controllers/Keyring/KeyringController.ts
```diff
@@ -68,5 +56,52 @@ export default class KeyringController
     }
   }
 
-  // TODO: add methods to sign transactions / messages
+  async signTransactions(address: string, tx: Operation) {
+    const privKey = this._getPrivateKeyFor(address);
+    const signer = await this._getBLSSinger(privKey);
+
+    return signer.sign(tx);
+  }
```
jacque006:
```suggestion
  async signTransactions(address: string, tx: Operation):Promise<Bundle> {
    const privKey = this._getPrivateKeyFor(address);
    const wallet = await this._getBLSWallet(privKey);

    return wallet.sign(tx);
  }
```

jacque006 (approved):
One other set of suggested changes, but I'm good :ship: ing this as is. We can figure out provider/wallet object issues in a later piece of work.

voltrevo:
> ah @voltrevo manually tested on local terminal by importing modules and checking the returned values.

Ohh nice 👍

voltrevo (approved):

extension/source/Controllers/Keyring/KeyringController.ts
chaitanyapotti: it should be the provider coming from other controllers.
it is an eip-1193 compatible provider.

chaitanyapotti:
ethers is a very heavy library.
Provider must satisfy an interface for bls-wallet-client instead of being an instance of ethers dependency

PR #88, opened 2022-01-10
Add welcome screen

### LINKED ISSUE #87: Create the welcome-to-quill screen

voltrevo:
## What is this PR doing?

Adds this screen:
Screen Shot 2022-01-10 at 6 08 36 pm

The next artwork slides in every 5 seconds. You can also click the radio buttons to jump to the corresponding artwork, which also prevents any further automated transitions.

Run the extension and view the above page and functionality.

Resolves #87.

## Checklist

## Guidelines


extension/source/Popup/Popup.tsx
```diff
@@ -17,6 +18,8 @@ type State = {
   appState?: AppState;
 };
 
+const useNewUI = true;
```
voltrevo, line 21: Note that you can go back to the old UI by setting this to false, for now.

extension/source/Popup/components/Carousel.tsx
```diff
@@ -0,0 +1,200 @@
...
+  };
+};
+
+const positionMap = {
+  left: 71 - 516,
+  middle: 71,
+  right: 71 + 516,
+};
+
+export default class Carousel extends React.Component<Props, State> {
```
kautukkundan: Should make this a function component. Most of the newer libraries are hook-only and it will be a good idea to follow the new pattern as it has several advantages as well.

For now we can leave the old components as is (as they are cross compatible to an extent)  and refactor them later but for any new one it will be a good idea to switch now.

kautukkundan (changes requested):
Some thoughts -

Since this is going to be a major UI overhaul, I suggest switching to the newer code patterns.

- The current react code has class based components. I suggest switching to function components. This will help in keeping the UI and logic separate by using hooks. The prop passing situation will also improve.

- For styling, instead of relying on individual stylesheets, I suggest switching to a theme engine based setup. Doing so will not only reduce the amount of code but also create consistency across the app. Tailwind will be a very nice choice.

- Finally, a minor one, 
  The app can be broken down into 3 main sections 
   1. The Popup
   2. Confirmation dialog
   3. Full screen app
So it will be a good idea to keep these things isolated. I guess this stuff is more or less already in place but could improve.  

Happy to discuss the above points!

extension/source/Popup/components/Carousel.tsx
voltrevo: Done 👍

voltrevo:
> Some thoughts -
> 
> Since this is going to be a major UI overhaul, I suggest switching to the newer code patterns.
> 
> * The current react code has class based components. I suggest switching to function components. This will help in keeping the UI and logic separate by using hooks. The prop passing situation will also improve.

Makes sense. I've converted the new components to be hook-based.

> * For styling, instead of relying on individual stylesheets, I suggest switching to a theme engine based setup. Doing so will not only reduce the amount of code but also create consistency across the app. Tailwind will be a very nice choice.

Does this just mean adding the engine usage? It sounds a bit like not having our own stylesheets or otherwise changing existing things. Not that change is bad, but I don't understand what you mean. Can you clarify?

> * Finally, a minor one,
>   The app can be broken down into 3 main sections
>   
>   1. The Popup
>   2. Confirmation dialog
>   3. Full screen app
>      So it will be a good idea to keep these things isolated. I guess this stuff is more or less already in place but could improve.

Could you be a bit more specific? If I had to guess - are you referring to the _quill.scss containing styles that should be in Popup/styles.scss? (And if not in scope for this PR, can we continue somewhere else?)

---

I think I've addressed everything that is actionable for this PR. Please re-review.

kautukkundan (approved):
Carousel component looks a bit complicated. `transitionImage` can itself be its own hook which can just be called inside the component. But Let's leave it like this for now and refactor it separately! 

Rest looks good to me 👍🏼

kautukkundan:
> Does this just mean adding the engine usage? It sounds a bit like not having our own stylesheets or otherwise changing existing things. Not that change is bad, but I don't understand what you mean. Can you clarify?

Basically adding tailwind and creating a custom custom config based on our design system. That way we can just create a simple central UI repository and pull styles from that without needing to rewrite css for each item. We would still have total control over the style but the overall design system will remain consistent across the app.

eg - 
we create a custom class like 

```css
.btn {
  @apply rounded-b-lg shadow-md text-lg font-bold text-gray-900;
}
```

and simply use `.btn` wherever needed. Note that the .btn class itself contains abstracted info which we set initially in the original config file.

more here - https://tailwindcss.com/docs/configuration
and here - https://tailwindcss.com/docs/functions-and-directives

-----

> Could you be a bit more specific? If I had to guess - are you referring to the _quill.scss containing styles that should be in Popup/styles.scss? (And if not in scope for this PR, can we continue somewhere else?)

I meant just tidying up the "source" directory a little bit as it has grown a bit. eg: having 5 folders - 
1. UI/Components - for the react components + styles
   1. Popup
   2. Confirmation
   3. Full Page
   4. Common
2. assets - for static items
3. scripts - content script/background script etc
3. hooks - for hooks/logic for components
4. controllers 

and nest things accordingly. Yeah let's do it separately. Also open to discussion about the structure.

voltrevo:
👍🙌

PR #92, opened 2022-01-12
Switch ./clients/deps/hubble-bls to @thehubbleproject/bls

jacque006:
## What is this PR doing?

Switching `./contracts/clients/deps/hubble-bls` to the `@thehubbleproject/bls` npm package.

- clients: `yarn test`
- contracts: `yarn hardhat test`
- aggregator: `./programs/premerge.ts`
- extension: run/smoke test extension

https://github.com/jzaki/bls-wallet/issues/9

## Checklist

## Guidelines

## TODO
- [x] Fix broken tests in clients
- [x] Fix crash in aggregator startup


contracts/clients/package.json
```diff
@@ -27,5 +27,9 @@
     "chai": "^4.3.4",
     "mocha": "^9.0.3",
     "source-map-support": "^0.5.20"
+  },
+  "resolutions": {
+    "ethers": "5.5.3",
+    "mcl-wasm": "1.0.0"
```
jacque006: https://github.com/thehubbleproject/hubble-bls/issues/24

contracts/clients/package.json
```diff
@@ -27,5 +27,9 @@
     "chai": "^4.3.4",
     "mocha": "^9.0.3",
     "source-map-support": "^0.5.20"
+  },
+  "resolutions": {
+    "ethers": "5.5.3",
```
jacque006: ~This is needed to make `deno` happy, it seems to have issues with multiple `ethers` versions being used. Will retest when other issues are resolved.~ Not needed

jacque006:
Two major issues blocking this now:

1. Most of the existing test cases fail due to new hex values being generated

`cd ./contracts/clients && yarn build && yarn test`
```sh
  index
    1) signs and verifies transaction
    2) aggregates transactions
    ✔ can aggregate transactions which already have multiple subTransactions (131ms)
    3) generates expected publicKeyStr


  1 passing (500ms)
  3 failing

  1) index
       signs and verifies transaction:

      AssertionError: expected [ Array(2) ] to deeply equal [ Array(2) ]
      + expected - actual

       [
      -  "0x0f8af80a400b731f4f2ddcd29816f296cca75e34816d466512a703631de3bb69"
      -  "0x023d76b485531a8dbc087b2d6f25563ad7f6d81d25f5f123186d0ec26da5e2d0"
      +  "0x0058b38298f3c486223de7c61f461ff3b47530d2619a383e49b298a56249e4fb"
      +  "0x0bf8beb03979073e57656af0a5bab043fe2d6bf4cbbf600f6a7190ce95fcf69c"
       ]
      
      at Context.<anonymous> (test/index.test.ts:59:38)

  2) index
       aggregates transactions:

      AssertionError: expected [ Array(2) ] to deeply equal [ Array(2) ]
      + expected - actual

       [
      -  "0x0008678ea56953fdca1b007b2685d3ed164b11de015f0a87ee844860c8e6cf30"
      -  "0x2bc51003125b2da84a01e639c3c2be270a9b93ed82498bffbead65c6f07df708"
      +  "0x091727df1b9834b31111c1d5c1e15989350de678232e46898c1c3c788e3c26ab"
      +  "0x1f69e3373d329564e875a1401c0b7a4a6f7ab3e9cf93ef73b59a1feb3286e693"
       ]
      
      at Context.<anonymous> (test/index.test.ts:105:39)

  3) index
       generates expected publicKeyStr:

      AssertionError: expected '0x027c3c0483be2722a29a0229bef64b2d8c1f8d4e954b0203d01ce342608b6eb8060c1136ac3aef9ba4b2a0272920fba5528f6e97b376c511bc746e47414a0d0411a697990758be620b0f09d3ad5bebe359964b74a29b89bdeb60671d243997fa2075298b12a51948b7a40dc69bdc91698ed95e5d2a69d04845af64aaf2eb1537' to equal '0x0127eaaf599e0e5997ebff004ae96b61afef4c35d9d67277ebd32a08bc42e2eb01dc98a8d2e5ac73933935dbb2888c6719914e86f14de30bf441b951867e27cd2f29d73e617ee30597f79b3fe567a11eb3bea1a2625ccb3fcb9635edf1d2f42927fac8b975dd299618e7f48753fdfa492f11f701445cdd7c3ca98bf5c386ec65'
      + expected - actual

      -0x027c3c0483be2722a29a0229bef64b2d8c1f8d4e954b0203d01ce342608b6eb8060c1136ac3aef9ba4b2a0272920fba5528f6e97b376c511bc746e47414a0d0411a697990758be620b0f09d3ad5bebe359964b74a29b89bdeb60671d243997fa2075298b12a51948b7a40dc69bdc91698ed95e5d2a69d04845af64aaf2eb1537
      +0x0127eaaf599e0e5997ebff004ae96b61afef4c35d9d67277ebd32a08bc42e2eb01dc98a8d2e5ac73933935dbb2888c6719914e86f14de30bf441b951867e27cd2f29d73e617ee30597f79b3fe567a11eb3bea1a2625ccb3fcb9635edf1d2f42927fac8b975dd299618e7f48753fdfa492f11f701445cdd7c3ca98bf5c386ec65
      
      at Context.<anonymous> (test/index.test.ts:171:52)
```

I am unsure when/what version of the `hubble-bls` repo was copied to `./contracts/clients/deps/hubble-bls` to see what may be causing this difference.

2. Aggregator fails to start due to missing global `document`

`cd ./aggregatgor && ./programs/aggregator.ts`
```sh
error: Uncaught (in promise) ReferenceError: document is not defined
    at https://cdn.esm.sh/v62/mcl-wasm@0.7.9/deno/mcl-wasm.js:2:2798
    at Object.EQ.A.init (https://cdn.esm.sh/v62/mcl-wasm@0.7.9/deno/mcl-wasm.js:2:532289)
    at Object.Pe [as init] (https://cdn.esm.sh/v62/@thehubbleproject/bls@0.4.0/deno/bls.js:2:3702)
    at Function.new (https://cdn.esm.sh/v62/@thehubbleproject/bls@0.4.0/deno/bls.js:2:8567)
    at b5 (https://cdn.esm.sh/v62/bls-wallet-clients@0.5.4-c47077f/deno/bls-wallet-clients.js:2:5578)
    at Function.create (file:///home/jacque006/workspaces/bls-wallet/aggregator/src/app/EthereumService.ts:60:35)
    at async app (file:///home/jacque006/workspaces/bls-wallet/aggregator/src/app/app.ts:30:27)
    at async file:///home/jacque006/workspaces/bls-wallet/aggregator/programs/aggregator.ts:6:1
```

@voltrevo Any ideas or help you could provide would be appreciated.

contracts/clients/package.json
jacque006: Not explicitly needed, removing for now.

contracts/clients/package.json
```diff
@@ -16,9 +16,12 @@
     "publish-experimental": "node scripts/showVersion.js >.version && npm version $(node scripts/showBaseVersion.js)-$(git rev-parse HEAD | head -c7) --allow-same-version && npm publish --tag experimental && npm version $(cat .version) && rm .version",
     "publish-experimental-dry-run": "node scripts/showVersion.js >.version && npm version $(node scripts/showBaseVersion.js)-$(git rev-parse HEAD | head -c7) --allow-same-version && npm publish --tag experimental --dry-run && npm version $(cat .version) && rm .version"
   },
+  "resolutions": {
+    "mcl-wasm": "1.0.0"
+  },
```
jacque006: @jacque006 Remove with new `@thehubbleproject/bls` release.

contracts/clients/test/index.test.ts
```diff
@@ -57,8 +54,8 @@ describe("index", () => {
     const bundle = sign(bundleTemplate, privateKey);
 
     expect(bundle.signature).to.deep.equal([
-      "0x0058b38298f3c486223de7c61f461ff3b47530d2619a383e49b298a56249e4fb",
-      "0x0bf8beb03979073e57656af0a5bab043fe2d6bf4cbbf600f6a7190ce95fcf69c",
+      "0x0f8af80a400b731f4f2ddcd29816f296cca75e34816d466512a703631de3bb69",
+      "0x023d76b485531a8dbc087b2d6f25563ad7f6d81d25f5f123186d0ec26da5e2d0",
     ]);
 
     expect(verify(bundle)).to.equal(true);
```
jacque006, line 61: Verifying the bad sign below is now hitting the same error as in PR https://github.com/thehubbleproject/hubble-bls/pull/20

jacque006: This was due to a malformed `otherPrivateKey` value. Using a hex generated by `openssl` fixes the error being thrown, see https://github.com/jzaki/bls-wallet/pull/92/files#diff-5d32e4c02bf142397e78ddf9b0f9755aa9214b461919f99ace968a7f6ba965f3R35-R36

contracts/clients/package.json
jacque006: Will do this in a follow up PR or if a new version of `@thehubbleproject/bls` is released while this is in flight.

aggregator/src/helpers/Rng.ts
```diff
@@ -1,18 +1,14 @@
...
-
 function byteToHex(byte: number): string {
   return byte.toString(16).padStart(2, "0");
 }
 
 function hash(data: string) {
   const byteValues = Array.from(new TextEncoder().encode(data));
-  const hex = `0x${byteValues.map(byteToHex).join("")}`;
-
-  return keccak256(hex);
```
jacque006: This was causing our favorite [err _wrapInput](https://github.com/thehubbleproject/hubble-bls/pull/20) error in tests when generating a random private key. Note that the way it is setup now closely mirrors how the extension generates its private keys.

contracts/clients/package.json
```diff
@@ -17,15 +17,15 @@
     "publish-experimental-dry-run": "node scripts/showVersion.js >.version && npm version $(node scripts/showBaseVersion.js)-$(git rev-parse HEAD | head -c7) --allow-same-version && npm publish --tag experimental --dry-run && npm version $(cat .version) && rm .version"
   },
   "dependencies": {
-    "ethers": "^5.5.2",
-    "mcl-wasm": "0.7.6",
-    "typescript": "^4.3.5"
```
jacque006, line 22: `typescript` should only be needed as a dependency to run type checking and transpilation, moved to dev deps.

contracts/clients/src/signer/defaultDomain.ts
```diff
@@ -1,4 +1,3 @@
-import { arrayify } from "@ethersproject/bytes";
-import { keccak256 } from "@ethersproject/keccak256";
```
jacque006, line 2: We may end up wanting to move these back to more specific imports depending on what we decide in https://github.com/jzaki/bls-wallet/issues/109

contracts/clients/test/index.test.ts
```diff
@@ -57,8 +55,8 @@ describe("index", () => {
     const bundle = sign(bundleTemplate, privateKey);
 
     expect(bundle.signature).to.deep.equal([
-      "0x0058b38298f3c486223de7c61f461ff3b47530d2619a383e49b298a56249e4fb",
-      "0x0bf8beb03979073e57656af0a5bab043fe2d6bf4cbbf600f6a7190ce95fcf69c",
+      "0x0f8af80a400b731f4f2ddcd29816f296cca75e34816d466512a703631de3bb69",
+      "0x023d76b485531a8dbc087b2d6f25563ad7f6d81d25f5f123186d0ec26da5e2d0",
```
jacque006, line 58: `mcl-wasm@1.0.0` generates new hexes in these tests. They are still valid in contract and aggregator tests.

contracts/test/recovery-test.ts
```diff
@@ -54,7 +52,12 @@ describe("Recovery", async function () {
   });
 
   it("should update bls key", async function () {
-    const newKey: PublicKey = [1, 2, 3, 4].map(BigNumber.from);
+    const newKey: PublicKey = [
+      BigNumber.from(1),
+      BigNumber.from(2),
+      BigNumber.from(3),
+      BigNumber.from(4),
+    ];
```
jacque006, line 75: New Typescript version caught this. Is there a better way we could to this?

aggregator/programs/premerge.ts
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env -S deno run --unstable --allow-run --allow-read --allow-write
+#!/usr/bin/env -S deno run --unstable --allow-run --allow-read --allow-write --allow-env
```
jacque006, line 1: `--allow-env` now appears to be required for most scripts. Unsure what change did this, looks to be Deno std lib related.

aggregator/deps.ts
```diff
@@ -40,17 +40,17 @@ export type {
   PublicKey,
   Signature,
   VerificationGateway,
-} from "https://esm.sh/bls-wallet-clients@0.5.3";
+} from "https://esm.sh/bls-wallet-clients@0.5.4-05c23aa";
```
jacque006: With the rebase workflow I prefer to use to consolidate commits, this will never match the merged commit unless I append it afterwards.

@voltrevo You've had some confusion in the past based on this, thoughts?

voltrevo: Mmm. I understand the issue you're referring to but not the preference you're indicating above.

I think it makes sense to republish with your final changes anyway (why use a commit that represents an incomplete PR?), but let's have a chat to straighten this out.

voltrevo: Just realized you can't use the final commit if you're trying to use that commit id to update the version being used. Yeah let's chat.

aggregator/src/helpers/Rng.ts
voltrevo: But... that's not a hash now. The whole idea of `Rng` is to provide reproducible random numbers for testing and `keccak256` is the core of that. (Which could be documented better, that's on me.)

I'm surprised that this fixes the issue. I suppose it's by chance? Unless there's something more to it that I don't understand.

It seems like the issue is that hubble-bls has requirements (valid or otherwise) about what a private key can be. A hex string of the right length isn't enough. I think we need to address that directly.

contracts/clients/package.json
voltrevo:
```diff
-This
+typescript
```

👍... for a minute there I was like... why is ethers a dev dependency? 🤯

contracts/clients/test/index.test.ts
voltrevo: Weird! But ok :).

contracts/test/recovery-test.ts
voltrevo: > New Typescript version caught this.

I don't think that's what happened :/. The issue is that the type of `.map` works on arrays, not tuples, so using `.map` converts your tuple into an array, so the assignment absolutely shouldn't work. This hasn't changed, I checked back to 3.3.3 and it still 'correctly' rejects constructions like this:

https://www.typescriptlang.org/play?ts=3.3.3#code/C4TwDgpgBACg0lAvFA2gOwK4FsBGEBOANFJrgcaXkSdlQLoDcAUEwGYZoDGwAlgPZoowPgDlaBABQBnYPgBcUGfh5oA5gEoFlAlADeTKIaj4IwDPkFiy+abPXMAvi04CZUMAGsF8JKgDkAIx+xH4ATMFQfgDMEX4ALH50AHRYAIZgEsJWVPZMQA

I'm curious how this happened. Any ideas?

...the presence of CI would have been useful here :(

Anyway, in terms of a better way, I don't think you can. I mean, I think it's possible to write a `map` function that is tuple-aware, but I don't think it's worth doing here.

voltrevo (changes requested):
A few things here.

I also tested this with the extension (`const useNewUI = false;` in `Popup.tsx`) and I'm getting the `err _wrapInput` issue. Need to fix that before this can go ahead.

On the bright side, I did manually switch to a valid private key value and was able to confirm that the transfers do still work even though we're generating different signatures. Wild.

contracts/test/recovery-test.ts
jacque006: Right, we could do `...map<PublicKey>(...)` or something similar, but better to be explicit.

I attempted to update your example to more closely match a `BigNumber` [example](https://www.typescriptlang.org/play?ts=3.3.3#code/C4TwDgpgBAQglgcwHIFcC2AjCAnOBnACygF5ZFVMcoAfKDROAO2Bqj2F0YVcfS2wBQoSFAAKAaRJQA2vGR8c+AgBoy8yrkKq5Ffku3kFmggF0A3AIEBjADYBDPHjW6qAbwFRPUKwHtG7bBQrYB9sAAoPLyiwFAwbOCsobAg7ABM-GxAoADc7GxQIAC4oXg1IzwBKKFcAX0to2PjE9jtgBKgAM2wfNDDcm2K7RhAK4p0javKoqBsIFgxGADU84tL+KSQ7JAtprzgOqDDhCB8D-pJiUgByAKYEK6r3Xd2F5ZspMDtsPAgASWY+nkKjtnjUoBAbD8oPtDsdTjk8hdrmscA9Js9pq9EaR+iDdnUplFksAUNhGCUIAB3ZxGMJYmzAqZ1Am+fwsMAAa2KEik0iuAEYrqorgAmIVQK4AZnFVwALFcTAA6NB2MBhcYaRVdHqMoA) but still get the same error. Unsure what changed to now get that error.

+1 Getting some CI processes in place, can prioritize after current round of work.

jacque006:
@voltrevo I assumed the extension would Just Work :tm: given the other changes, good catch. Will triage and fix.

aggregator/src/helpers/Rng.ts
jacque006: https://github.com/thehubbleproject/hubble-bls/issues/27 , investigating.

jacque006: Fixed with `@thehubbleproject/bls@0.5.1`, reverted this change back to original code.

aggregator/deps.ts
jacque006: @voltrevo I tried our discussed process of pinning experimental version in last commit, let me know how it looks.

contracts/clients/src/BlsWalletWrapper.ts
```diff
@@ -164,6 +167,11 @@ export default class BlsWalletWrapper {
     return this.blsWalletSigner.sign(operation, this.privateKey);
   }
 
+  /** Sign a message */
+  signMessage(message: string): Signature {
+    return this.blsWalletSigner.signMessage(message, this.privateKey);
+  }
```
jacque006, line 173: This seemed appropriate to expose, as someone may need to use it during the wallet recovery/upgrade process.

contracts/package.json
```diff
@@ -38,7 +36,7 @@
     "eslint-plugin-prettier": "^3.4.0",
     "eslint-plugin-promise": "^5.1.0",
     "ethereum-waffle": "^3.0.0",
-    "ethers": "^5.5.1",
+    "ethers": "5.5.4",
```
jacque006, line 39: `ethers@5.6.0` makes everything very sad, explicitly pinned `5.5.4` until someone can investigate upgrade path, likely in https://github.com/jzaki/bls-wallet/issues/160

jacque006:
@voltrevo Needed changes made, fully retested. Ready for your re-review.

For the extension, I made sure the old UI path properly generated a key but was unable to confirm the txn send as that path is now tied with extension dApp work.

voltrevo (approved):

PR #93, opened 2022-01-12
Rough onboarding flow part 1

voltrevo:
## What is this PR doing?

Adds a rough version of the onboarding ui with very little styling:

Screen Shot 2022-01-12 at 5 22 03 pm
Screen Shot 2022-01-12 at 5 22 09 pm
Screen Shot 2022-01-12 at 5 22 13 pm
Screen Shot 2022-01-12 at 5 22 16 pm

## How can these changes be manually tested?

Click the `Get Started →` button on the popup to view the screens above (click the numbers).

## Does this PR resolve or contribute to any issues?

Contributes to #90.

## Checklist
- [x] I have manually tested these changes
- [x] Post a link to the PR in the group chat

## Guidelines
- If your PR is not ready, mark it as a draft
- The `resolve conversation` button is for reviewers, not authors
  - (But add a 'done' comment or similar)

PR #95, opened 2022-01-13
Rough onboarding flow part 2

### LINKED ISSUE #90: Rough set-up flow
Don't try to match the design yet because that depends on getting @kautuk's engine stuff working.

voltrevo:
# _Dependent PR_

This PR builds on top of #93. Please review and merge that first.

## What is this PR doing?

Finishes the (mostly) unstyled onboarding flow by adding the confirmation phase for the secret words.

Click through the review phase.

Resolves #90.

## Checklist

## Guidelines


extension/source/QuillPage/components/OnboardingActionPanel.tsx
```diff
@@ -13,17 +13,30 @@ const OnboardingActionPanel: React.FunctionComponent<{ pageIndex: number }> = ({
...
     {
       [
-        <PasswordCreationPanel key={1} />,
-        <SetNicknamePanel key={2} />,
-        <ViewSecretPhrasePanel key={3} />,
+        <PasswordCreationPanel
+          key={1}
+          onComplete={() => {
+            window.location.href = `?p=2`;
+          }}
```
kautukkundan, line 20:
```suggestion
          next={2}
          }}
```
How about instead of passing a whole function as prop, just pass in the next page number and handle onComplete inside the target component? Prop passing should be kept as light as possible and individual components should remain as independent as possible.

extension/source/QuillPage/components/ReviewSecrectPhrasePanel.tsx
```diff
@@ -0,0 +1,100 @@
+import * as React from 'react';
```
kautukkundan:
```diff
-ReviewSecrectPhrasePanel.tsx
+ReviewSecretPhrasePanel.tsx
```
Typo in the file name

extension/source/components/Button.tsx
```diff
@@ -20,6 +21,10 @@ const Button = (props: {
     classes.push('loading');
   }
 
+  if (props.disabled) {
+    classes.push('disabled');
+  }
+
   return (
     <div
       className={classes.join(' ')}
```
kautukkundan:
```suggestion
      className={`button ${highlight && "highlight"} ${loading && "loading"} ${disabled && "disabled"}`}
```

Will keep this component stateless

extension/source/QuillPage/components/ViewSecretPhrasePanel.tsx
```diff
@@ -47,7 +35,7 @@ const ViewSecretPhrasePanel: React.FunctionComponent = () => {
           <QuickRow key={`row${i}`}>
             {Range(3).map((j) => (
               <QuickColumn key={`column${j}`}>
-                {3 * i + j + 1}. {secretWords[3 * i + j]}
+                {3 * i + j + 1}. {secretPhrase[3 * i + j]}
```
kautukkundan: 👍🏼

kautukkundan (changes requested):
Minor Changes

extension/source/QuillPage/components/OnboardingActionPanel.tsx
voltrevo: I think this is the same point as https://github.com/jzaki/bls-wallet/pull/93#discussion_r784374486. If so, let's followup there.

extension/source/QuillPage/components/ReviewSecrectPhrasePanel.tsx
voltrevo: Oops. Fixed 👍.

extension/source/components/Button.tsx
voltrevo: Similar to before, I don't like this because it inserts `false`.

I think this pattern comes up a lot though, and I agree that the `classes.join` approach is a bit verbose. How about a utility like this?

```ts
export default function classFlags(classes: Record<string, boolean>) {
  return Object.entries(classes)
    .filter(([, flag]) => flag)
    .map(([className]) => className)
    .join(' ');
}
```

That way we could write:

```tsx
<div className={classFlags({ button: true, highlight, loading, disabled })}/>
```

kautukkundan: This does look fine.
Also there's a popular one called [classnames](https://www.npmjs.com/package/classnames).

but again, it works only for this component as its static and the local variable doesn't change after this is rendered. In other components, conditional styles must set via state mutation and not local variable mutation.

Reason : 
If the component's state changes, only the part where the state is consumed is re-rendered. However, in case local vars change in a component, ENTIRE component along with ALL the children are re-rendered as react sees it as a brand new component

voltrevo: > In other components, conditional styles must set via state mutation and not local variable mutation.

I'm not sure why but I'm not able to follow what you're getting at here. Maybe an example where something similar to what I'm doing goes badly would help?

kautukkundan (approved):
LGTM 👍🏼

extension/source/components/Button.tsx
kautukkundan: apologies for not being clear, I meant that
state variables = variables declared using useState method
local variables = any other variable declared in the body of the component

Also I played around in a fresh project and I think that the reason I gave is wrong. Instead of re-rendering every child component, it wont do anything if a local variable is mutated. 

Btw this doesn't apply in this particular case anymore, so I guess let's come back to this later if such a case arises again.

voltrevo: 👍

PR #102, opened 2022-01-14
Make password creation form functional

### LINKED ISSUE #97: Add functionality to the password creation form
- [ ] Use the password measurement from #96.
- [ ] Define thresholds for `Weak`, `Average`, `Good`, `Strong` (or similar) and display them.
- [ ] In lieu of a strength bar, add some text that displays the fill percentage that should be used when it becomes available.
- [ ] Check the passwords match, add a className for the correct and incorrect states. Also prefer a className for incomplete for non-empty entries that are a substring of the password.

### LINKED ISSUE #98: Prevent clicking ahead when the password form isn't done

voltrevo:
# _Dependent PR_

This PR builds on top of https://github.com/jzaki/bls-wallet/pull/95. Please review and merge that one first.

## What is this PR doing?

Adds functionality to the password creation form:
- Displays password strength using method from previous work
  - Shows weak/strong/etc
  - Description elaborating on weak/strong/etc
  - Horizontal bar with a width corresponding to granular strength measurement
- Confirm password box shows whether it is correct/incorrect/incomplete
- Continue button is disabled if a password hasn't been entered or the confirm field doesn't match

I included some naive css this time because I couldn't help it. Just things like `background-color: green` though. Intended to be thrown away.

Screen Shot 2022-01-14 at 3 59 08 pm

Visit the password creation form and verify the above.

Resolves #97, resolves #98.

## Checklist

## Guidelines

### DIFF
--- a/extension/source/QuillPage/components/PasswordCreationForm.tsx
+++ b/extension/source/QuillPage/components/PasswordCreationForm.tsx
@@ -33,12 +33,12 @@ const PasswordCreationForm: React.FunctionComponent<{
       setConfirmPasswordFieldValue(newConfirmPassword);
     }
 
-    const newResultPasswd =
+    const newResultPassword =
       newPassword === newConfirmPassword ? newPassword : undefined;
 
-    if (newResultPasswd !== password) {
-      setPassword(newResultPasswd);
-      onPasswordUpdate(newResultPasswd);
+    if (newResultPassword !== password) {
+      setPassword(newResultPassword);
+      onPasswordUpdate(newResultPassword);
     }
   }
 
--- a/extension/source/QuillPage/components/PasswordCreationForm.tsx
+++ b/extension/source/QuillPage/components/PasswordCreationForm.tsx
@@ -18,9 +18,6 @@ const passwordCommentaryMap: Record<PasswordStrength['descriptor'], string> = {
 const PasswordCreationForm: React.FunctionComponent<{
   onPasswordUpdate: (password: string | undefined) => void;
 }> = ({ onPasswordUpdate }) => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-expressions
-  onPasswordUpdate;
-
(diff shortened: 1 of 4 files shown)


extension/source/QuillPage/components/PasswordCreationForm.tsx
```diff
@@ -1,27 +1,113 @@
...
+
+  if (confirmPasswordFieldValue === '') {
+    confirmFieldClasses.push('empty');
+  } else if (confirmPasswordFieldValue === passwordFieldValue) {
+    confirmFieldClasses.push('correct');
+  } else if (passwordFieldValue.startsWith(confirmPasswordFieldValue)) {
+    confirmFieldClasses.push('incomplete');
+  } else {
+    confirmFieldClasses.push('incorrect');
+  }
```
jacque006, line 57: To prevent redundant computation in render cycles and prevent subcomponents from unnecessarily re-rendering, generating values like this can be memoized. https://reactjs.org/docs/hooks-reference.html#usememo

```suggestion
  const confirmFieldClasses = React.useMemo(() => {
    const classes = ['password-confirm-field'];
    if (confirmPasswordFieldValue === '') {
      classes.push('empty');
    } else if (confirmPasswordFieldValue === passwordFieldValue) {
      classes.push('correct');
    } else if (passwordFieldValue.startsWith(confirmPasswordFieldValue)) {
      classes.push('incomplete');
    } else {
      classes.push('incorrect');
    }
    return classes;
  }, [confirmPasswordFieldValue, passwordFieldValue]);
```
Now `confirmFieldClasses` will only become a new `Array` object when either `confirmPasswordFieldValue` or `passwordFieldValue` changes.

I don't think this is needed for this round of changes, but something to keep in mind.

jacque006 (approved):

voltrevo:
This was merged before #95. Oops.

PR #103, opened 2022-01-17
Kk/theme engine

kautukkundan:
## What is this PR doing?
- Adds tailwind css config
- replaces styling of WelcomeScreen component to use tailwind

- build and import extension into chrome 

## Checklist

## Guidelines


extension/tailwind.config.js
```diff
@@ -0,0 +1,85 @@
+module.exports = {
```
kautukkundan, line 1: Added only required config for current scope. Will add more preset styles based on the Quill Design system as required.

extension/source/components/Button.tsx
```diff
@@ -1,52 +1,26 @@
+import { IconProps } from 'phosphor-react';
 import * as React from 'react';
 
 const Button = (props: {
+  className?: string;
```
kautukkundan, line 5: I realized that the same button component will be reused with variation at different places so I moved the styling outside such that it is implemented in parent where button is used.

extension/source/components/Button.tsx
```diff
@@ -1,52 +1,26 @@
...
   onPress: () => void;
-  highlight?: boolean;
+  highlight?: boolean; //TODO - remove dependency
```
jacque006, line 7: What dependency needs to be removed for this? Would be nice to document in this comment.

jacque006 (approved):
:+1: Nice simple start

extension/source/components/Button.tsx
voltrevo: https://github.com/jzaki/bls-wallet/issues/110

voltrevo (approved):

extension/source/components/Button.tsx
kautukkundan: added issue description 👍🏼

PR #114, opened 2022-01-25
Prevent setting non-gateway in setTrustedGateway

### LINKED ISSUE #107: 5. Verification gateway could accidentally be set to non-conformant contract
The function _setTrustedBLSGateway in VerificationGateway.sol allows a wallet
user to migrate to a new verification gateway. The function checks that the address supplied
as a parameter is a smart contract rather than an externally owned account, but does not
verify that the contract supports the required interface. This may lead to accidentally
migrating to a smart contract that does not support verification gateway functionality.
Recommendation
Consider implementing interface (https://eips.ethereum.org/EIPS/eip-165). detection, for example EIP-165

voltrevo:
## What is this PR doing?

Adds `isVerificationGateway() -> true` and uses it to prevent accidentally upgrading the gateway to something that isn't a gateway.

`yarn hardhat test` (See test that has been added)

Resolves #107.

## Checklist

## Guidelines

### DIFF
--- a/contracts/contracts/VerificationGateway.sol
+++ b/contracts/contracts/VerificationGateway.sol
@@ -267,6 +267,13 @@ contract VerificationGateway
     }
 
     /**
+    Declare that this is a verification gateway. See BLSWallet#setTrustedGateway.
+     */
+    function isVerificationGateway() public pure returns (bool) {
+        return true;
+    }
+
+    /**
     Create a new wallet if not found for the given bls public key.
      */
     function createNewWallet(
--- a/contracts/contracts/VerificationGateway.sol
+++ b/contracts/contracts/VerificationGateway.sol
@@ -8,6 +8,7 @@ import "@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol";
 import "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
 
 import "./interfaces/IWallet.sol";
+import "./interfaces/IVerificationGateway.sol";
 
 /**
 A non-upgradable gateway used to create BLSWallets and call them with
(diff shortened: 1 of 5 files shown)


contracts/contracts/VerificationGateway.sol
```diff
@@ -266,6 +267,13 @@ contract VerificationGateway
         }
     }
 
+    /**
+    Declare that this is a verification gateway. See BLSWallet#setTrustedGateway.
+     */
+    function isVerificationGateway() external pure returns (bool) {
+        return true;
+    }
```
jacque006, line 275: I would recommend using the [EIP-165](https://eips.ethereum.org/EIPS/eip-165) spec for this sort of check. `OpenZepplin` should have all the needed contracts/utils at https://docs.openzeppelin.com/contracts/4.x/api/utils#IERC165 . This will also allow 3rd parties a standardized way to check if a contract is a verification gateway.

In the future, this could also be worth doing for `BLSWallet` as well.

jacque006 (commented):
Otherwise looks good :+1:

voltrevo:
Marked as draft because we expect this to be superseded by adding some more meaningful checks for https://github.com/jzaki/bls-wallet/issues/105

contracts/contracts/VerificationGateway.sol
voltrevo: Hmm doing eip165 was the original plan actually. But we decided against it because the whole idea of changing gateways is that the new gateway is different. There aren't any functions on a gateway which truly define what a gateway is, because the defining feature is that the wallet allows calls from the gateway.

Therefore, we wanted to be able to signal that the gateway is a gateway without putting any additional constraints on what a gateway needs to be.

Having said that, this is probably moot, see https://github.com/jzaki/bls-wallet/pull/114#issuecomment-1023738629.

voltrevo:
Obsoleted by https://github.com/jzaki/bls-wallet/pull/163

PR #115, opened 2022-01-25
Test atomicity of operations and actions

### LINKED ISSUE #106: Confirm atomic wallet calls, not whole transaction
Audit issue 4. "A single operation causes all other operations in an execution cycle to fail"
"Should one of these calls fail, the whole transaction reverts"

Show this is not the case with a unit test.

voltrevo:
## What is this PR doing?

Adds tests:
- should allow other operations when one fails
- should prevent other actions within an operation when one fails

Run the new tests.

Resolves #106.

## Checklist

## Guidelines

### DIFF
--- a/contracts/shared/helpers/TokenHelper.ts
+++ b/contracts/shared/helpers/TokenHelper.ts
@@ -3,12 +3,13 @@ import { utils, BigNumber, Signer, Contract } from "ethers";
 import { BlsWalletWrapper } from "../../clients/src";
 
 import Fixture from "./Fixture";
+import { IERC20 } from "../../typechain";
 
 export default class TokenHelper {
   static readonly initialSupply = utils.parseUnits("1000000");
   readonly userStartAmount: BigNumber;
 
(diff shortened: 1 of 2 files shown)


contracts/shared/helpers/TokenHelper.ts
```diff
@@ -56,7 +57,7 @@ export default class TokenHelper {
   async walletTokenSetup(): Promise<BlsWalletWrapper[]> {
     const wallets = await this.fx.createBLSWallets();
 
-    this.testToken = await TokenHelper.deployTestToken();
+    this.testToken = (await TokenHelper.deployTestToken()) as IERC20;
```
jacque006, line 60: You could modify [TokenHelper.deployTestToken](https://github.com/jzaki/bls-wallet/blob/main/contracts/shared/helpers/TokenHelper.ts#L22-L29) to use the Typechain factory and contract type instead of doing the type cast here.

jacque006 (approved):

PR #117, opened 2022-01-28
Avoid redundant calculations

### LINKED ISSUE #108: 6. Hash is calculated twice in transaction processing
Review calculations. Not as relevant for L2.

The function processBundle in VerificationGateway.sol always calls
createNewWallet to either find an existing wallet or create a new one. This function
calculates the hash of the public BLS key. The same calculation is performed again just after
the function call, causing the same hash to be calculated twice in short succession, which is
inefficient.
Recommendation
We recommend refactoring the code to perform the calculation once to reduce gas
consumption.

voltrevo:
## What is this PR doing?

Avoids some redundant calculations in `processBundle`:
- Duplicate calculation of `publicKeyHash`
- Duplicate call to `walletFromHash`

It's also a bit simpler this way, imo.

Run the tests.

I also tried logging the gas used in `should send ETH (empty call)` and found this change reduced the gas from ~300k to ~280k.

Resolves #108.

## Checklist

## Guidelines

### DIFF
--- a/contracts/contracts/VerificationGateway.sol
+++ b/contracts/contracts/VerificationGateway.sol
@@ -237,8 +237,7 @@ contract VerificationGateway
         successes = new bool[](opLength);
         results = new bytes[][](opLength);
         for (uint256 i = 0; i<opLength; i++) {
-            // create wallet if not found
-            wallet = createNewWallet(bundle.senderPublicKeys[i]);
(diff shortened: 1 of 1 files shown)


jacque006 (approved):

jacque006:
@voltrevo Should this be targeting/comparing against `audit-feedback` instead?

jzaki (approved):

PR #136, opened 2022-02-10
Require rewards

### LINKED ISSUE #133: Require rewards (configurable)
Currently the aggregator merely checks whether transactions succeed. Make it also check that it gets rewarded.

The required reward needs to depend both on the size of the transaction and the estimated gas, so there should be configuration parameters for both of these.

voltrevo:
## What is this PR doing?

- Tarball `ts` in build directory to prevent interfering with subsequent builds
- Adds environment variables for configuring rewards:
  - `FEE_TYPE`: `ether` if using eth, otherwise `token:address`
  - `FEE_PER_GAS`: the required reward for each gas unit in wei
  - `FEE_PER_BYTE`: as above but per byte of call data
- Changes the way submission bundles are assembled to ensure they meet the reward requirement
  - Includes a bisect strategy when necessary

`./programs/premerge.sh --env local`

Resolves #133.

## Checklist

## Guidelines


jacque006 (approved):
Solid work :+1: 

Only minor thing I can think of is documenting the new env vars in `./aggregator/README.md`.
>    FEE_TYPE: ether if using eth, otherwise token:address
>    FEE_PER_GAS: the required reward for each gas unit in wei
>    FEE_PER_BYTE: as above but per byte of call data

@jzaki Likely worth a pass from you as well if you have time.

PR #137, opened 2022-02-11
Require rewards - partial: Updates to contracts and clients

voltrevo:
## What is this PR doing?

This PR contributes to the larger work to require rewards in the aggregator.

That work is built on top of these changes to contracts, and clients - most importantly the new `Utilities` contract and its bindings.

That's not everything though, so here's a list:
- Add `Utilities` contract with two methods
  - `performSequence` which takes a sequence of functions to call and returns all their results. This will allow the aggregator to test a sequence of bundles and figure out the rewards for each bundle resulting from executing in that sequence.
  - `ethBalanceOf` returns the eth balance of a contract to allow this to be represented as a function call so you can use it with `performSequence`. (For tokens, you can use the function call already, but there isn't one for eth).
- Export ERC20 typechain definitions
  - Also use this to replace the abi copies in aggregator project and manual MockErc20 wrapper
- Implement `callStaticSequence` and `callStaticSequenceWithMeasure` in `EthereumService` which provides an additional layer around `performSequence` allowing decoded types to be extracted (rather than the byte arrays that are actually returned by `performSequence`
- Fix handling of empty bundles when aggregating and verifying (hubbleBls rejects zero signatures, but they ought to be valid and are useful when building up bundles in the aggregator (similar to the way the number zero is a useful thing))
- Improve detail when wrapping errors in the aggregator `Fixture`

Contributes to #133.

## Checklist

## Guidelines


aggregator/src/app/EthereumService.ts
```diff
@@ -115,6 +158,85 @@ export default class EthereumService {
...
+    calls: CallHelper<CallReturn>[],
+  ): (Promise<{
+    measureResults: CallResult<Measure>[];
+    callResults: CallResult<CallReturn>[];
+  }>) {
+    const fullCalls: CallHelper<unknown>[] = [measureCall];
+
+    for (const call of calls) {
+      fullCalls.push(call);
+      fullCalls.push(measureCall);
```
jacque006:
```suggestion
      fullCalls.push(call, measureCall);
```

contracts/contracts/Utilities.sol
```diff
@@ -0,0 +1,37 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity >=0.8.0 <0.9.0;
+pragma abicoder v2;
+
+contract Utilities {
```
jacque006: No a huge fan of this name, very generic. Maybe `Functions`, `FunctionsUtilities` instead?

We could also add this to an existing contract such as `VerificationGateway` or `BLSExpander` so we can avoid another contract deploy.

contracts/contracts/Utilities.sol
```diff
@@ -0,0 +1,37 @@
...
+    address contractAddress;
+    bytes encodedFunction;
+  }
+
+  struct FunctionResult {
+    bool success;
+    bytes returnValue;
+  }
+
+  function performSequence(
```
jacque006: Would be good to have [NatSpec](https://docs.soliditylang.org/en/v0.8.11/natspec-format.html) comments/docs for this function and contract. Can do this as follow up.

jacque006 (commented):
Nice start!

contracts/contracts/Utilities.sol
voltrevo: Done

contracts/contracts/Utilities.sol
voltrevo: Hmm. They're very generic utilities though. In fact I'm kinda hoping this already exists in some form and we can use that instead of this.

Anyway, I renamed it to `AggregatorUtilities`. Wdyt?

voltrevo: > We could also add this to an existing contract such as VerificationGateway or BLSExpander so we can avoid another contract deploy.

Adding to `VerificationGateway` would increase the audit burden (and complexity that may be considered by users) though.

For `BLSExpander`, it really doesn't fit. That contract is about compressing calldata for `processBundle`.

Also, this may be out of date, but I believe there was some concern about the maximum allowed size of a contract and pushing up against that limit.

Is there much benefit in avoiding a contract deploy? Is it more efficient to deploy A+B functions into one contract than A functions into contract A and B functions into contract B?

jacque006 (approved):

PR #138, opened 2022-02-14
Added Arbitrum gas results handling feature

kcharbo3:
## What is this PR doing?
Adds a feature to retrieve and compute gas costs for Arbitrum whenever the measure-gas.ts script is run for the Arbitrum testnet. This script will send a BLS tx and a normal token transfer tx to the Arbitrum testnet. Then, the contracts/scripts/util/arbitrum-gas-util.ts file is called to pull in gas prices from Arbitrum mainnet and append the results to the contracts/ArbitrumGasCosts.md file.

Also updates the measure-gas script to run with the new version of the BLSExpander contract.

'npx hardhat run scripts/test/measure_gas.ts --network arbitrum_testnet' 
command should succeed. After it runs, check the contracts/ArbitrumGasCosts.md file to see if two new rows have been added to the table.

No

## Checklist
- [X] I have manually tested these changes
- [X] Post a link to the PR in the group chat

## Guidelines

### DIFF
--- a/contracts/package.json
+++ b/contracts/package.json
@@ -17,7 +17,8 @@
     "@openzeppelin/contracts": "^4.3.2",
     "dotenv": "^10.0.0",
     "ethers": "^5.5.1",
-    "hardhat": "^2.6.7"
+    "hardhat": "^2.6.7",
+    "web3": "^1.7.0"
   },
   "devDependencies": {
     "@nomiclabs/hardhat-etherscan": "^2.1.3",
@@ -46,6 +47,7 @@
     "solidity-coverage": "^0.7.16",
     "ts-node": "^10.1.0",
     "typechain": "^5.1.2",
-    "typescript": "^4.3.5"
+    "typescript": "^4.3.5",
+    "web3-utils": "^1.7.0"
   }
 }
--- a/contracts/scripts/test/measure_gas.ts
+++ b/contracts/scripts/test/measure_gas.ts
@@ -1,10 +1,10 @@
 /* eslint-disable no-process-exit */
-
 import { BigNumber } from "ethers";
 import { solidityPack } from "ethers/lib/utils";
-import { Bundle } from "../../clients/src";
 import Fixture from "../../shared/helpers/Fixture";
 import TokenHelper from "../../shared/helpers/TokenHelper";
+import { processGasResultsToFile } from "../util/arbitrum_gas_util";
+import { network } from "hardhat";
 
 let fx: Fixture;
 let th: TokenHelper;
@@ -16,7 +16,7 @@ async function main() {
 }
 
 async function logGasForTransfers() {
-  const transferCounts = [29, 30, 31];
+  const transferCounts = [31];
   console.log("Batch transfers for: ", transferCounts);
   for (let i = 0; i < transferCounts.length; i++) {
     const transferCount = transferCounts[i];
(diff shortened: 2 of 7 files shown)


contracts/package.json
```diff
@@ -17,7 +17,8 @@
     "@openzeppelin/contracts": "^4.3.2",
     "dotenv": "^10.0.0",
     "ethers": "^5.5.1",
-    "hardhat": "^2.6.7"
+    "hardhat": "^2.6.7",
+    "web3": "^1.7.0"
```
jacque006, line 21: This (and a number of other dependencies) could be a `devDependencies` as it/they are not needed for contracts directly. @jzaki Any thoughts on separation of prod/dev dependencies for the contracts project? It's more relevant for projects where an npm package/runnable binary is produced and we want to keep it slim, but might be good for separation here as well.

contracts/scripts/test/measure_gas.ts
```diff
@@ -33,85 +33,102 @@ async function logGasForTransfers() {
...
       "airdropper balance before",
       await th.testToken.balanceOf(blsWallets[0].address),
     );
 
-    const AddressZero = "0x" + "0".repeat(40);
-
     console.log("Signing txs from nonce", nonce);
+
+    let actionsArr = [];
+    let encodedFunctions = [];
```
jacque006: These vars are never reassigned and so can be `const`s.
```suggestion
    const actionsArr = [];
    const encodedFunctions = [];
```

contracts/scripts/util/arbitrum_gas_util.ts
```diff
@@ -0,0 +1,291 @@
...
+        blsTxHash, numTxsAggregated, true
+    );
+    const normalEntry = await newTableEntry(
+        normalTxHash, numTxsAggregated, false
+    );
+
+    const input = '\n' + blsEntry + '\n' + normalEntry;
+
+    console.log("Writing gas results to file");
+    fs.appendFileSync(ARBITRUM_GAS_COST_FILE, input);
```
jacque006: Non-blocking i/o ops preferred.
```
import { appendFile } from 'fs/promises';
...
```
```suggestion
    await fs.appendFile(ARBITRUM_GAS_COST_FILE, input);
```

contracts/scripts/util/arbitrum_gas_util.ts
```diff
@@ -0,0 +1,291 @@
+const fs = require('fs');
+/*
+ Web3 needs to be used over Ethers since web3 tx receipts
+ for Arbitrum correctly store the feeStats object.
+ Ethers' tx receipts aren't updated to store that yet.
+ */
+const Web3 = require('web3');
+require('dotenv').config();
```
jacque006: Should be able to use imports here.
```suggestion
import fs from 'fs';
/*
 Web3 needs to be used over Ethers since web3 tx receipts
 for Arbitrum correctly store the feeStats object.
 Ethers' tx receipts aren't updated to store that yet.
 */
import Web3 from 'web3';
import dotenv from 'dotenv';

dotenv.config();
```

jacque006 (commented):
Nice work :+1: . Few nitpicks, but nothing that should hold this back.

Before we merge, could you please run the linter on the new/modified files.
```sh
# From project root
cd ./contracts
yarn lint --fix
```

Also @jzaki would be good to have you take a look as well.

contracts/package.json
jzaki: Good point. In the context of contracts, @openzeppelin/contracts is a dependency, and most others are `devDependencies`

jacque006: :+1: Moved to https://github.com/jzaki/bls-wallet/issues/139

kcharbo3:
Ran the linter that made some changes. No errors for the new/modified files.

jacque006 (approved):

PR #153, opened 2022-03-07
Separate out AggregationStrategy

### LINKED ISSUE #152: Make aggregation logic more modular

voltrevo:
## What is this PR doing?

Separates `AggregationStrategy` from `BundleService` so that all the logic for converting a known set of eligible bundles into a concrete aggregate bundle for submitting is in one place. (Basically just cut and paste half of `BundleService` and then fix some imports.)

This will make it easier eg to implement mev and other more complex strategies.

It also makes `BundleService` more focused on the admin side of receiving bundles and deciding when to try to submit things on chain.

Run the premerge script.

Resolves #152.

## Checklist

## Guidelines

### DIFF
--- a/aggregator/src/app/helpers/plus.ts
+++ b/aggregator/src/app/helpers/plus.ts
@@ -1,0 +1,3 @@
+export default function plus(a: number, b: number) {
+  return a + b;
+}
--- a/aggregator/src/app/AggregationStrategy.ts
+++ b/aggregator/src/app/AggregationStrategy.ts
@@ -1,0 +1,396 @@
+import {
+  BigNumber,
+  BlsWalletSigner,
+  Bundle,
+  ERC20,
+  ERC20__factory,
+  ethers,
+} from "../../deps.ts";
+
+import nil from "../helpers/nil.ts";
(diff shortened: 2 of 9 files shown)


aggregator/src/app/helpers/plus.ts
```diff
@@ -0,0 +1,3 @@
+export default function plus(a: number, b: number) {
+  return a + b;
+}
```
jacque006, line 3: Must've missed this last round, seems like overkill to have this separated out when something like `(acc: number, numActions: number) => acc + numActions` would do and be more descriptive.

jacque006 (approved):

aggregator/src/app/helpers/plus.ts
voltrevo: @jacque006 Hmm. As a reducer function I think `plus` is a bit simpler than a lambda that tries to add more context. Also, `plus` is used in two places so using lambdas would be a bit more verbose.

jacque006: Ah I missed the second usage, makes a little more sense now.

PR #165, opened 2022-03-11
Fix delay bypass + PROPOSAL: Refactor BLSWallet with generalized auth mechanism

### LINKED ISSUE #155: Prevent bypass of week delay by calling setExternalWallet

voltrevo:
## What is this PR doing?

tl;dr:
- Enforces authorization delay for changing your bls key (see `setPublicKey`) (ie "prevents bypass")
- Replaces all pending variables with generalized auth mechanism suitable for configuring future exotic gateways
- Replaces recovery with direct intervention by bls wallet to change the gateway to a recovery address

For this issue (#155), I was facing adding another triplet of time-delay variables and associated functions to BLSWallet.

Instead of doing that, I've had a go at adding a generalized auth mechanism which achieves the same thing and a bit more without having to be explicit about what all these variables are.

I hope it's not too much change - it could be a big improvement as it makes BLSWallet much more general along the lines we've talked about, but still preserves pretty much the way we currently do things.

The reason I changed how recovery works is because it was doing a special case for the way pending variables work. So it didn't really fit. I also feel this method is much more future-proof, since this way a wallet always has this fallback recovery mechanism regardless of what kind of gateway they end up using.

`yarn hardhat test`

Resolves #155.

## Checklist

## Guidelines


jzaki:
Can reconsider post-audit.
A problem with this design is that it depends on `delay` (and possibly other `id`s) to be stored off-chain to look up an authorisation, or search through past events. That said, I think there are might be some things to take forward into v2 with a potential refactor.

Issue #63, opened 2021-11-24
UI <-> Controller Bindings


jacque006:
Create and setup bindings from extension controllers to UI components. This could be done via a `redux` pattern or something simpler.

jacque006:
@kautukkundan Can you share the simplified `redux` package you mentioned to me awhile back here?

kautukkundan:
here - https://easy-peasy.vercel.app

voltrevo:
This was solved in #217 / #243.

E.g. to update state you call rpc like `quill.rpc.addHDAccount` here: https://github.com/web3well/bls-wallet/blob/main/extension/source/Home/Wallet/Wallets/WalletWrapper.tsx#L28

PR #168, opened 2022-03-17
Onboarding and Account creation

### LINKED ISSUE #131: Add HD wallet support
### Background

https://wiki.trezor.io/Hierarchical_deterministic_wallet

### Current Behavior

In the extension, we currently generate private keys by generating a random hex: https://github.com/jzaki/bls-wallet/blob/main/extension/source/App.ts#L151-L153

This is also how we generate private keys with new incoming updates: https://github.com/jzaki/bls-wallet/pull/42/files#diff-7ae0a811d81712d110aacca5950b30f483297c0182f6aba6ec817dc3b6ea312eR36-R39

### New Behavior

Have private keys generated using the HD wallet specs.

- `ether.js` provides utlity classes and functions to do this: https://docs.ethers.io/v5/api/utils/hdnode/
- [eth-hd-keyring](https://github.com/MetaMask/eth-hd-keyring) is a MetaMask compatible implementation which could work nicely with extension dApp interaction updates.

### Issues

[EIP-2333](https://eips.ethereum.org/EIPS/eip-2333) (stagnant, not finalized)

From `Deficiencies of the existing mechanism `:
```
The curve BLS12-381 used for BLS signatures within Ethereum 2.0 (alongside many other projects) mandates a new key derivation scheme. The most commonly used scheme for key derivation within Ethereum 1.x is BIP32 (also known as HD derivation) which deems keys greater than the curve order invalid. Based on the order of the private key subgroup of BLS12-381 and the size of the entropy utilised, more than 54% of keys generated by BIP32 would be invalid. (secp256k1 keys derived by BIP32 are invalid with probability less than 1 in 2-127.)
```

We currently/will get around this by [hashing the private key value versus directly setting it](https://github.com/thehubbleproject/hubble-bls/pull/28#discussion_r794143913), but it would be great if we could generate an HD wallet whose derived keys work natively for BLS operations.

kautukkundan:
## What is this PR doing?
- Updated Onboarding flow - 
-- HD Wallet generation + Account creation
- Interaction with QuillController via window object
- commented out some currently unused static UI elements

- run a local node with contracts deployed
- build and install extension

closes #131 
partially addresses #63 

## Checklist

## Guidelines


extension/source/Controllers/Keyring/KeyringController.ts
```diff
@@ -33,6 +34,33 @@ export default class KeyringController
...
+  isOnboardingComplete = (): boolean => {
+    return this.state.HDPhrase !== '';
+  };
+
+  async createHDAccount(): Promise<string> {
+    const mnemonic = this.state.HDPhrase;
+
+    if (mnemonic === '') {
+      const { phrase } = ethers.Wallet.createRandom().mnemonic;
+      this.setHDPhrase(phrase);
```
voltrevo, line 50: It looks like you intended `mnemonic` to be `phrase` after this point but forgot to set it.

extension/source/Controllers/Keyring/KeyringController.ts
```diff
@@ -33,6 +34,33 @@ export default class KeyringController
...
+    }
+
+    const node = ethers.utils.HDNode.fromMnemonic(mnemonic);
```
voltrevo, line 54: See comment above - if `mnemonic` was an empty string before it'll still be an empty string here.

extension/source/QuillPage/Onboarding/ReviewSecretPhrasePanel.tsx
```diff
@@ -64,6 +64,18 @@ const ReviewSecretPhrasePanel: FunctionComponent<{
     setAllCorrect(reviewWordStates.every((item) => item === true));
   };
 
+  const navigate = useNavigate();
+
+  const setHDWalletPhrase = async () => {
+    // @ts-ignore
```
voltrevo: I don't think there is a valid reason to ignore typescript here?

(There are lots of examples of this in this PR.)

extension/source/QuillPage/Wallet/Wallets/WalletDetail.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
```
voltrevo, line 1: Lots of commented code in this file. Is it possible you intended these changes to be a draft?

voltrevo (changes requested):

voltrevo:
> Interaction with QuillController via window object

Similar to my comments about `ts-ignore`, could you update this to use a react context?

Alternatively, if there's any complication there that would cause delay, there are a few different ways to preserve the type:
1. Use a module for the shared object `export default new QuillController(...);`
2. Use a helper function that adds the type information before returning it: `function getQuillController(): QuillController { return (window as any).quillController; }`
3. Use inline casts: `((window as any).quillController as QuillController).someQuillControllerFunction();`

voltrevo:
> Updated Onboarding flow

Could you be a bit more specific?

> commented out some currently unused static UI elements

Is the commented code still valuable? If so could you convert it into some documentation that explains it as an example or at least add some kind of minimal comment explaining why we're keeping commented code around.

voltrevo:
Also, I'm not sure what we'd do about this PR once the points above have been addressed since it's targeting `extension-controllers-local-rpc` instead of `main`. Could you do something about that? It looks like the right thing to do is rebase onto `main` and then include a diff in the PR description representing `extension-controllers-local-rpc` to explain how to do manual testing.

extension/source/Controllers/Keyring/KeyringController.ts
kautukkundan: Ah you are right, I updated the state but did not refetch the new state. Will fix.

extension/source/QuillPage/Onboarding/ReviewSecretPhrasePanel.tsx
kautukkundan: yeah, basically added `// @ts-ignore` wherever window object is used. This was done just to get the account generation working asap. The correct way to do would be to store the controller in state and pass it visa context/global state. Discussed this with @jacque006

kautukkundan:
> Could you be a bit more specific?

The second point described it but I guess the nesting was confusing, updated.

> Similar to my comments about ts-ignore, could you update this to use a react context?

Yes, intended to do that in a follow-up PR. We decided to just use the window object for now for quick POC.

> minimal comment explaining why we're keeping commented code around.

To avoid confusion about the dynamic and static elements on page. eg - it will create confusion to the reviewer if clicking a button doesn't change anything on the static UI. Will keep bringing back commented elements as they get integrated with the controllers.

> Also, I'm not sure what we'd do about this PR 

To me it makes the most sense to continue development with local RPC (into main) for faster testing and once the things are in place, create another PR for changing the target from local to testnet and/or create a env config to auto select that. Would like to hear from @jacque006 about it as well.

kautukkundan (commented):
.

jacque006:
> Similar to my comments about ts-ignore, could you update this to use a react context?

Per just getting this to work, I recommended using the `window` for now. I have some changes I'm going to try and wrap up today that make this type safe and make it easier for us to refactor that pattern later. This includes extending the `Window` type so we don't need to do any weird casts, and we can easily refactor to another access pattern later.

> To me it makes the most sense to continue development with local RPC (into main) for faster testing and once the things are in place, create another PR for changing the target from local to testnet and/or create a env config to auto select that. Would like to hear from @jacque006 about it as well.

I'm fine including this in the local RPC work I started if it will make things easier from a local dev perspective.

extension/source/Controllers/Keyring/KeyringController.ts
jacque006: Fixed.

extension/source/types/window.d.ts
```diff
@@ -0,0 +1,10 @@
+import type QuillController from '../Controllers/QuillController';
+import type KeyringController from '../Controllers/Keyring/KeyringController';
+
+export declare global {
+  interface Window {
+    // TODO (merge-ok) Remove both of these when full controller linking is completed.
+    QuillController: () => QuillController;
+    KeyringController: () => KeyringController;
+  }
+}
```
jacque006, line 10: @voltrevo This addresses typing issues, and give us an easy point to remove the ad-hoc injection once we get more hooked up.

jacque006:
> Use a helper function that adds the type information before returning it: function getQuillController(): QuillController { return (window as any).quillController; }

Ended up going with something similar to this with throws if the controller is accessed before being initialized.

@voltrevo I believe all comments have been addressed/commented on, ready for your re-review.

voltrevo (approved):

voltrevo:
> I'm fine including this in the local RPC work I started if it will make things easier from a local dev perspective.

I agree, changed the base to main so we can just merge this in

PR #185, opened 2022-04-05
Simplify success

voltrevo:
## What is this PR doing?

Simplifies the way `success` is reported by `performOperation`.

`(_results.length > 0)` appears to be derived from the call to `_performOperation`, but in fact the length of `_results` is fixed to be equal to the number of actions (even if the first action fails, `_results` is default initialized to e.g. `[0x, 0x, 0x]` for 3 actions, so `.length` will still be 3). `success` is instead best understood as whether the operations were successfully processed, which is always true in this branch by virtue of `_performOperation` not throwing an exception.

If requiring actions to be non-empty is desirable, then I suggest adding `require(op.actions.length > 0);` to `_performOperation` (I don't think it's desirable though - we actually already do this in testing).

(I burned a lot of time today trying to troubleshoot an unexpected `success === false` problem, and I thought there was something funky going on with empty byte arrays. Turned out it was unrelated, but I would have found the problem faster if it wasn't for this point of confusion.)

Try running `processBundle` via `callStatic` with a bundle that has an operation with an empty list of actions. `success` should be `true` (was `false`).

## Checklist

## Guidelines

### DIFF
--- a/contracts/contracts/BLSWallet.sol
+++ b/contracts/contracts/BLSWallet.sol
@@ -202,7 +202,7 @@ contract BLSWallet is Initializable, IBLSWallet
         try this._performOperation(op) returns (
             bytes[] memory _results
         ) {
-            success = (_results.length > 0); // false when no actions given
+            success = true;
             results = _results;
         }
         catch {


contracts/contracts/BLSWallet.sol
```diff
@@ -202,7 +202,7 @@ contract BLSWallet is Initializable, IBLSWallet
         try this._performOperation(op) returns (
             bytes[] memory _results
         ) {
-            success = (_results.length > 0); // false when no actions given
```
jacque006, line 205: @jzaki Can you provide more context on why this was changed from feedback from the last audit? https://github.com/jzaki/bls-wallet/commit/3bbcdb29e7de1e48d23a5634df06449611d4db2c

jzaki: The audit feedback (just commentary) indicated that the return value would be true or the call to `_performOperation` would revert, so the return value was not functionally necessary. So in [3bbcdb2](https://github.com/jzaki/bls-wallet/commit/3bbcdb29e7de1e48d23a5634df06449611d4db2c) I had changed success to be true for all actions succeeding, or false if a call with no actions succeeded.

It looks like the change to `true` allows no-action operations to be considered successful operations, which might make somethings easier for the aggregator? Yet I don't see an incentive for anyone to use that.
@voltrevo?

voltrevo: @jzaki Covered this in the PR description:

> If requiring actions to be non-empty is desirable, then I suggest adding require(op.actions.length > 0); to _performOperation (I don't think it's desirable though - we actually already do this in testing).

There's a use case where an empty action list can be used as a no-op to increment your nonce (or if you just want your wallet contract to be deployed). But for me it's just as much about consistency. Inconsistency leads to mistakes. `[].every(x => x)` is `true`, not `false`, because every element of that empty list is true.

jacque006: Thanks @jzaki & @voltrevo for more context, I think it's fine to allow no-ops.

jacque006 (approved):

Issue #179, opened 2022-04-04
Create rpc endpoint `eth_getTransactionReceipt`


kautukkundan:

jacque006:
Resolved in https://github.com/web3well/bls-wallet/pull/186

Issue #178, opened 2022-04-04
Return Transaction hash from Aggregator


kautukkundan:

jacque006:
Resolved in https://github.com/web3well/bls-wallet/pull/186

PR #186, opened 2022-04-08
Add receipts

voltrevo:
## What is this PR doing?

https://user-images.githubusercontent.com/9291586/162421564-0813d1d9-0f32-479f-8e55-f3222ddd3f50.mov

(I forgot to refresh the page to show the billboard is successfully updated. It does work though 😄.)

- Adds support for transaction receipts when using extension+aggregator
- Uses random ids (instead of autoincrement) for the primary key of the bundles table, which is better suited as a stand-in for `transactionHash`. It also simplifies things a bit since we don't need to rely on the database to generate the ids for us, and so `BundleRow["id"]` is no longer optional.
- Changes the response to POST `/bundle` to include the row id on successful insertion
- Removes old background script replaced by Chai's work

Replicate the result shown in the video. This requires running:
- Local node
- Deploying contracts
- Deploying billboard
- Aggregator

Note that the billboard uses the block timestamp to determine whether it's ok to set it to a new value. This can cause issues in the aggregator because callStatic won't see the new timestamp and won't submit the bundle until it does. This isn't a big deal on a real network with a constant flow of transactions, but to resolve it locally, just add an artificial transaction, such as by running `aggregator/manualTests/send1Wei.ts`.

#178, #179

## Checklist

## Guidelines


aggregator/src/app/BundleService.ts
```diff
@@ -32,7 +34,13 @@ export default class BundleService {
 
   unconfirmedBundles = new Set<Bundle>();
   unconfirmedActionCount = 0;
-  unconfirmedRowIds = new Set<number>();
+  unconfirmedRowIds = new Set<string>();
+
+  // TODO: Use database table?
```
jacque006: Yes, In case agg service crashes. We also will likely need to parse current chain state and extract existing transactions on startup. This is needed for:
- New clean aggregator just starting up.
- Existing aggregator catching up after a crash/restart/version upgrade

Def should do this as multiple new issues :)

aggregator/src/app/BundleService.ts
```diff
@@ -250,22 +281,35 @@ export default class BundleService {
...
+            bundle: row.bundle,
+            receipt,
+          });
+        }
+
+        setTimeout(() => {
+          for (const row of includedRows) {
+            this.confirmedBundles.delete(row.id);
+          }
+        }, 5 * 60_000);
```
jacque006: This seems like a dangerous/racey way delete, should have it occur off another emitted event.

aggregator/src/app/BundleTable.ts
```diff
@@ -22,23 +23,29 @@ import { parseBundleDto } from "./parsers.ts";
...
   bundle: Bundle;
   eligibleAfter: BigNumber;
   nextEligibilityDelay: BigNumber;
 };
 
+export function makeId() {
+  const buf = new Uint8Array(32);
+  crypto.getRandomValues(buf);
+  return ethers.utils.hexlify(buf);
+}
```
jacque006: May want to use a UUID here https://examples.deno.land/uuids so that we don't have any collisions and use a more standardized id format.

extension/source/Controllers/Network/createJsonRpcClient.ts
```diff
@@ -46,6 +49,85 @@ export function createProviderConfigMiddleware(
...
+  unknown
+> {
+  return (
+    req: JRPCRequest<unknown>,
+    res: JRPCResponse<unknown>,
+    next: JRPCEngineNextCallback,
+    end: JRPCEngineEndCallback,
+  ) => {
+    if (req.method === 'eth_getTransactionByHash') {
+      const hash: string = (req.params as any)[0];
```
jacque006: See if we can use a better type than `any` here.

extension/source/Controllers/Network/createJsonRpcClient.ts
```diff
@@ -46,6 +49,85 @@ export function createProviderConfigMiddleware(
...
+
+        // Here we're just relaying the information that we already know
+        // internally about the transaction. ethers needs this response to
+        // function properly.
+        res.result = {
+          hash,
+          from: knownTx.from,
+          nonce: knownTx.nonce,
+          value: knownTx.value,
+          gasLimit: '0x0',
```
jacque006, line 77: Should this be pulled from the higher level txn or the estimated fee? Would it be the average of all bundled txn fees?

extension/source/Controllers/Network/createJsonRpcClient.ts
```diff
@@ -46,6 +49,85 @@ export function createProviderConfigMiddleware(
...
+
+function createAggregatorMiddleware(): JRPCMiddleware<unknown, unknown> {
+  return async (
+    req: JRPCRequest<unknown>,
+    res: JRPCResponse<unknown>,
+    next: JRPCEngineNextCallback,
+    end: JRPCEngineEndCallback,
+  ) => {
+    if (req.method === 'eth_getTransactionReceipt') {
+      const hash: string = (req.params as any)[0];
```
jacque006: See if we can use a better type than any here.

extension/source/Controllers/Network/createJsonRpcClient.ts
```diff
@@ -46,6 +49,85 @@ export function createProviderConfigMiddleware(
...
+          transactionIndex: bundleReceipt.transactionIndex,
+          blockHash: bundleReceipt.blockHash,
+          blockNumber: bundleReceipt.blockNumber,
+          from: knownTx.from,
+          to: knownTx.to,
+          logs: [],
+          cumulativeGasUsed: '0x0',
+          gasUsed: '0x0',
+          status: '0x1',
+          effectiveGasPrice: '0x0',
```
jacque006, line 120: Can we included gas stats? Likely split to another issue.

extension/source/Controllers/Network/createJsonRpcClient.ts
```diff
@@ -46,6 +49,85 @@ export function createProviderConfigMiddleware(
...
+        }
+
+        res.result = {
+          transactionHash: hash,
+          transactionIndex: bundleReceipt.transactionIndex,
+          blockHash: bundleReceipt.blockHash,
+          blockNumber: bundleReceipt.blockNumber,
+          from: knownTx.from,
+          to: knownTx.to,
+          logs: [],
```
jacque006, line 116: We can likely figure out a way to parse all logs for a bundled txn and then split into the segment for this bundle/action.

This would likely require changes to the VG/BLS Wallet contracts to emit events between every bundle, and maybe even at the granularity of between each action. @jzaki what do you think?

jacque006 (commented):
Overall solid work getting this functionality baked in. I think there is a good amount of discovered work/new issues from this and a few things we can try to tweak now.

aggregator/src/app/BundleService.ts
jacque006: We actually want this confirmation to stick around forever as someone may want to check on it well into the future. Once we persist these to db will be stable across aggregator restarts. Removing delete.

aggregator/src/app/BundleTable.ts
jacque006: What we actually should do here is hash the bundle (a la the example in https://stackoverflow.com/a/57638023) so that anyone with the bundle payload can infer the hash and query its status similarly to how they could with a normal transaction. Can in follow up work.

aggregator/src/app/BundleService.ts
jacque006: - https://github.com/web3well/bls-wallet/issues/190
- https://github.com/web3well/bls-wallet/issues/191

aggregator/src/app/BundleTable.ts
jacque006: https://github.com/web3well/bls-wallet/issues/192

extension/source/Controllers/Network/createJsonRpcClient.ts
jacque006: Need to think a bit more about this. don't think it's currently worth pursuing in the near/medium future.

extension/source/Controllers/Network/createJsonRpcClient.ts
jacque006: Need to think a bit more about this. don't think it's currently worth pursuing in the near/medium future.

extension/source/Controllers/Network/createJsonRpcClient.ts
jacque006: https://github.com/web3well/bls-wallet/issues/193

jacque006 (approved):

PR #196, opened 2022-04-22
Arbitrum testnet update

voltrevo:
## What is this PR doing?

Records the addresses of new arbitrum testnet deployment

Also:
- Comments out some code in an aggregator manual test that isn't working properly
  - (Should be fixed instead but not introduced here so should be fixed separately)
- Adds some logging in `Create2Fixture`

- Inspect the addresses using https://testnet.arbiscan.io/
- Run an aggregator pointing to arbitrum testnet and these addresses, run manualTests/mint1ViaAggregator.ts and confirm mint works successfully

## Checklist

## Guidelines

### DIFF
--- a/contracts/shared/helpers/Create2Fixture.ts
+++ b/contracts/shared/helpers/Create2Fixture.ts
@@ -83,7 +83,19 @@ export default class Create2Fixture {
 
     // If contract doesn't exist at expected address, deploy it there
     if ((await ethers.provider.getCode(contractAddress)) === "0x") {
-      await (await create2Deployer.deploy(salt, initCode)).wait();
+      const receipt = await (
+        await create2Deployer.deploy(salt, initCode)
+      ).wait();
+
+      console.log(
+        [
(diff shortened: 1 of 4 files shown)


contracts/shared/helpers/Create2Fixture.ts
```diff
@@ -83,7 +83,19 @@ export default class Create2Fixture {
...
+        await create2Deployer.deploy(salt, initCode)
+      ).wait();
+
+      console.log(
+        [
+          `Deployed ${contractName}`,
+          `using ${receipt.gasUsed.toString()} gas:`,
+          contractAddress,
+        ].join(" "),
+      );
```
jacque006, line 96: You should be able to pass this in as [one formatted string or as a comma separated list](https://developer.mozilla.org/en-US/docs/Web/API/Console/log) and let the CLI output format it depending on terminal width. Reason you wanted to use this specific format?

jacque006 (approved):

PR #206, opened 2022-05-19
Radically simplify how window.ethereum methods are defined

voltrevo:
## What is this PR doing?

`createEthMiddleware` was unnecessarily prescriptive about how we define our `window.ethereum` methods. Two bad things were happening:
1. `createEthMiddleware` restricted exactly which methods could be provided without helping to define how those methods work in any way
2. `createEthMiddleware` did multiple unnecessary name transforms on each method, for example `submitBatch` is the basis for `submitTransaction` and is then used to define `eth_sendTransaction` rather than simply using the name `eth_sendTransaction` from the start

This PR fixes both of these problems. Now the method names in `QuillController` map exactly to the `window.ethereum` method names (so it's easy to search the codebase and find what you're looking for instead of finding where to start tracing the plumbing back to what you actually want).

If we define any more rpc methods, it's also much simpler. As an example, you can simply add this to the method list in `QuillController`:

```diff
 eth_accounts: async () =>
   // Expose no accounts if this origin has not been approved, preventing
   // account-requiring RPC methods from completing successfully
   // only show address if account is unlocked
   this.selectedAddress ? [this.selectedAddress] : [],
+
+foo: async () => 'bar',
 
 eth_coinbase: async () => this.selectedAddress || null,
```

Yep just one line. Then in the console:
Screen Shot 2022-05-19 at 5 28 52 pm

Replicate adding a new `window.ethereum` method as described above. Check existing methods still work as expected.

Also try searching for `eth_sendTransaction` and note that you'll immediately land on the code that's relevant. Compare this to the same thing on the `main` branch and observe the significant work involved to then figure out where the relevant code is.

Nope.

## Checklist

## Guidelines

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -10,7 +10,7 @@
     "url": "https://blswallet.org"
   },
   "engines": {
-    "node": ">=10.0.0",
+    "node": ">=12.0.0",
     "yarn": ">= 1.0.0"
   },
   "scripts": {
--- a/extension/source/Controllers/Network/createEthMiddleware.ts
+++ b/extension/source/Controllers/Network/createEthMiddleware.ts
@@ -7,6 +7,7 @@ import {
 } from '@toruslabs/openlogin-jrpc';
 import { BigNumberish, BytesLike } from 'ethers';
 import { PROVIDER_JRPC_METHODS } from '../../common/constants';
+import web3_clientVersion from './web3_clientVersion';
 
 export type SendTransactionParams = {
   from: string;
@@ -18,7 +19,6 @@ export type SendTransactionParams = {
 };
 
 export interface IProviderHandlers {
-  version: string;
   getAccounts: (req: JRPCRequest<unknown>) => Promise<string[]>;
   requestAccounts: (req: JRPCRequest<unknown>) => Promise<string[]>;
   getProviderState: (
@@ -29,7 +29,6 @@ export interface IProviderHandlers {
 }
 
 export function createWalletMiddleware({
-  version,
   getAccounts,
   requestAccounts,
   getProviderState,
@@ -92,7 +91,7 @@ export function createWalletMiddleware({
   }
 
   return createScaffoldMiddleware({
-    web3_clientVersion: `Quill/v${version}`,
+    web3_clientVersion,
     // account lookups
     eth_accounts: createAsyncMiddleware(lookupAccounts),
     eth_coinbase: createAsyncMiddleware(lookupDefaultAccount),
--- a/extension/source/Controllers/Network/createEthMiddleware.ts
+++ b/extension/source/Controllers/Network/createEthMiddleware.ts
@@ -6,7 +6,6 @@ import {
   JRPCResponse,
 } from '@toruslabs/openlogin-jrpc';
 import { BigNumberish, BytesLike } from 'ethers';
-import { PROVIDER_JRPC_METHODS } from '../../common/constants';
 import web3_clientVersion from './web3_clientVersion';
 
 type SimpleHandler = (req: JRPCRequest<unknown>) => Promise<unknown>;
@@ -53,8 +52,7 @@ export function createWalletMiddleware({
     eth_accounts: toAsyncMiddleware(getAccounts),
     eth_coinbase: toAsyncMiddleware(eth_coinbase),
     eth_requestAccounts: toAsyncMiddleware(requestAccounts),
-    [PROVIDER_JRPC_METHODS.GET_PROVIDER_STATE]:
-      toAsyncMiddleware(getProviderState),
+    wallet_get_provider_state: toAsyncMiddleware(getProviderState),
     eth_setPreferredAggregator: toAsyncMiddleware(setPreferredAggregator),
     eth_sendTransaction: toAsyncMiddleware(eth_sendTransaction),
   });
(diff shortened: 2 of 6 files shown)


extension/package.json
```diff
@@ -10,7 +10,7 @@
     "url": "https://blswallet.org"
   },
   "engines": {
-    "node": ">=10.0.0",
+    "node": ">=12.0.0",
```
jacque006, line 13: Might be safe to go to v14 or v16 (LTS) as well https://nodejs.org/en/about/releases/

extension/source/Controllers/Network/web3_clientVersion.ts
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const web3_clientVersion = 'Quill/v0.1.0';
+
+export default web3_clientVersion;
```
jacque006, line 4: ~Lint disable seems like overkill vs. using a more standard name.~ I see the usage point now.
Also, you could pull the version segment from `package.json` so we don't have to track in multiple places.

```suggestion
import { version } from '../../../package.json';
// Match createScaffoldMiddleware payload name
// eslint-disable-next-line @typescript-eslint/naming-convention
const web3_clientVersion = `Quill/v${version}`;

export default web3_clientVersion;
```

extension/source/Controllers/Network/createEthMiddleware.ts
```diff
@@ -6,7 +6,17 @@ import {
   JRPCResponse,
 } from '@toruslabs/openlogin-jrpc';
 import { BigNumberish, BytesLike } from 'ethers';
-import { PROVIDER_JRPC_METHODS } from '../../common/constants';
+import web3_clientVersion from './web3_clientVersion';
+
+type ProviderHandler = (req: JRPCRequest<unknown>) => Promise<unknown>;
```
jacque006, line 11: This is likely beyond the scope of this work, but would be nice if we could genericize this vs. using `unknown`.

```suggestion
type ProviderHandler<ReqParams, ResResult> = (req: JRPCRequest<ReqParams>) => Promise<ResResult>;
```

extension/source/Controllers/QuillController.ts
```diff
@@ -512,28 +508,30 @@ export default class QuillController extends BaseController<
...
+      wallet_get_provider_state: async () => {
         return {
           accounts: this.selectedAddress ? [this.selectedAddress] : [],
           chainId: this.networkController.state.chainId,
           isUnlocked: !!this.selectedAddress,
         };
       },
 
-      setPreferredAggregator: async (req: any) => {
+      eth_setPreferredAggregator: async (req: any) => {
```
jacque006, line 527: Similarly, could type these functions instead of `any`.

```suggestion
      eth_setPreferredAggregator: async (req: JRPCRequest<[string]>) => {
```

jacque006 (approved):
Few suggestions, overall good improvement :+1:

voltrevo:
Tracking followup here: https://github.com/web3well/bls-wallet/issues/207

extension/source/Controllers/QuillController.ts
voltrevo: Nah it's more complicated than that. The request data is ultimately an external source that could send us any json data so in order to add type information we need to add runtime type checking (e.g. using io-ts).

There's also `.tabId` which isn't in the type information for `JRPCRequest` so typescript thinks it's not there no matter what type we give to it.

Definitely in favour of improving our typing but all of this stuff is very half-hearted in its type information so it's difficult to improve that situation without doing a deeper dive.

extension/package.json
voltrevo: This was fun... so LTS is actually v18 now so I tried that.

This broke webpack because... openssl removed support for md4 and therefore node removed it, but webpack uses it: https://github.com/webpack/webpack/blob/v4.46.0/lib/NamedModulesPlugin.js#L11 😱😱😭.

I don't suppose webpack needs a secure hash function, but, considering md4 was broken in 1995 - long before webpack or nodejs for that matter, it's amazing that webpack would choose md4. Even md5 would be less surprising.

Anyway, I guess we're stuck on node 16 until we upgrade to webpack 5 😂.

voltrevo:
@jacque006 https://github.com/web3well/bls-wallet/pull/216

extension/package.json
voltrevo: done

extension/source/Controllers/Network/web3_clientVersion.ts
voltrevo: done

extension/source/Controllers/Network/createEthMiddleware.ts
voltrevo: done

extension/source/Controllers/QuillController.ts
jacque006: We could add `.tabId` by using an extended type (`TabIDJRPCRequest extends JRPCRequest...`) but I see your point about unknown external data. Would be good to sanitize/check it for anything malicious but since this is not intended to be a production wallet less of a priority.

Given that, in general I think we can switch many of our `any` types to `unknown`. This will force relevant type checks such as using `in` or `typeof` and explicit casts to the expected type. If we want to be really strict, we can enable https://www.typescriptlang.org/tsconfig#noImplicitAny & https://typescript-eslint.io/rules/no-explicit-any

PR #208, opened 2022-05-26
Updated VerificationGateway to use EIP-712 

### LINKED ISSUE #25: Update BLS_DOMAIN to EIP-712
Similarly to https://github.com/thehubbleproject/hubble-contracts/issues/587 , it would make sense to update the BLS signing domain at `VerificationGateway.BLS_DOMAIN` to match the EIP-712 domain spec.

kautukkundan:
## What is this PR doing?
Added EIP-712 support to verification gateway to use more secure domain and prevent signature replay on different chains.  

- currently only some of the tests pass that too on a separate hardhat node
- TODO - fix gas limit exceeded error on hardhat test node

closes #25 

## Checklist

## Guidelines

### DIFF
--- a/contracts/clients/test/index.test.ts
+++ b/contracts/clients/test/index.test.ts
@@ -8,7 +8,7 @@ import { initBlsWalletSigner, Bundle, Operation } from "../src/signer";
 
 import Range from "./helpers/Range";
 
-const domain = arrayify(keccak256("0xfeedbee5"));
+const domain = arrayify("0x425bd46b7016e0395c00f2e05fd74f938023d31f355d5a62fb9c63756c6a5d87");
 const weiPerToken = BigNumber.from(10).pow(18);
 
 const samples = (() => {
--- a/contracts/contracts/VerificationGateway.sol
+++ b/contracts/contracts/VerificationGateway.sol
@@ -8,6 +8,7 @@ import "@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol";
 import "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
 
 import "./interfaces/IWallet.sol";
+import {EIP712} from "@openzeppelin/contracts/utils/cryptography/draft-EIP712.sol";
 
 /**
 A non-upgradable gateway used to create BLSWallets and call them with
@@ -16,11 +17,12 @@ The gateway holds a single ProxyAdmin contract for all wallets, and can
 only called by a wallet that the VG created, and only if the first param
 is the calling wallet's address.
  */
-contract VerificationGateway
+contract VerificationGateway is EIP712
 {
-    /** Domain chosen arbitrarily */
-    bytes32 BLS_DOMAIN = keccak256(abi.encodePacked(uint32(0xfeedbee5)));
     uint8 constant BLS_KEY_LEN = 4;
+
+    string public constant BLS_DOMAIN = "BLS_WALLET";
+    string public constant BLS_DOMAIN_VERSION = "1";
 
     IBLS public blsLib;
     ProxyAdmin public immutable walletProxyAdmin;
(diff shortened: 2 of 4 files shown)


contracts/clients/test/index.test.ts
```diff
@@ -8,7 +8,7 @@ import { initBlsWalletSigner, Bundle, Operation } from "../src/signer";
 
 import Range from "./helpers/Range";
 
-const domain = arrayify(keccak256("0xfeedbee5"));
+const domain = arrayify("0x425bd46b7016e0395c00f2e05fd74f938023d31f355d5a62fb9c63756c6a5d87");
```
jacque006, line 11: This and the value above should be pulled directly from the  `VerificationGateway` (VG) linked to a wallet rather than hardcoded. It is possible someone may:
1. Deploy a VG to a different address if they use a different deployer which would cause this hash to change.
2. Upgrade the VG a wallet uses to a newer/different version, which could completely change the domain for signing. This is a risk currently with how we have the client code.

This requires us to change points consuming `domain` or `defaultDomain` to pull an async value from the wallet which is a larger change set but I think ultimately beneficial.

contracts/contracts/VerificationGateway.sol
```diff
@@ -8,6 +8,7 @@ import "@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol";
 import "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
 
 import "./interfaces/IWallet.sol";
+import {EIP712} from "@openzeppelin/contracts/utils/cryptography/draft-EIP712.sol";
```
jacque006, line 11: I would stick with the import format of the file as is.

```suggestion
import "@openzeppelin/contracts/utils/cryptography/draft-EIP712.sol";
```

contracts/contracts/VerificationGateway.sol
```diff
@@ -16,12 +17,13 @@ The gateway holds a single ProxyAdmin contract for all wallets, and can
...
 is the calling wallet's address.
  */
-contract VerificationGateway
+contract VerificationGateway is EIP712
 {
-    /** Domain chosen arbitrarily */
-    bytes32 BLS_DOMAIN = keccak256(abi.encodePacked(uint32(0xfeedbee5)));
     uint8 constant BLS_KEY_LEN = 4;
 
+    string public constant BLS_DOMAIN = "BLS_WALLET";
```
jacque006, line 24: #nitpick, prefer something like `BLS_DOMAIN_NAME`, as this is not the actual domain.

```suggestion
    string public constant BLS_DOMAIN_NAME = "BLS_WALLET";
```

contracts/contracts/VerificationGateway.sol
```diff
@@ -16,12 +17,13 @@ The gateway holds a single ProxyAdmin contract for all wallets, and can
 only called by a wallet that the VG created, and only if the first param
 is the calling wallet's address.
  */
-contract VerificationGateway
+contract VerificationGateway is EIP712
```
jacque006, line 20: @jacque006 Need to update docs/comments for VG to indicate it is already close to max byte size in current form.

contracts/contracts/VerificationGateway.sol
```diff
@@ -361,4 +363,7 @@ contract VerificationGateway
         );
     }
 
+     function domainSeparator() public view returns (bytes32) {
```
jacque006, line 366: If this is not going to be used inside the contract, such as at the `_domainSeparatorV4` callpoints above, it can be marked as external. In this case it doesn't matter much, but see https://ethereum.stackexchange.com/questions/19380/external-vs-public-best-practices for a reason why it can make a difference.

```suggestion
     function domainSeparator() external view returns (bytes32) {
```

jacque006 (commented):
Thanks for implementing this. Per conversation with @jzaki , we need to put this on the backburner due to the `VerificationGateway` contract size, and we can re-evaluate when we decide to break it apart.

I'll remove this issue from the audit milestone, and I would close this PR but leave the branch up so we can reference this PR when we decide to pick it back up.

jacque006:
Closing until we can resolve `VerificationGateway` contract size issue or figure out a workaround.

PR #209, opened 2022-05-27
Fix public key override to zero after creation

### LINKED ISSUE #134: Wallet public key can be set to the 0 key by anyone after wallet creation
The state variable pendingBLSPublicKeyTime is not initialized to type(uint256).max in the constructor like the other time variables in the BLSWallet contract. When a wallet is created via the VerificationGateway, it calls wallet.latchBLSPublicKey right after with the correct public key. Since the default value is 0 for pendingBLSPublicKey, after a wallet is created and its public key latched, any user can call wallet.setAnyPending() and therefore set the blsPublicKey state variable to the 0 key.

https://github.com/jzaki/bls-wallet/blob/main/contracts/contracts/BLSWallet.sol#L150-L155

This can be a major issue if other applications rely on wallet.blsPublicKey() to get the expected public key for the wallet. Additionally the wallet owner would have to wait a week until they can reset their blsPublicKey field via setBLSPublicKey() and setAnyPending().

Similarly, the pendingRecoveryHashTime state variable is also not set to max value at construction. However when the recoveryHash is set for the first time, the pendingRecoveryHashTime variable is set to the max value. So no one can easily reset it to the 0 value.

kautukkundan:
## What is this PR doing?
- update `pendingBLSPublicKeyTime` after latchBLSPublicKey is called 
- 6510604 adds a test to check the presence of the bug
- 45d1a50 updates the contract and adds a test case

run npx hardhat test

closes #134 

## Checklist

## Guidelines

### DIFF
--- a/contracts/contracts/BLSWallet.sol
+++ b/contracts/contracts/BLSWallet.sol
@@ -65,6 +65,8 @@ contract BLSWallet is Initializable, IBLSWallet
         trustedBLSGateway = blsGateway;
         pendingGatewayTime = type(uint256).max;
         pendingPAFunctionTime = type(uint256).max;
+        pendingRecoveryHashTime = type(uint256).max;
+        pendingBLSPublicKeyTime = type(uint256).max;
     }
 
     /** */
@@ -73,7 +75,6 @@ contract BLSWallet is Initializable, IBLSWallet
     ) public onlyTrustedGateway {
         require(isZeroBLSKey(blsPublicKey), "BLSWallet: public key already set");
         blsPublicKey = blsKey;
-        pendingBLSPublicKeyTime = type(uint256).max;
     }
 
     function isZeroBLSKey(uint256[4] memory blsKey) public pure returns (bool) {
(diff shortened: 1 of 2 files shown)


contracts/contracts/BLSWallet.sol
```diff
@@ -73,6 +73,7 @@ contract BLSWallet is Initializable, IBLSWallet
     ) public onlyTrustedGateway {
         require(isZeroBLSKey(blsPublicKey), "BLSWallet: public key already set");
         blsPublicKey = blsKey;
+        pendingBLSPublicKeyTime = type(uint256).max;
```
jzaki: For consistency could be done in `intialize`.

jzaki: Also could initialise other pending times:
```
+        pendingRecoveryHashTime = type(uint256).max;
         pendingPAFunctionTime = type(uint256).max;
+        pendingBLSPublicKeyTime = type(uint256).max;
+        pendingGatewayTime = type(uint256).max;
```

kautukkundan: resolved in 02d64c1

jzaki (approved):

jacque006 (approved):

Issue #104, opened 2022-01-18
2. Multiple BLS keys can be registered for each smart contract wallet


jzaki:
The function _setExternalWallet in VerificationGateway.sol can be called
several times with different BLS keys, resulting in several entries to the same wallet address
with different BLS keys. This breaks the assumption that one key controls one wallet.
Recommendation
Consider allowing only one BLS key per wallet to maintain the one-to-one mapping.

PR #211, opened 2022-05-30
added reverse mapping for external wallets from hash

kautukkundan:
## What is this PR doing?
added reverse mapping for external wallets so that the address can be looked up when `setExternalWallet` is called and the wallets which are already registered can be denied re-registration to maintain 1:1 mapping with BLS Key and Wallet.

#104 

## Checklist

## Guidelines

### DIFF
--- a/contracts/contracts/VerificationGateway.sol
+++ b/contracts/contracts/VerificationGateway.sol
@@ -26,6 +26,7 @@ contract VerificationGateway
     ProxyAdmin public immutable walletProxyAdmin;
     address public blsWalletLogic;
     mapping(bytes32 => IWallet) externalWalletsFromHash;
+    mapping(address => bool) registeredWalletAddresses;
 
 
     /** Aggregated signature with corresponding senders + operations */
@@ -127,6 +128,7 @@ contract VerificationGateway
         uint256[2] calldata messageSenderSignature,
         uint256[BLS_KEY_LEN] calldata publicKey
     ) public {
+        require(!registeredWalletAddresses[msg.sender], "mapping to this wallet already exists");
         safeSetWallet(messageSenderSignature, publicKey, msg.sender);
     }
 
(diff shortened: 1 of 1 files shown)


voltrevo:
@kautukkundan I've pushed a new failing test but it's not quite right... I'm pretty sure there's something going on here though, please bear with me while I try to sort it out 🤔.

voltrevo:
It looks like this mechanism only protects in the case that an upgraded wallet makes a duplicate call to `setExternalWallet`, and doesn't cover the case of calling `setExternalWallet` where the wallet is native to the gateway.

However, I think you guys are across this and you can basically disregard my changes where I was getting tangled up in some other ideas. I've pushed this branch back to where it was. If you're interested in what I've removed, I created another branch here - https://github.com/web3well/bls-wallet/tree/feat/reverse-lookup--andrew-wip.

contracts/contracts/VerificationGateway.sol
```diff
@@ -26,6 +26,7 @@ contract VerificationGateway
     ProxyAdmin public immutable walletProxyAdmin;
     address public blsWalletLogic;
     mapping(bytes32 => IWallet) externalWalletsFromHash;
+    mapping(address => bool) registeredWalletAddresses;
```
jzaki, line 29: For the reverse lookup, this should be the wallet address to bls key hash.

contracts/contracts/VerificationGateway.sol
```diff
@@ -127,6 +128,7 @@ contract VerificationGateway
         uint256[2] calldata messageSenderSignature,
         uint256[BLS_KEY_LEN] calldata publicKey
     ) public {
+        require(!registeredWalletAddresses[msg.sender], "mapping to this wallet already exists");
```
jzaki, line 131: I don't think this handles the scenario of a wallet upgrading it's VG, then wanting to go back.
Wallets should be able to unregister past bls key hash associations.

kautukkundan:
closed in favour of #212

PR #212, opened 2022-05-31
Fixes: one-to-one hash wallet 

### LINKED ISSUE #104: 2. Multiple BLS keys can be registered for each smart contract wallet

jzaki:
## What is this PR doing?
Ensuring a wallet only has one bls key registered for it.

@kautukkundan can you write a test for this?

Fixes #104 

## Checklist

## Guidelines

### DIFF
--- a/contracts/contracts/VerificationGateway.sol
+++ b/contracts/contracts/VerificationGateway.sol
@@ -85,12 +85,6 @@ contract VerificationGateway
     }
 
     /**
-    Returns a BLSWallet if deployed from this contract, otherwise 0.
-    @param hash BLS public key hash used as salt for create2
-    @return BLSWallet at calculated address (if code exists), otherwise zero address
-     */
-
-    /**
     If an existing wallet contract wishes to be called by this verification
     gateway, it can directly register itself with a simple signed msg.
     NB: this is independent of the proxyAdmin, and if desired can be changed
--- a/contracts/contracts/VerificationGateway.sol
+++ b/contracts/contracts/VerificationGateway.sol
@@ -52,11 +52,12 @@ contract VerificationGateway
      */
     constructor(
         IBLS bls,
-        address blsWalletImpl
+        address blsWalletImpl,
+        ProxyAdmin proxyAdmin
     ) {
         blsLib = bls;
         blsWalletLogic = blsWalletImpl;
-        walletProxyAdmin = new ProxyAdmin();
+        walletProxyAdmin = proxyAdmin;
     }
 
     /** Throw if bundle not valid or signature verification fails */
--- a/contracts/contracts/VerificationGateway.sol
+++ b/contracts/contracts/VerificationGateway.sol
@@ -66,11 +66,11 @@ contract VerificationGateway
     constructor(
         IBLS bls,
         address blsWalletImpl,
-        ProxyAdmin proxyAdmin
+        address proxyAdmin
     ) {
         blsLib = bls;
         blsWalletLogic = blsWalletImpl;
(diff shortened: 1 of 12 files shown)


contracts/package.json
```diff
@@ -17,6 +17,7 @@
     "@nomiclabs/hardhat-ethers": "^2.0.0",
     "@nomiclabs/hardhat-etherscan": "^2.1.3",
     "@nomiclabs/hardhat-waffle": "^2.0.0",
+    "@thehubbleproject/bls": "^0.5.1",
```
jacque006: Note that you can get an instance of the signer off of [BlsWalletWrapper.blsWalletSigner](https://github.com/web3well/bls-wallet/blob/main/contracts/clients/src/BlsWalletWrapper.ts#L27) if you don't want/need this lower level dependency.

contracts/contracts/VerificationGateway.sol
```diff
@@ -89,30 +89,31 @@ contract VerificationGateway
...
+        //             getInitializeData()
+        //         )
+        //     ))
+        // )))));
+    //     IWallet walletAddress = registeredWalletFromHash[hash];
+    //     if (!hasCode(address(walletAddress))) {
+    //         walletAddress = IWallet(address(0));
+    //     }
+    //     return walletAddress;
+    // }
```
jacque006: 1. Is there a reason to keep this code around? If so, might be better to move it into a docstring, other documentation, or a Solidity lib.
2. This will break [BlsWalletWrapper.Nonce](https://github.com/web3well/bls-wallet/blob/main/contracts/clients/src/BlsWalletWrapper.ts#L144) plus a number of test callpoints. We will need to update the client lib, publish a new version, and update consumers.

jacque006 (commented):
Overall core changes look good :+1:

contracts/contracts/VerificationGateway.sol
jzaki: 1. No, this is a preliminary commit
2. I don't think this should break anything since the deleted function interface `walletFromHash` is the same is the hash/key mapping, now of the same name. All smart contract tests pass unchanged

jzaki: @kautukkundan as well as the specific tests for the fix, feel free to remove this too.

jacque006: Re `#2`, that makes perfect sense, magic. Thanks ethers.js/Typechain!

contracts/package.json
kautukkundan: removed in 140f5e7, did not require any other change. the contract tests pass without this additional dependency.

jacque006 (approved):

voltrevo:
@jzaki @jacque006 I think we should generally be merging things as soon as we approve them rather than giving it back to the author to decide. Or maybe we need to talk about this policy?

(Is there a reason why this one wasn't merged last week?)

jzaki:
@voltrevo There are some things that have surfaced in my review that I'm in the process of addressing. Can share details once complete.

jzaki:
- Made room in the VerificationGateway by making proxy admin deployment external (requires setting address in constructor as per bls lib param)
- added a delay between wallet's setting of a new blskey hash (using one-to-one reverse mapping)
- removed bls key from wallet
- updated tests
Closes #104 #155

PR #221, opened 2022-06-07
Cells

voltrevo:
## [Video walkthrough](https://www.youtube.com/watch?v=0BwxE_EUcug)

[Video walkthrough](https://www.youtube.com/watch?v=0BwxE_EUcug)

## What is this PR doing?

Adds a new internal library called `cells`. See [README.md](https://github.com/web3well/bls-wallet/blob/cells/extension/source/cells/README.md) for more information.

Visit `/quillPage.html#/cells-demo` and try it out.

None specifically (but it should contribute to the resolution of a high percentage of future ui-related issues).

## Checklist

## Guidelines


extension/source/QuillPage/CellsDemo/BalanceWidget.tsx
```diff
@@ -0,0 +1,69 @@
...
+const BalanceWidget: FunctionComponent = () => {
+  const quillCtx = useQuill();
+
+  const cells = useMemo(() => {
+    const address = new MemoryCell('');
+
+    const balanceDisplay = new FormulaCell(
+      { address, _: quillCtx.blockNumber },
+      // eslint-disable-next-line @typescript-eslint/no-shadow
+      async ({ address }) => {
```
jacque006: I really, really don't like having these disables everywhere. The [no-shadow](https://eslint.org/docs/rules/no-shadow) prevents programmatic/logical errors, so I would like to leave it enabled.

I would just rename the shadows, or consider using array destructuring simlar to `React.useState`
```suggestion
      async ({ address: addr }) => {
```

extension/source/QuillPage/CellsDemo/BalanceWidget.tsx
```diff
@@ -0,0 +1,69 @@
...
+        const addressError = (() => {
+          if (!/^0x[0-9a-fA-F]{40}$/.test(address)) {
+            return new Error('Invalid address');
+          }
```
jacque006: I would just rely on the logic in `ethers.utis.getAddress` versus a custom regex. This will also make it easier for us to support new address formats in the future.
```suggestion
```

extension/source/QuillPage/CellsDemo/CellsDemoPage2.tsx
```diff
@@ -0,0 +1,13 @@
...
+export const CellsDemoPage2: FunctionComponent = () => {
+  return (
+    <DemoTable>
+      <tr>
+        <td>Hello:</td>
+        <td>World</td>
+      </tr>
+    </DemoTable>
+  );
+};
```
jacque006: Is this file needed anymore? Know it was used in the video.

extension/source/QuillPage/index.tsx
```diff
@@ -4,5 +4,8 @@ import '../ContentScript';
 import '../styles/index.scss';
 import './styles.scss';
 import QuillPage from './QuillPage';
+import Browser from 'webextension-polyfill';
+
+(window as any).Browser ??= Browser;
```
jacque006: Add this to `window.d.ts`.

```suggestion
window.Browser ??= Browser;
```

extension/source/QuillPage/CellsDemo/TextBox.tsx
```diff
@@ -0,0 +1,17 @@
...
+import useCell from '../../cells/useCell';
+
+const TextBox: FunctionComponent<{ value: ICell<string> }> = ({ value }) => {
+  const valueValue = useCell(value) ?? '';
+
+  return (
+    <input
+      type="text"
+      value={valueValue}
+      onInput={(evt) => value.write((evt.target as HTMLInputElement).value)}
```
jacque006:
```suggestion
      onInput={(evt: React.FormEvent<HTMLInputElement>) => value.write(evt.target.value)}
```

extension/source/QuillPage/QuillContext.tsx
```diff
@@ -1,21 +1,53 @@
...
   provider: QuillInPageProvider,
 ): QuillContextValue {
+  const Cell = elcc.Cell.bind(elcc);
+  const time = TimeCell(100);
+
+  const blockNumber = new FormulaCell(
+    { _: approximate(time, 5000) },
+    async () => {
+      console.log('getting block number');
+      return Number(await provider.request({ method: 'eth_blockNumber' }));
```
jacque006, line 38: You can use `ethersProvider.on("block", (blockNumber: number) => ...)` to get more realtime updates on the block number via websocket/polling and save on the rountrip RPC call. See https://docs.ethers.io/v5/api/providers/provider/#Provider--events for more info.

extension/source/QuillPage/QuillPage.tsx
```diff
@@ -3,17 +3,24 @@ import { HashRouter, Route, Routes } from 'react-router-dom';
...
-          <Route path="/wallet/*" element={<WalletPage />} />
-        </Routes>
-      </HashRouter>
+      <Theme>
+        <HashRouter>
+          <Routes>
+            <Route path="/onboarding" element={<OnboardingPage />} />
+            <Route path="/wallet/*" element={<WalletPage />} />
+            <Route path="/cells-demo" element={<CellsDemoPage />} />
+            <Route path="/cells-demo2" element={<CellsDemoPage2 />} />
```
jacque006: These should be moved to a separate index/entry point for demoing cells, or should be hidden behind a flag. Don't think most dApp devs that will use Quill would need to see them by default.

extension/source/QuillPage/TimeCell.ts
```diff
@@ -0,0 +1,17 @@
...
+export default function TimeCell(accuracy: number): IReadableCell<number> {
+  const cell = new MemoryCell(Math.floor(Date.now() / accuracy));
+
+  (async () => {
+    while (true) {
+      const now = Date.now();
+      await delay(accuracy * Math.ceil(now / accuracy) - now);
+      await cell.write(accuracy * Math.round(Date.now() / accuracy));
+    }
+  })();
```
jacque006, line 14: How does this loop terminate? Could it run indefinitely in the background until page close even if the component it is used in is unmounted?

extension/source/cells/CellCollection.ts
```diff
@@ -0,0 +1,206 @@
...
+
+import ExplicitAny from '../types/ExplicitAny';
+import ICell, { CellEmitter } from './ICell';
+import CellIterator from './CellIterator';
+import jsonHasChanged from './jsonHasChanged';
+import assert from '../helpers/assert';
+import IAsyncStorage from './IAsyncStorage';
+
+export default class CellCollection {
+  cells: Record<string, CollectionCell<ExplicitAny> | undefined> = {};
```
jacque006, line 13: Rather than using string keys, would be better to use a typed struct. That way typescript can properly point out any invalid property accesses.

```suggestion
export default class CellCollection<T> {
  cells: T;
  ...
  constuctor(cells: T) {
```

```typescript
const cc = new CellCollection<{ foo: CollectionCell<string>, bar: CollectionCell<number[]> }>();
```

extension/source/cells/CellCollection.ts
```diff
@@ -0,0 +1,206 @@
...
+    public key: string,
+    public type: io.Type<T>,
+    public defaultValue: T,
+    public hasChanged: ICell<T>['hasChanged'] = jsonHasChanged,
+  ) {
+    this.versionedType = io.type({ version: io.number, value: type });
+
+    this.initialRead = this.versionedRead().then((versionedReadResult) => {
+      this.lastSeen = versionedReadResult;
+    });
```
jacque006, line 98: Running this side effect in the constructor may cause issues/race condition in `applyType` if this read does not complete when `this.lastSeen` is checked.

jacque006 (commented):
Asides from my in code comments, I think a core choice we need to make for the extension is what kind of data management/lifecycle (model/controller) framework we want to use. Using our current controller/object state approach was a good start but we are growing beyond that. I see two main paths.  

### Pros of using cells or a custom internal data framework
- We have complete control/customization of our data modeling layer.
- Allows some ease/utility in running more complex async calculations for derived data.
- Already implemented in extension codebase :)

### Pros of using redux or a similar framework
- Heavily tested, wide use in production projects.
- Ecosystem support (middleware, add-ons, exisiting examples)
- Used by exisiting web wallets (Metamask, Tally), easier to show BLS Wallet examples to those wallet devs.
- Easier to onboard new developers, accept external contributions with understanding of flux patterns.

### Some other general thoughts
 - We should use properties rather than string keys to access cells/data so we can leverage TS to enforce type usage.
 - Using `io-ts` is handy when we need to accept external inputs or validate rehydrated data from storage, but seems like overkill to use everywhere internally when can use TS to control read/write types. If will also add runtime overhead to check those types over and over again.
 - I like the versioned data, and think we should stick with that pattern when loading from storage.
 - There are a ton of commits (> # of files changed) in this branch that will flood our commit history. Worth merging as a rebase or consolidating some commits.

extension/source/QuillPage/QuillContext.tsx
voltrevo: Yeah I'd hoped something like that would work but it turns out it's just polling too. I tested it and it's simply generating an `eth_blockNumber` call every 4 seconds. It also exposes events for in-between blocks when multiple blocks have come in in that 4 second window.

We also have our own block tracking in the background script and this is already deleted in followup changes to use that instead anyway.

extension/source/QuillPage/CellsDemo/CellsDemoPage2.tsx
voltrevo: Nah. I figured we'd delete it after these changes are settled so that during review there's this quick starting point.

extension/source/QuillPage/CellsDemo/BalanceWidget.tsx
voltrevo: Yeah I touched on this in the video. The awkward thing is that this will lead to artificial renames/synonyms similar to your suggestion, which is also its own antipattern.

I have an idea. Will follow-up.

extension/source/QuillPage/CellsDemo/BalanceWidget.tsx
voltrevo: Will do.

extension/source/QuillPage/index.tsx
voltrevo: Will do.

extension/source/QuillPage/CellsDemo/TextBox.tsx
voltrevo: Ack. I keep forgetting the right way to do this. Will fix.

extension/source/QuillPage/QuillPage.tsx
voltrevo: Hmm, yeah. I'll put it in the `cells` dir.

extension/source/QuillPage/TimeCell.ts
voltrevo: It doesn't lol. Time never stops right?

Yeah it's not ideal 😄. I think we can address this later though since this is only used in the demo. I'm thinking about making a specialized cell for time+accuracy which will update lazily in a similar way to a FormulaCell to avoid this problem.

extension/source/cells/CellCollection.ts
voltrevo: We could, but I think that's unnecessary complexity actually.

Instead, if you have concrete types for the cells you want to use, I think it makes more sense to simply expose concrete cells like this:

```ts
const cc = new CellCollection(/* ... */);

const appCells = {
  foo: cc.Cell('foo', io.string, ''),
  bar: cc.Cell('bar', io.array(io.number), []),
};
```

This way you don't need to duplicate the type information, since you'll need to construct the runtime types at some point anyway.

Additionally, this allows for unrelated subsystems to share the same `CellCollection` without sharing any code. This is advantageous because `CellCollection` is usually some singleton-like persistent storage and `CellCollection` will use the same `CollectionCell` instance when using the same key which helps reduce overhead and checks the same type is being used if the `CollectionCell` is accessed by name in multiple places.

extension/source/cells/CellCollection.ts
voltrevo: I think this is ok. The checks in `applyType` are just proactive checks to help raise errors earlier/closer to where the mistake is made. Your app will still break either way if the types are not consistent.

If `initialRead` completes before `applyType` then you'll get the benefit of `applyType` throwing an exception if the concrete value doesn't match that type.

If `applyType` is run before `initialRead` completes then `initialRead` will be rejected if the concrete value doesn't match the type (which will show in tools/console) and all writes will also fail with this error since they have `await this.initialRead;`.

voltrevo:
> ### Pros of using redux or a similar framework
> * Heavily tested, wide use in production projects.
> * Ecosystem support (middleware, add-ons, exisiting examples)
> * Used by exisiting web wallets (Metamask, Tally), easier to show BLS Wallet examples to those wallet devs.
> * Easier to onboard new developers, accept external contributions with understanding of flux patterns.

While I acknowledge the above points, my dominant thought here is that I'd actually prefer to use an existing library if it allows us to more easily express the same ideas:
- Share a reference to an identity with interior mutability (not a sequence of updates)
- Access to the internal value is asynchronous
- The value is set to a default if it doesn't exist, but this is *not* a substitute for having not yet loaded
- Built for TypeScript, with types derived from runtime type information and types checked at runtime
- A way to compose these identities with a function to create a read-only identity which is only executed when needed

To my knowledge, there isn't a good substitute. Indeed, this is the reason why I'm rather excited about cells - I think this is a really compelling abstraction that has potential to inspire others to build upon it and bring attention to blswallet.

Additionally, I think there is a psychological factor to consider here. I'm a human, not a robot, and I'm the most active contributor to the extension (in recent times, although it's probably true in the lifetime of the extension too). My productivity is linked to being passionate, and I'm passionate about this. It's not a very good reason, and I would never ask anyone to accept my code if I didn't also believe it was separately a good idea, but I believe this reason is worth factoring in.

> * Using `io-ts` is handy when we need to accept external inputs or validate rehydrated data from storage, but seems like overkill to use everywhere internally when can use TS to control read/write types. If will also add runtime overhead to check those types over and over again.

I guess I'm a bit paranoid at the moment because of the amount of inaccurate types we have flying around because of https://github.com/web3well/bls-wallet/pull/42. I think the overhead is worth it for now, and indeed the overhead is kindof a non-issue. Happy to chat more about it though.

> * There are a ton of commits (> # of files changed) in this branch that will flood our commit history. Worth merging as a rebase or consolidating some commits.

I've never quite understood this. Having the detail is useful when things go sideways and you want to explore how the code came to be. You can still look at the pull requests or merge commits if you want a coarse/curated history instead.

I've always preferred keeping the original history. Having said that, I don't mind too much if the team prefers squashing.

Oh and another thought - I believe that build artifacts should reference the git sha, allowing you to see what they were built from. Squashing causes these to be deleted. [We actually ran into this previously with bls-wallet-clients.](https://github.com/web3well/bls-wallet/pull/92#discussion_r790093363)

extension/source/QuillPage/CellsDemo/BalanceWidget.tsx
voltrevo: Here's my proposed resolution: https://github.com/web3well/bls-wallet/pull/239.

In summary, it uses [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) to apply a consistent rename internally and allow you to use the renames in a type-checked way:

```ts
const ab = new FormulaCell({ a, b }, ({ $a, $b }) => $a * $b);
```

extension/source/QuillPage/CellsDemo/BalanceWidget.tsx
voltrevo: Done.

extension/source/QuillPage/index.tsx
voltrevo: Done.

extension/source/QuillPage/CellsDemo/TextBox.tsx
voltrevo: Done.

extension/source/QuillPage/QuillPage.tsx
voltrevo: Done.

extension/source/QuillPage/QuillContext.tsx
jacque006: Gotcha. It looks like if we wanted the to keep more in sync, there is a [WebSocketProvider](https://docs.ethers.io/v5/api/providers/other/#WebSocketProvider) we could use instead, but there is some backend overhead associated. Could be worth adding in on protocol (`ws://` or `wss://`) detection from RPC URL, but definitely out of scope for this work.

extension/source/QuillPage/TimeCell.ts
jacque006: Will be addressed in followup work, not huge risk ATM so leaving as is in this work.

jacque006:
Summary from offline conversation with @voltrevo :
- Using cells over another framework as @voltrevo is very passionate about it's use and thinks we can use it effectively without slowing down/complicating new devs onboarding or external contributions. We may eventually extract it out of this repo if it gets general enough.
- There is ongoing followup work to optimize or fix a number of mentioned issues in follow up PRs
- Unit tests will be added in future work.
- We will consider in future iterations being a little less heavy on `io-ts` in certain areas and in parts of the runtime checks as makes sense.
- This PR will be squashed to limit # of commits from this PR.

I will create some follow up issues for the above.

jacque006 (approved):

PR #239, opened 2022-06-15
Use prefixKeys to avoid shadowing

voltrevo:
## What is this PR doing?

Addresses feedback on other PR: https://github.com/web3well/bls-wallet/pull/221#discussion_r896956229.

## How can these changes be manually tested?

Use the cells demo page.

## Does this PR resolve or contribute to any issues?

Nope.

## Checklist
- [x] I have manually tested these changes
- ~[ ] Post a link to the PR in the group chat~ NA: Linked in other PR instead

## Guidelines
- If your PR is not ready, mark it as a draft
- The `resolve conversation` button is for reviewers, not authors
  - (But add a 'done' comment or similar)

Issue #171, opened 2022-03-23
Update extension to use manifest v3


jacque006:
`Manifest version 2 is deprecated, and support will be removed in 2023. See https://developer.chrome.com/blog/mv2-transition/ for more details.`

![Screenshot from 2022-03-23 18-13-21](https://user-images.githubusercontent.com/10394161/159804948-8bd4df55-bc13-41b4-a34a-4d173ece3ca5.png)

jacque006:
This is not a high priority now, but wanted to document so we can prioritize closer to the end of the year.

jacque006:
Note, we will not be able to use our injected background script in extension manifest v3, would have to be a service worker https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/

jacque006:
Deprecation of v2 manifests has been delayed until at least March https://groups.google.com/u/1/a/chromium.org/g/chromium-extensions/c/zQ77HkGmK9E

PR #243, opened 2022-06-21
The Big One

### LINKED ISSUE #225: Bug: Opening the extension resets the selected account to the first one

voltrevo:
# Whoa, what's going on?

Yeah, I understand if this is a bit alarming:
Screen Shot 2022-06-22 at 9 21 49 am

This is not a normal PR. It's a huge refactor. I would not have done this if the extension was being actively developed by anyone else on the team.

The thing is, all this functionality could not be reviewed properly in its previous form. It was merged even though it mostly did not work. None of the team members properly absorbed how it worked. It was also subjectively a mess (I believe this is the dominant opinion of the team, not just my own). I now know how all that code worked, but only because I've replaced almost all of it incrementally and gone through it with fine tooth comb to check whether I missed anything important.

I believe the replacement code here is of very high quality. I'm so happy with how it all turned out, and can't wait to do some deep dives to show you and others the simplicity of our design, and how it provides developers with an awesome foundation to build new features.

## RPC has been completely replaced

<details>
<summary>with a system that is much smaller, simpler, properly typed, and so much easier for developers to use</summary>

No more JRPC, no more port streams, pumping, duplexes, object multiplexers, muxes, or middleware. All unnecessary complexity.

Instead we just use the standard browser apis directly: `window.postMessage` between dApp and content script, and `runtime.sendMessage` between content script and background script.

We can't avoid doing that hop through the content script so there's a certain amount of complexity that's still needed, but the communication is handled in under 300 lines of code. (And I haven't actively tried to make the code small, it's really just not doing very much.)

Now, if you wanted to add a new method, let's call it `prDemoMethod`, the diff should look like this:

```diff
diff --git a/extension/source/background/QuillController.ts b/extension/source/background/QuillController.ts
index c5048ac..dec9d43 100644
--- a/extension/source/background/QuillController.ts
+++ b/extension/source/background/QuillController.ts
@@ -160,6 +160,10 @@ export default class QuillController {
         };
       },
 
+      prDemoMethod: async ({ params: [a, b] }) => {
+        return `Distance: ${(a ** 2 + b ** 2) ** 0.5}`;
+      },
+
       setSelectedAddress: this.preferencesController.rpc.setSelectedAddress,
       longPoll: this.longPollingController.rpc.longPoll,
       longPollCancel: this.longPollingController.rpc.longPollCancel,
diff --git a/extension/source/types/Rpc.ts b/extension/source/types/Rpc.ts
index e027423..c2dabad 100644
--- a/extension/source/types/Rpc.ts
+++ b/extension/source/types/Rpc.ts
@@ -81,6 +81,11 @@ export const rpcMap = {
     Params: io.tuple([io.string, io.number, io.string]),
     Response: io.unknown,
   },
+  prDemoMethod: {
+    origin: '*', // <--- Use '<quill>' if dApps aren't allowed to call it
+    Params: io.tuple([io.number, io.number]),
+    Response: io.string,
+  },
   addAccount: {
     origin: '<quill>',
     Params: io.tuple([io.union([io.undefined, io.string])]),
```

Optionally, if you also wanted your method to be handled by one of the controllers, apply a followup change like this:

```diff
diff --git a/extension/source/background/KeyringController.ts b/extension/source/background/KeyringController.ts
index 4c439cd..b24d0ab 100644
--- a/extension/source/background/KeyringController.ts
+++ b/extension/source/background/KeyringController.ts
@@ -88,6 +88,10 @@ export default class KeyringController {
       return address;
     },
 
+    prDemoMethod: async ({ params: [a, b] }) => {
+      return `Distance: ${(a ** 2 + b ** 2) ** 0.5}`;
+    },
+
     removeAccount: async ({ params: [address] }) => {
       const { wallets } = await this.keyring.read();
       const newWallets = wallets.filter((w) => w.address !== address);
diff --git a/extension/source/background/QuillController.ts b/extension/source/background/QuillController.ts
index dec9d43..aac1440 100644
--- a/extension/source/background/QuillController.ts
+++ b/extension/source/background/QuillController.ts
@@ -160,9 +160,7 @@ export default class QuillController {
         };
       },
 
-      prDemoMethod: async ({ params: [a, b] }) => {
-        return `Distance: ${(a ** 2 + b ** 2) ** 0.5}`;
-      },
+      // All rpc starts in QuillController, so we still delegate to KeyringController here
+      prDemoMethod: this.keyringController.rpc.prDemoMethod,
 
       setSelectedAddress: this.preferencesController.rpc.setSelectedAddress,
       longPoll: this.longPollingController.rpc.longPoll,
```

The type information is helpfully different between `QuillController` and other controllers.

QuillController's `.rpc` implements `RpcImpl`, so it must define every rpc method and always return the type specified (or throw an exception).

Other controllers' `.rpc` implement `PartialRpcImpl`, so they are free to define their appropriate subset. Additionally, they can return `undefined` instead of the required type, which will force (via static inference) `QuillController` to handle that case after the attempted delegation.

</details>

## RPC cells (aka long polling cells)

<details>
<summary>use a similar typing strategy and are pretty cool too</summary>

We're using this for a few things:
- Communicating the `providerState` to the dApp, so it knows if it's connected and can emit `chainChanged` etc
- `blockNumber`
- `currencyConversion`

You can now share a cell maintained by the backend without having to store it.

As an example, we're going to add a lovely weather indicator to our wallet (not included in the actual changes):

Screen Shot 2022-06-27 at 5 56 08 pm

Since we're using cells, we expect this data to never be stale. We'd rather it was blank, thanks. Also, definitely don't actively fetch this, only get it when the user needs it. Oh yeah and if you could also make it so that you could show it in multiple places and even multiple tabs without having to fetch it multiple times, that'd be great.

To do this we're going to use ip-api.com to estimate the current location and then use that with weatherapi.com.

Inspecting the docs of these apis, I figure I can get city, temperature, and a description of the current conditions (among many other things). So I'll start by defining our cell:

```diff
diff --git a/extension/source/types/LongPollingCells.ts b/extension/source/types/LongPollingCells.ts
index 6ac7b76..a166ade 100644
--- a/extension/source/types/LongPollingCells.ts
+++ b/extension/source/types/LongPollingCells.ts
@@ -18,6 +18,14 @@ export const longPollingCellMap = {
     origin: '<quill>',
     Type: io.union([io.undefined, io.number]),
   },
+  weather: {
+    origin: '*',
+    Type: io.type({
+      city: io.string,
+      condition: io.string,
+      temperature: io.number,
+    }),
+  },
 };
 
 export type LongPollingCellMap = typeof longPollingCellMap;
```

If you're running TypeScript, it'll helpfully point out that `QuillController` is not passing the required `weather` cell to `LongPollingController`, so let's do that:

```diff
diff --git a/extension/source/background/QuillController.ts b/extension/source/background/QuillController.ts
index c5048ac..eb04a8b 100644
--- a/extension/source/background/QuillController.ts
+++ b/extension/source/background/QuillController.ts
@@ -29,6 +29,7 @@ import RandomId from '../helpers/RandomId';
 import mapValues from '../helpers/mapValues';
 import LongPollingController from './LongPollingController';
 import isPermittedOrigin from './isPermittedOrigin';
+import TimeCell from '../cells/TimeCell';
 
 export default class QuillController {
   networkController: NetworkController;
@@ -89,6 +90,56 @@ export default class QuillController {
       blockNumber: this.networkController.blockNumber,
       providerState: this.cells.providerState,
       currencyConversion: this.currencyConversion,
+      // You might want to put this in its own file, but for demo purposes I'm
+      // just inlining it here
+      weather: new FormulaCell(
+        {
+          // Weather doesn't change very often, right? 😅 (Poll once every 10m)
+          time: TimeCell(600_000),
+        },
+        async () => {
+          const ipData = await fetch('http://ip-api.com/json').then((res) =>
+            res.json(),
+          );
+
+          assertType(
+            ipData,
+            io.type({
+              status: io.literal('success'),
+              city: io.string,
+              lat: io.number,
+              lon: io.number,
+            }),
+          );
+
+          const weatherData = await fetch(
+            [
+              `https://api.weatherapi.com/v1/current.json?`,
+              `key=a976bb0a789f4434bc472841222706&`,
+              `q=${ipData.lat},${ipData.lon}&`,
+              `aqi=no`,
+            ].join(''),
+          ).then((res) => res.json());
+
+          assertType(
+            weatherData,
+            io.type({
+              current: io.type({
+                temp_c: io.number,
+                condition: io.type({
+                  text: io.string,
+                }),
+              }),
+            }),
+          );
+
+          return {
+            city: ipData.city,
+            condition: weatherData.current.condition.text,
+            temperature: weatherData.current.temp_c,
+          };
+        },
+      ),
     });
 
     this.rpc = {
```

To actually display this on the frontend, we need to use `QuillLongPollingCell` which uses the special `longPoll` and `longPollCancel` RPC endpoints:

```diff
diff --git a/extension/source/Home/QuillContext.tsx b/extension/source/Home/QuillContext.tsx
index e20e9be..2882f7e 100644
--- a/extension/source/Home/QuillContext.tsx
+++ b/extension/source/Home/QuillContext.tsx
@@ -75,6 +75,8 @@ function QuillContextCells(
     'rpcLogging',
   );
 
+  const weather = QuillLongPollingCell(ethereum, 'weather');
+
   return {
     ...storageCells,
     networkJson: new FormulaCell(
@@ -85,5 +87,11 @@ function QuillContextCells(
     rpcBackgroundLogging: TransformCell.Sub(rpcLogging, 'background'),
     rpcInPageLogging: TransformCell.Sub(rpcLogging, 'inPage'),
     currencyConversion: QuillLongPollingCell(ethereum, 'currencyConversion'),
   };
 }
```

We're going to want to simply render this as text for now, so let's have the frontend also convert it to the string we want using a FormulaCell:

```diff
diff --git a/extension/source/Home/QuillContext.tsx b/extension/source/Home/QuillContext.tsx
index e20e9be..2882f7e 100644
--- a/extension/source/Home/QuillContext.tsx
+++ b/extension/source/Home/QuillContext.tsx
@@ -75,6 +75,8 @@ function QuillContextCells(
     'rpcLogging',
   );
 
   const weather = QuillLongPollingCell(ethereum, 'weather');

   return {
     ...storageCells,
     networkJson: new FormulaCell(
@@ -85,5 +87,11 @@ function QuillContextCells(
     rpcBackgroundLogging: TransformCell.Sub(rpcLogging, 'background'),
     rpcInPageLogging: TransformCell.Sub(rpcLogging, 'inPage'),
     currencyConversion: QuillLongPollingCell(ethereum, 'currencyConversion'),
+    weather,
+    weatherText: new FormulaCell(
+      { weather },
+      // eslint-disable-next-line @typescript-eslint/no-shadow
+      ({ weather }) => `${weather.condition}, ${weather.temperature}°C`,
+    ),
   };
 }
```

Now to actually show it in our ui, just `Display` it in `QuillHeading`:

```diff
diff --git a/extension/source/components/QuillHeading.tsx b/extension/source/components/QuillHeading.tsx
index 0be7c61..da74f67 100644
--- a/extension/source/components/QuillHeading.tsx
+++ b/extension/source/components/QuillHeading.tsx
@@ -1,7 +1,11 @@
 import * as React from 'react';
 import { runtime } from 'webextension-polyfill';
+import Display from '../cells/components/Display';
+import { useQuill } from '../Home/QuillContext';
 
 const QuillHeading = (): React.ReactElement => {
+  const { cells } = useQuill();
+
   return (
     <>
       <div className="flex justify-center">
@@ -12,6 +16,10 @@ const QuillHeading = (): React.ReactElement => {
           height="60"
         />
       </div>
+
+      <div style={{ textAlign: 'center' }}>
+        <Display cell={cells.weatherText} />
+      </div>
     </>
   );
 };
```

</details>

## RPC intellisense

<details>
<summary>is now amazing</summary>

You now get great intellisense when authoring your method:

Screen Shot 2022-06-27 at 10 09 05 am

And TypeScript isn't getting that from the `a ** 2` usage below. Indeed, if you had typed `a` as a `string` in `Rpc.ts` instead:

Screen Shot 2022-06-27 at 10 11 41 am

Screen Shot 2022-06-27 at 10 12 23 am

When working on the wallet frontend, you'll get intellisense like this:

Screen Shot 2022-06-27 at 10 15 50 am

Screen Shot 2022-06-27 at 10 16 32 am

If you choose to use `ethereum.request` on the wallet frontend, the intellisense is not as good, but it can still be quite helpful:

Screen Shot 2022-06-27 at 10 20 00 am

In fact, I've also added `ethereum.rpc` as an optional api you can expose to dApps. If you do, you'll even get some intellisense when using dev tools:

Screen Shot 2022-06-27 at 10 26 26 am

Compare this to what you get using the standard api:

Screen Shot 2022-06-27 at 9 08 09 am

Here's what it looks like to actually call the method in dev tools:

Screen Shot 2022-06-27 at 10 30 25 am

And if you pass invalid arguments:

Screen Shot 2022-06-27 at 10 31 35 am

Note that this is thrown inside `ethereum.bundle.js` because it knows about `Rpc.ts` and doesn't need to wait for the background script to tell it that the arguments were wrong (which dramatically simplifies the call stack and means it'll still work even if the connection is severed).

The benefits of TypeScript are extraordinary when you use it this way. That's why I push for thorough typing so much.

</details>

## Reduced transpilation

<details>
<summary>emulating really old browsers makes really yucky code</summary>

We used to support all the way back to Chrome 49. Chrome 49 came out 6 years ago.

All modern browsers now force you to update pretty darn frequently, so we now only support Chrome and Firefox versions up to one year old. We also had Opera and Edge in our babelrc before but these have now been Chromium browsers for a long time so they aren't different from a transpilation perspective.

Before:

```js
  update(updates) {
    var _this3 = this;

    return _asyncToGenerator( /*#__PURE__*/_babel_runtime_regenerator__WEBPACK_IMPORTED_MODULE_0___default.a.mark(function _callee3() {
      return _babel_runtime_regenerator__WEBPACK_IMPORTED_MODULE_0___default.a.wrap(function _callee3$(_context3) {
        while (1) switch (_context3.prev = _context3.next) {
          case 0:
            _context3.t0 = _this3;
            _context3.t1 = _objectSpread;
            _context3.t2 = _objectSpread;
            _context3.t3 = {};
            _context3.next = 6;
            return _this3.read();

          case 6:
            _context3.t4 = _context3.sent;
            _context3.t5 = (0, _context3.t2)(_context3.t3, _context3.t4);
            _context3.t6 = updates;
            _context3.t7 = (0, _context3.t1)(_context3.t5, _context3.t6);
            _context3.next = 12;
            return _context3.t0.write.call(_context3.t0, _context3.t7);

          case 12:
          case "end":
            return _context3.stop();
        }
      }, _callee3);
    }))();
  }
```

After:

```js
  async update(updates) {
    await this.write({ ...(await this.read()),
      ...updates
    });
  }
```

</details>

## Sourcemaps

<details>
<summary>are life changing</summary>

If you debug `QuillController`'s constructor in Chrome:

Before:

Screen Shot 2022-06-27 at 5 04 11 pm

After:

Screen Shot 2022-06-27 at 5 05 46 pm

</details>

## RPC logging

<details>
<summary>has gone from useless to awesome</summary>

Before:

Screen Shot 2022-06-27 at 1 07 13 pm
(^^^ background script only, nothing in the dApp)

After:

Screen Shot 2022-06-27 at 9 16 47 am
(^^^ background script)

Screen Shot 2022-06-27 at 9 14 08 am
(^^^ inside the dApp)

</details>

## Exception logging

<details>
<summary>has been dramatically improved</summary>

Before:

If we added this code in the old system:

```ts
      debugMe: async (_origin, [a, b, c]) => {
        console.log('debugMe', { a, b, c });

        throw new Error('boom'); // <--- Added

        return {
          // This special debugging method uses io.unknown for Response, which
          // is what enables returning anything here. All real methods should
          // use precise typing.
          ahoy: `Looks like you sent ${[a, b, c]}`,
        };
      },
```

then it would show you this:

Screen Shot 2022-06-27 at 9 38 29 am

After:

Screen Shot 2022-06-27 at 9 21 58 am

(This also works when using `ethereum.request({ ... })`)

Note that there are actually two errors shown - one in the content script and one inside the actual dApp. The one inside the actual dApp is exposed to the dApp if it catches the exception, which is why the detail is hidden. Since the content script is our own isolated context we can safely log the full detail there and the developer can still see it in the dApp console.

The code for the new system is slightly different:

```ts
      debugMe: async ({ params: [a, b, c] }) => {
        console.log('debugMe', { a, b, c });

        throw new Error('boom'); // <--- Added

        return {
          // This special debugging method uses io.unknown for Response, which
          // is what enables returning anything here. All real methods should
          // use precise typing.
          ahoy: `Looks like you sent ${[a, b, c]}`,
        };
      },
```

</details>

## Developer Settings

<details>
<summary>have been added</summary>

Screen Shot 2022-06-27 at 10 57 35 am

</details>

## Assertion failures

<details>
<summary>will very often make mistakes 10x faster to diagnose</summary>

There are now 64 places we call `assert` or `assertType`. For example:

```ts
    lookupPrivateKey: async ({ params: [rawAddress] }) => {
      const address = ethers.utils.getAddress(rawAddress);

      const keyPair = (await this.keyring.read()).wallets.find(
        (x) => x.address === address,
      );

      assert(keyPair !== undefined, () => new Error('key does not exist')); // <--- Awesome

      return keyPair.privateKey;
    },
```

If you have the dev tools open for Quill's background page and turn on `Break on assertion failures` in Quill's new Developer Settings (well, it defaults to enabled for now actually):

Screen Shot 2022-06-27 at 1 24 26 pm

then if you try to send a transaction from an address we don't have:

Screen Shot 2022-06-27 at 1 33 42 pm

the background script dev tools will be brought to foreground paused on this `debugger` statement:

Screen Shot 2022-06-27 at 1 33 23 pm

then just click the next frame of the call stack:

Screen Shot 2022-06-27 at 1 37 31 pm

a really nice further improvement here would be fixing our source maps. The transpiled version isn't too bad though.

EDIT: I have fixed sourcemaps 🎉.

</details>

## ethereum.bundle.js

<details>
<summary>is a much more appropriate name, and we no longer obscure its existence</summary>

Before:

Screen Shot 2022-06-27 at 1 48 46 pm
(if it wasn't being unhelpfully removed, you'd find `pageContentScript.bundle.js`, which would still be confusing)

After:

Screen Shot 2022-06-27 at 1 53 21 pm

</details>

## Smarter polling

<details>
<summary>we used to poll slowly all day long, now we poll fast when we need it</summary>

Before:

If you open the network tab of the background script without doing anything else (don't even open the wallet page), then you'll see this:

Screen Shot 2022-06-27 at 2 07 31 pm

This is bad for 3 reasons:
1. It's unnecessary clutter when using the network tab
2. It harasses these servers all day long for no reason
3. It updates way too slowly when the user is actually interested in it

After:

Screen Shot 2022-06-27 at 2 16 45 pm

and if you visit Settings -> Network:

Screen Shot 2022-06-27 at 2 23 55 pm

Screen Shot 2022-06-27 at 2 23 30 pm

and if you switch to Settings -> Security:

Screen Shot 2022-06-27 at 2 23 30 pm

(this will also happen if you close the tab, and if you opened two tabs viewing network settings, it would still poll just once per 4 seconds, not twice per 4 seconds)

now switch to Settings -> General:

Screen Shot 2022-06-27 at 2 36 23 pm

Screen Shot 2022-06-27 at 2 42 01 pm

back to Settings -> Security

Screen Shot 2022-06-27 at 2 23 30 pm

getting these smart polling benefits is all seamlessly handled by cells - the General Settings stub has no idea that tracking is being turned on and off, it just displays the cell:

```ts
const GeneralSettings: FunctionComponent = () => {
  const { cells } = useQuill();

  return (
    <>
      {/* TODO: Currency selection */}
      Currency conversion rate: <Display cell={cells.currencyConversion} />
    </>
  );
};
```

Oh and also, the block number and currency conversion rate are now just cells, replacing `PollingBlockTracker`, `BaseBlockTracker`, and `CurrencyController`, so it's a lot less code too.

`blockNumber` is simply this inside `NetworkController`:

```ts
    this.blockNumber = new FormulaCell(
      {
        network,
        time: TimeCell(4_000),
      },
      () => this.fetchBlockNumber(),
    );
```

</details>

## `debug` globals

<details>
<summary>are super handy when using dev tools</summary>

In dev tools, there's now a global called `debug` in the background script and homepage contexts.

background script:

Screen Shot 2022-06-27 at 7 35 01 pm

- `Browser`: The webextensions polyfill so you can use promise versions of the special extension-only browser apis
- `quillController`: The main class that governs the background script
- `reset`: Resets the extension, wiping all the data and sending you back to the onboarding flow (shows a warning first)
- `storageCells`: All the stored data in the form of cells

homepage:

Screen Shot 2022-06-27 at 7 38 28 pm

- `Browser`: The webextensions polyfill so you can use promise versions of the special extension-only browser apis
- `quill`: The thing you get when calling `useQuill()`, ie the quill context

</details>

## Great docstrings

<details>
<summary>worth a read</summary>

In `assertType.ts`:
https://github.com/web3well/bls-wallet/blob/more-cleanup/extension/source/cells/assertType.ts#L6-L10
https://github.com/web3well/bls-wallet/blob/more-cleanup/extension/source/cells/assertType.ts#L26-L53
https://github.com/web3well/bls-wallet/blob/more-cleanup/extension/source/cells/assertType.ts#L59-L75

`forEach` (for cells):
https://github.com/web3well/bls-wallet/blob/more-cleanup/extension/source/cells/forEach.ts#L4-L15

All about the content script:
https://github.com/web3well/bls-wallet/blob/more-cleanup/extension/source/contentScript.ts#L1-L35

`assert`:
https://github.com/web3well/bls-wallet/blob/more-cleanup/extension/source/helpers/assert.ts#L5-L32

`ensureType`:
https://github.com/web3well/bls-wallet/blob/more-cleanup/extension/source/helpers/ensureType.ts#L1-L108

</details>

## Other

<details>
<summary>It's hard to express how many nice little things have also changed</summary>

- We're no longer using these messy and unnecessary base classes: `BaseController`, `BaseProvider`
- `AccountTrackerController` has been completely removed (we weren't using it and it didn't do anything of significance)
- We're making good use of cells, not just replacing the storage (e.g. the currency rate is just a cell, not a controller)
- The frontend now has access to non-storage cells (currently `blockNumber`, `currencyConversion`, and `providerState`) via a long-polling system with a typed interface similar to the RPC typing
- `Controllers` has been renamed to `background` (since they are really about driving the background script and there is important non-controller code in here) and no longer has any nested directories
- The content script doesn't have any directory nesting, doesn't use any exotic dependencies, and has an awesome docstring explaining how it'd be nice to actually remove the content script entirely
- `chainId` is now stored in one place instead of three
- There's a new kind of cell called `TransformCell` which allows you to expose something inside a cell as a cell by providing mapping functions for reading and writing. Most commonly this is used to expose a property of a cell, and so there's a convenience method called `TransformCell.Sub` for that. There's also `FormulaCell.Sub` if you only need to read.
- Setting the preferred aggregator and tracking aggregator receipts is now done properly in an `AggregatorController` instead of the `knownTransactions` thing I did before (because of the opaqueness of the old system)
- quillPage.html is now home.html
- I've commented out the popup and made it just open the wallet/homepage, since we we're not currently using it for anything and it's annoying to go through that popup all the time when you really want the homepage
- Buttons now automatically use their loading state if their click handler returns a promise
- `TimeCell` no longer updates proactively
- `strictPropertyInitialization` is now turned on (which means TypeScript will correctly tell you when you're not initializing your fields)
- When you click 'confirm secret phrase' during onboarding the button enters a loading state until its action (setting up the wallet) is completed, followed by the navigation
- When you click 'Add Wallet' the button enters a loading state as above, and when adding is complete the page no longer reloads, but the new wallet is displayed correctly
- If you have another instance of the quillPage open, a new wallet will appear automatically when added by the other instance
- Selected address is now in sync with the actual underlying state and syncs between open windows as above

</details>

Give everything in the extension a once-over. You can also try out the features and improvements I've described above.

Resolves #225, #226, #227

(But that really only scratches the surface.)

# Checklist

# Guidelines


extension/yarn.lock
```diff
@@ -1592,7 +1592,7 @@
 
 "@ethersproject/wallet@5.5.0":
   version "5.5.0"
-  resolved "https://registry.yarnpkg.com/@ethersproject/wallet/-/wallet-5.5.0.tgz#322a10527a440ece593980dca6182f17d54eae75"
+  resolved "https://registry.yarnpkg.com/@ethersproject/wallets/-/wallet-5.5.0.tgz#322a10527a440ece593980dca6182f17d54eae75"
```
voltrevo: Oops.

jacque006:
@voltrevo Appreciate the gusto in this refactor, but this is going to be difficult to review. Possible it can be broken up into multiple, smaller chunks?

voltrevo:
> @voltrevo Appreciate the gusto in this refactor, but this is going to be difficult to review. Possible it can be broken up into multiple, smaller chunks?

Quite possibly, although the way to review this is probably going to be going over all the relevant parts from scratch. I know that's a lot, but it's actually nice to finally dive into this functionality.

I'm also going to put a significant effort into documenting these changes.

voltrevo:
@jacque006 I said before that I'd have updates today, buuut I'm gonna defer til next week instead. Just got done with what I needed to do with my other job and I've only got Friday left now. Time to take a rest. Still happy to chat sooner if you wish.

extension/.babelrc
```diff
@@ -8,37 +8,16 @@
         // Do not transform modules to CJS
         "modules": false,
         "targets": {
-          "chrome": "49",
-          "firefox": "52",
-          "opera": "36",
```
jacque006, line 13: If we're dropping Opera as a target, should remove from `package.json` build commands, and CD as well. We can do this in another issue/PR, just wanted to note here.

extension/package.json
```diff
@@ -105,22 +97,22 @@
...
-    "resolve-url-loader": "^3.1.3",
-    "sass-loader": "^10.2.0",
+    "resolve-url-loader": "^5.0.0",
+    "sass-loader": "^13.0.1",
     "tailwindcss": "^3.0.13",
-    "terser-webpack-plugin": "^4.2.3",
-    "webpack": "^4.46.0",
-    "webpack-cli": "^4.9.1",
+    "terser-webpack-plugin": "^5.3.3",
+    "webpack": "~5.46.0",
```
jacque006, line 112: :1st_place_medal: :100: MVP

extension/source/Home/Onboarding/OnboardingActionPanel.tsx
```diff
@@ -8,11 +8,14 @@ import WorkflowNumbers from './WorkflowNumbers';
...
   const [params, setParams] = useSearchParams();
-  const [pageIndex, setPageIndex] = useState(1);
+  const [pageIndex, setPageIndex] = useState(0);
 
   useEffect(() => {
-    const currentPage = Number(params.get('p'));
-    setPageIndex(currentPage - 1);
+    const p = params.get('p');
+
+    if (p !== null) {
```
jacque006, line 17: Any chance this could come back as `undefined` or an empty string?
```
Number(undefined)  === NaN
Number("") === 0 // !!"" === true, probably ok that we set on this case
```

```suggestion
    if (p) {
```

extension/source/Home/Onboarding/WorkflowNumbers.tsx
```diff
@@ -10,8 +10,12 @@ const WorkflowNumbers: FunctionComponent<{
   const [currentPage, setCurrentPage] = useState(1);
 
   useEffect(() => {
-    const page = Number(params.get('p'));
-    setCurrentPage(page);
+    // TODO: use hash, deduplicate page detection with other components
```
jacque006: Is this `TODO` also valid for other uses of `p` query param in other components?

extension/source/helpers/RandomId.ts
```diff
@@ -0,0 +1,5 @@
+import { encode } from 'bs58check';
+
+export default function RandomId() {
+  return encode(Buffer.from(crypto.getRandomValues(new Uint8Array(16))));
```
jacque006, line 4: We probably don't need something this secure/random since this is used in request/rpc ids, unless I'm missing something?

extension/source/components/Button.tsx
```diff
@@ -1,21 +1,37 @@
...
+      onClick={async () => {
+        const pressResult = props.onPress();
+
+        if (pressResult instanceof Promise) {
+          setPressLoading(true);
+
+          try {
+            await pressResult;
+          } catch {
+            // Enhancement: Give the button an error state
```
jacque006: I think this will swallow the error, would be good to log.

```suggestion
          } catch(err) {
            // Enhancement: Give the button an error state
            console.error(err);
```

extension/source/cells/CellCollection.ts
```diff
@@ -146,9 +143,10 @@ export class CollectionCell<T> implements ICell<T> {
 
     const latest = await this.versionedRead();
 
-    if (!(newVersionedValue.version > latest.version)) {
-      throw new Error('Rejecting write which is not newer than remote');
-    }
+    assert(
+      newVersionedValue.version > latest.version,
+      () => new Error('Rejecting write which is not newer than remote'),
```
jacque006:
```suggestion
      () => new Error(`Rejecting write which is not newer than remote. write v: ${newVersionedValue.version}, remote v: ${latest.version}`),
```

extension/source/QuillEthereumProvider.ts
```diff
@@ -0,0 +1,191 @@
...
+        params: body.params ?? [],
+      };
+
+      window.postMessage(request, '*');
+
+      const backgroundResponse = await new Promise((resolve, reject) => {
+        const responseListener = (evt: MessageEvent<unknown>) => {
+          if (!isType(evt.data, RpcResponse) || evt.data.id !== id) {
+            return;
+          }
```
jacque006, line 127: Should we `console.warn` on this?

extension/source/QuillEthereumProvider.ts
```diff
@@ -0,0 +1,191 @@
...
+        console.log(
+          `ethereum.request({\n  method: ${JSON.stringify(
+            body.method,
+          )},\n  params: [${body.params
+            .map((p) => JSON.stringify(p))
+            .join(', ')}],\n}) ->`,
+          response,
+        );
+      }
+    }
```
jacque006, line 177: If we start adding in more functionality like this to this file, would be good to move to a middleware pattern.

extension/source/QuillStorageCells.ts
```diff
@@ -0,0 +1,124 @@
...
+      { preferences: rootCells.preferences },
+      ({ preferences: { selectedAddress, identities } }): Theme => {
+        if (selectedAddress === undefined) {
+          return 'light';
+        }
+
+        const identity = identities[selectedAddress];
+        assert(identity !== undefined);
+
+        return identity.theme;
```
jacque006, line 102: What is the use case for having a theme per account?

Since we are trying to be a multichain wallet, we don't really have the idea of a selected account asides from:
- Viewing account details across chains
- If multiple accounts and multiple of them have been connected to a dApp, selecting which one to use for signing/transaction

EDIT: Upon further thought, there are larger changes we need to make outside the scope of this to move the multi chain/account paradigm. Still confused why there is a theme per account instead of one theme for the whole extension.

extension/source/background/KeyringController.ts
```diff
@@ -0,0 +1,111 @@
...
+    },
+
+    eth_requestAccounts: async (_request) => {
+      const selectedAddress = await this.selectedAddress.read();
+      const accounts = selectedAddress ? [selectedAddress] : [];
+      return accounts;
+    },
+
+    eth_coinbase: async (_request) =>
+      (await this.selectedAddress.read()) ?? null,
```
jacque006, line 42: I'm pretty sure this should get the current block’s beneficiary address.

```suggestion
    eth_coinbase: async (_request) => {
      const blockNum = await this.ethersProvider.getBlockNumber();
      const block = await this.ethersProvider.getBlock(blockNum);
      return block.miner;
    }
```

extension/source/background/QuillController.ts
```diff
@@ -0,0 +1,222 @@
...
+
+      eth_coinbase: this.keyringController.rpc.eth_coinbase,
+      eth_accounts: this.keyringController.rpc.eth_accounts,
+      eth_requestAccounts: this.keyringController.rpc.eth_requestAccounts,
+      addHDAccount: this.keyringController.rpc.addHDAccount,
+      setHDPhrase: this.keyringController.rpc.setHDPhrase,
+      lookupPrivateKey: this.keyringController.rpc.lookupPrivateKey,
+      removeAccount: this.keyringController.rpc.removeAccount,
+
+      debugMe: async ({ params: [a, b, c] }) => {
```
jacque006, line 152: Are a, b, and c anything specific here? Or just arbitrary.

extension/source/cells/FormulaCell.ts
```diff
@@ -152,6 +170,44 @@ export class FormulaCell<
...
+    key: K,
+    defaultValue: Exclude<Input[K], undefined | null>,
+    hasChanged: (
+      previous: Input[K] | undefined,
+      latest: Input[K],
+    ) => boolean = jsonHasChanged,
+  ): IReadableCell<Exclude<Input[K], undefined | null>> {
+    return new FormulaCell(
+      { input },
+      // eslint-disable-next-line @typescript-eslint/no-shadow
```
jacque006, line 206: Would be nice to finish https://github.com/web3well/bls-wallet/pull/239 so we don't need to do this anymore.

extension/source/cells/assertType.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * At the time of writing, `as` doesn't seem to work with control flow. If
+ * TypeScript makes something like the code below work in the future, we should
+ * delete `castType`.
+ *
+ * ```ts
+ *     function shout(msg: unknown) {
+ *       msg = msg as string;
+ *       console.log(msg.toUpperCase());
+ * //                ^^^ TypeScript doesn't consider this to be a string 😭
```
jacque006, line 72: I don't think TS will ever update this. `msg`'s type has been defined as `unknown`, so even though you are explicitly assigning it a type (`string`), the var will still be treated as the `unknown` type. You would need a new var of the correct type:

```typescript
*     function shout(msg: unknown) {
 *       const msgString = msg as string;
 *       console.log(msgString.toUpperCase());
```

extension/source/contentScript.ts
```diff
@@ -0,0 +1,139 @@
...
+ * prototype pollution, the content script will not be affected.
+ *
+ * We do make use of this trusted context feature by setting the origin and
+ * providerId here, so that the background script can safely decide which
+ * information is appropriate for its responses using these fields.
+ *
+ * However, the browser could easily provide this kind of information securely
+ * as part of an api that allows direct communication between the in-page script
+ * and the background script. If that ever happens, we should most likely use
+ * that and delete the content script entirely.
```
jacque006, line 21: https://github.com/web3well/bls-wallet/issues/171 -> https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/

Based on understanding of that last link I think we will need to do something like this when we move to v3. There also has been some talk on using an extension registry instead of `window.ethereum` that would communicate either with `iframe`s/`postMessage` or WebRTC, the later of which would just be like connecting to a socket as a message bus.

extension/source/contentScript.ts
```diff
@@ -0,0 +1,139 @@
...
+import assertType from './cells/assertType';
+import isType from './cells/isType';
+import RandomId from './helpers/RandomId';
+import { RpcRequest, RpcResponse, RpcResult, toRpcResult } from './types/Rpc';
+
+(() => {
+  if (
+    document.doctype?.name !== 'html' ||
+    location.pathname.endsWith('.pdf') ||
+    document.documentElement.nodeName.toLowerCase() !== 'html'
```
jacque006, line 47: Probably not an issue, but any chance we could load an `.htm` page?

Also, why is the `.pdf` blocking needed?

Would it be better to do a MIME type check for these, assuming a way to check that exists on `document`, `location`, or somewhere else in a web API.

jacque006 (commented):
Some questions/suggestions we can discuss offline, but on the whole looks like a big improvement on what we had :+1:

Are we going to squash this on merge as well?

extension/source/QuillEthereumProvider.ts
voltrevo: Possibly 👍

extension/source/cells/assertType.ts
voltrevo: discussed, interesting

extension/source/background/KeyringController.ts
voltrevo: Issue created

jacque006 (approved):

PR #253, opened 2022-07-07
Send eth

### LINKED ISSUE #230: Enable sending ETH using button
Screen Shot 2022-06-14 at 5 39 05 pm

voltrevo:
## What is this PR doing?

Enables sending ETH

[Demo video](https://www.youtube.com/watch?v=oqquFOiyysY)

Replicate behavior shown in video.

Resolves #230.

## Checklist

## Guidelines


extension/source/Home/Wallet/Wallets/shortenAddress.ts
```diff
@@ -0,0 +1,3 @@
+export default function shortenAddress(address: string) {
+  return `${address.slice(0, 6)}...${address.slice(38)}`;
+}
```
jacque006: This duplicates [formatCompactAddress](https://github.com/web3well/bls-wallet/blob/main/extension/source/Popup/helpers/formatCompactAddress.ts). I would consolidate them into a util in `extension/source/helpers`.

extension/source/Home/Wallet/Wallets/SendDetail/SendDetail.tsx
```diff
@@ -0,0 +1,141 @@
...
+                  assert(error instanceof Error);
+                  setSendState({ step: 'error', message: error.message });
+                }
+              }}
+            />
+          );
+        }
+
+        return sendState && <SendProgress state={sendState} />;
+      })()}
```
jacque006: Instead of doing this as an inline anonymous component, could be split up into two more components:
 - `SendDetailSelector`, which would just be the early returns of components based on conditions
 - `SendDetailAmountSelector`, which would be returned by `SendDetailSelector` and would contain the `onSend` eth send logic.
You could then pass the top level `cells` down to those children components to keep state in sync.

I had a difficult time initially reading through this component, which was an indication it might be good to break up.

jacque006: Another easier alternative would be to move the inline function for `AmountSelector.onSend` to a component level function so that complexity is not part of the component render branching.

```typescript
const sendEth = async (wei) => {... // use React.useCallback for bonus points

...

<AmountSelector
              selectedAsset={cells.selectedAsset}
              onSend={sendEth}
/>
```

jacque006 (commented):
Although we use the $ symbol to prevent shadowing in certain cell calls, I'm against its proliferation into other variable names such as here and elsewhere in this PR. $ is used commonly for templating and in frameworks like Angular, and is not a common TS/JS variable symbol that I have run across. Adds confusion when reading, especially when `$`, `s`, and `S` are in close proximity.

If there are two vars with similar names, I would just be more specific/verbose with naming one of them. This could also be a good case for enabling a lint rule around variable naming, but don't want to have to disable on every line with $ to prevent shadowing :)

extension/source/Home/Wallet/Wallets/shortenAddress.ts
voltrevo: Oh cool, I was deduplicating this one:
https://github.com/web3well/bls-wallet/pull/253/files#diff-a5c9f5586de6a8376cc8805360a4f556869e1bc1851e3c0064c2fd9d88546e13L22

But using [formatCompactAddress](https://github.com/web3well/bls-wallet/blob/main/extension/source/Popup/helpers/formatCompactAddress.ts) is even better. Will do.

voltrevo:
> Although we use the $ symbol to prevent shadowing in certain cell calls, I'm against its proliferation into other variable names such as here and elsewhere in this PR. $ is used commonly for templating and in frameworks like Angular, and is not a common TS/JS variable symbol that I have run across. Adds confusion when reading, especially when `$`, `s`, and `S` are in close proximity.
> 
> If there are two vars with similar names, I would just be more specific/verbose with naming one of them. This could also be a good case for enabling a lint rule around variable naming, but don't want to have to disable on every line with $ to prevent shadowing :)

Hmm fair enough. I was still figuring out how I felt about it too. Will fix.

extension/source/Home/Wallet/Wallets/SendDetail/SendDetail.tsx
voltrevo: Will fix 👍.

voltrevo:
> Although we use the $ symbol to prevent shadowing in certain cell calls, I'm against its proliferation...

Fixed

extension/source/Home/Wallet/Wallets/shortenAddress.ts
voltrevo: Done

extension/source/Home/Wallet/Wallets/SendDetail/SendDetail.tsx
voltrevo: Fixed

jacque006 (approved):

PR #254, opened 2022-07-10
Confirmation Popup and Transaction Lifecycle

kautukkundan:
## What is this PR doing?

[Demo video](https://youtu.be/Ya7WdBNmK8Q)

1. Added confirmation popup before sending transactions. 
2. Storing all transactions locally (will be used later to show historical txs)
3. Added Transaction lifecycle - new -> approved/rejected -> confirmed/failed/cancelled4. 

does not follow the exact same design as the design doc though

## How can these changes be manually tested?
Replicate behavior shown in video.

## Does this PR resolve or contribute to any issues?
closes #61 
closes #180

## Checklist
- [x] I have manually tested these changes
- [x] Post a link to the PR in the group chat

## Guidelines
- If your PR is not ready, mark it as a draft
- The `resolve conversation` button is for reviewers, not authors
  - (But add a 'done' comment or similar)

PR #261, opened 2022-07-11
Confirmation dialog followup

voltrevo:
## What is this PR doing?

Implements some followup improvements to the [confirmation dialog PR](https://github.com/web3well/bls-wallet/pull/254).

In particular:
- Display the confirm dialog faster
- Don't require the confirm dialog for sending ETH within Quill

See code comments for more changes and details.

Check that Kautuk's functionality all still works: https://youtu.be/Ya7WdBNmK8Q

With one exception:
- Sending ETH within quill no longer uses the confirm dialog

Nope.

## Checklist

## Guidelines

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -56,7 +56,6 @@
     "stream-browserify": "^3.0.0",
     "typed-emitter": "^1.4.0",
     "typescript": "^4.5.3",
-    "uuid": "^8.3.2",
     "webext-base-css": "^1.3.2",
     "webextension-polyfill": "^0.8.0",
     "zxcvbn": "^4.4.2"
@@ -78,7 +77,6 @@
     "@types/react-dom": "^17.0.11",
     "@types/react-router-dom": "^5.3.3",
     "@types/react-table": "^7.7.9",
-    "@types/uuid": "^8.3.4",
     "@types/webextension-polyfill": "^0.8.2",
     "@types/webpack": "^5.28.0",
     "@types/zxcvbn": "^4.4.1",
--- a/extension/source/background/AggregatorController.ts
+++ b/extension/source/background/AggregatorController.ts
@@ -41,6 +41,10 @@ export default class AggregatorController {
     eth_sendTransaction: async (request) => {
       const { providerId, params } = request;
 
+      const userAction = await this.transactionsController
+        .InternalRpc()
+        .requestTransaction(...params);
+
       // FIXME: We should not be assuming that the first from is the same as all
       // the other froms!
       // Supporting multiple froms is actually super awesome - we can show off
@@ -64,10 +68,6 @@ export default class AggregatorController {
         NETWORK_CONFIG.addresses.verificationGateway,
         this.ethersProvider,
       );
-
-      const userAction = await this.transactionsController
-        .InternalRpc()
-        .requestTransaction(...params);
 
       if (userAction === TransactionStatus.APPROVED) {
         // const nonce = (await wallet.Nonce()).toString();
--- a/extension/source/background/AggregatorController.ts
+++ b/extension/source/background/AggregatorController.ts
@@ -39,7 +39,7 @@ export default class AggregatorController {
     eth_sendTransaction: async (request) => {
       const { providerId, params } = request;
 
-      const userAction = await this.InternalRpc().requestTransaction(...params);
+      await this.InternalRpc().requestTransaction(...params);
 
       // FIXME: We should not be assuming that the first from is the same as all
       // the other froms!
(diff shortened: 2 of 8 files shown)


extension/package.json
```diff
@@ -56,7 +56,6 @@
     "stream-browserify": "^3.0.0",
     "typed-emitter": "^1.4.0",
     "typescript": "^4.5.3",
-    "uuid": "^8.3.2",
```
voltrevo, line 59: No need to bring in a new dependency since we already have `RandomId`.

extension/source/types/Rpc.ts
```diff
@@ -31,6 +31,36 @@ export const SendTransactionParams = io.type({
...
+  io.literal('failed'),
+]);
+
+export type TransactionStatus = io.TypeOf<typeof TransactionStatus>;
+
+export const QuillTransaction = io.type({
+  id: io.string,
+  chainId: io.string,
+  from: io.string,
+  status: TransactionStatus,
```
voltrevo, line 49: This was `io.string` before. Using `TransactionStatus` is important to get the most out of TypeScript's checking and make things fit together nicely. This is also a reason why I replaced the enum with a union because enums don't translate as nicely into io types.

extension/source/background/AggregatorController.ts
```diff
@@ -38,8 +36,10 @@ export default class AggregatorController {
   ) {}
 
   rpc = ensureType<PartialRpcImpl>()({
-    eth_sendTransaction: async (request) => {
-      const { providerId, params } = request;
+    eth_sendTransaction: async ({ providerId, params, origin }) => {
+      if (origin !== window.location.origin) {
```
voltrevo: We don't require approval inside the extension.

voltrevo: Also, `BlsWalletWrapper.connect` is slow (about 2 seconds I think (it probably shouldn't, we should figure that out too)), so putting this earlier means the user doesn't wait for those two seconds before seeing the approval dialog which is a much nicer experience.

extension/source/background/AggregatorController.ts
```diff
@@ -65,43 +65,31 @@ export default class AggregatorController {
         this.ethersProvider,
       );
 
-      const userAction = await this.transactionsController
-        .InternalRpc()
-        .requestTransaction(...params);
-
-      if (userAction === TransactionStatus.APPROVED) {
```
voltrevo, line 72: No longer testing whether `userAction` is approved, since `requestTransaction` now throws if the user doesn't approve.

extension/source/background/AggregatorController.ts
```diff
@@ -65,43 +65,31 @@ export default class AggregatorController {
...
+      assert(!('failures' in result), () => new Error(JSON.stringify(result)));
+
+      this.knownTransactions[result.hash] = {
+        ...params[0],
+        nonce,
+        value: params[0].value ?? '0',
+        aggregatorUrl,
+      };
 
-      // TODO - FIX error when tx is rejected and this does not return a hash
```
voltrevo, line 103: Resolved by throwing in `requestTransaction` when the user doesn't approve.

extension/source/background/TransactionsController.ts
```diff
@@ -27,78 +24,76 @@ export default class TransactionsController {
...
-              }
-            });
-
-            function onRemovedListener(windowId: number) {
-              if (windowId === popup.id) {
-                resolve(undefined);
-              }
+          function onRemovedListener(windowId: number) {
+            if (windowId === popup.id) {
+              resolve(TransactionStatus.REJECTED);
```
voltrevo, line 56: Closing the dialog implicitly rejects the transaction. We could add something more granular but `undefined` doesn't seem right.

extension/source/background/TransactionsController.ts
```diff
@@ -27,78 +24,76 @@ export default class TransactionsController {
...
-              if (getPropOrUndefined(message, 'id') !== id) {
-                return;
-              }
+          cleanupTasks.push(() => {
+            windows.onRemoved.removeListener(onRemovedListener);
+          });
 
-              resolve(getPropOrUndefined(message, 'result') as string);
+          function messageListener(message: unknown) {
+            if (!isType(message, PromptMessage) || message.id !== id) {
```
voltrevo, line 67: By checking whether the message is a `PromptMessage` we can then access `.id` and `.result` as we normally would, removing the need for `getPropOrUndefined`.

extension/source/background/TransactionsController.ts
```diff
@@ -27,78 +24,76 @@ export default class TransactionsController {
...
+        })();
+      });
 
       promptAction.finally(() => cleanupTasks.run());
       return promptAction;
     },
 
     createTransaction: async ({ params }) => {
-      const id = uuidv4().toString();
+      const id = RandomId();
```
voltrevo, line 83: No need for uuid since we have `RandomId`. (You could say we're duplicating uuid rather than the other way around, but uuid is unnecessarily verbose with its hex format and dashes so... good riddance imo.)

extension/source/background/TransactionsController.ts
```diff
@@ -115,15 +110,15 @@ export default class TransactionsController {
...
       const id = await this.InternalRpc().createTransaction(...params);
-      const result = await this.InternalRpc().promptUser(id);
+      const userAction = await this.InternalRpc().promptUser(id);
+
+      await this.InternalRpc().updateTransactionStatus(id, userAction);
 
-      if (
-        result === TransactionStatus.APPROVED ||
-        result === TransactionStatus.REJECTED
-      ) {
```
voltrevo, line 123: We now enforce that `promptUser` always responds with one of these, so no need to check it again.

extension/source/background/TransactionsController.ts
```diff
@@ -150,44 +146,44 @@ export default class TransactionsController {
...
+        [TransactionStatus.NEW]: [
+          TransactionStatus.APPROVED,
+          TransactionStatus.REJECTED,
+        ],
+        [TransactionStatus.APPROVED]: [
+          TransactionStatus.CANCELLED,
+          TransactionStatus.APPROVED,
+          TransactionStatus.FAILED,
+        ],
+      };
```
voltrevo, line 166: I find this a lot more readable than the `switch` strategy. Happy to chat about it if you feel differently.

extension/source/background/TransactionsController.ts
```diff
@@ -150,44 +146,44 @@ export default class TransactionsController {
...
+
+      assert(
+        currentValidTransitions.includes(status),
+        () =>
+          new Error(
+            [
+              `Forbidden transaction status transition:`,
+              `${transaction.status} -> ${status}`,
+            ].join(' '),
+          ),
```
voltrevo, line 179: I believe it's important to assert that the transition is valid rather than simply ignoring invalid transitions as before. Is there a reason to prefer ignoring invalid transitions?

extension/source/types/Rpc.ts
```diff
@@ -126,13 +161,14 @@ export const rpcMap = {
   },
 
   // TransactionController
+
   createTransaction: {
-    origin: '*',
+    origin: '<quill>',
```
voltrevo, line 166: I made these all `<quill>` methods since dApps shouldn't be calling them, afaict. E.g. `createTransaction` will insert a transaction into our history even if it isn't acted upon. Malicious dApps could flood our tx history with junk, so I really don't think it should be externally callable.

extension/source/types/Rpc.ts
```diff
@@ -31,6 +31,41 @@ export const SendTransactionParams = io.type({
...
+
+export const TransactionStatusType: io.Type<TransactionStatus> = io.union(
+  Object.values(TransactionStatus).map((v) => io.literal(v)) as ExplicitAny,
+);
+
+export const QuillTransaction = io.type({
+  id: io.string,
+  chainId: io.string,
+  from: io.string,
+  status: TransactionStatusType,
```
voltrevo, line 51: `status` was just `io.string` before.

PR #269, opened 2022-07-15
Local testing fixes

voltrevo:
## Note

aggregator/test is broken on this PR. However, it is also [broken in `main`](https://github.com/web3well/bls-wallet/pull/270), so please disregard it for this PR. [Issue to fix this](https://github.com/web3well/bls-wallet/issues/271).

## What is this PR doing?

Some miscellaneous things I fixed up to get my local environment working. See code comments for details.

Run a local environment.

Nope.

## Checklist

## Guidelines


aggregator/.env.example
```diff
@@ -10,7 +10,7 @@ PRIVATE_KEY_AGG=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff
 PRIVATE_KEY_ADMIN=0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d
 TEST_BLS_WALLETS_SECRET=test-bls-wallets-secret
 
-PG_HOST=localhost
+PG_HOST=127.0.0.1
```
voltrevo, line 13: Docker is a bit weird about `localhost`. I was getting 'connection refused' when using `--net=host` to connect to the postgres container, but for some reason it works with `127.0.0.1`. I don't fully understand it, but since `localhost` is meant to be `127.0.0.1`, I'm happy to just make this replacement and say 'docker made a bad decision that wastes people's time and probably has some stubborn reason for not undoing it'. I could be wrong, I'm curious if anyone knows more.

aggregator/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM denoland/deno:1.20.6
+FROM denoland/deno:1.23.4
```
voltrevo, line 1: This is necessary to make new builds work, due to the dastardly moving targets that are our dependencies.

aggregator/programs/build.ts
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env -S deno run --unstable --allow-run --allow-read --allow-write
+#!/usr/bin/env -S deno run --unstable --allow-run --allow-read --allow-write --allow-env
```
voltrevo, line 1: I'm not sure how this worked before. Possibly due to deno upgrades. Anyway, `build.ts` accesses environment variables, so it needs this flag.

docker-compose.yml
```diff
@@ -28,6 +28,7 @@ services:
     command: >
       --datadir dev-chain/
       --http
+      --http.api eth,web3,personal,net
```
voltrevo, line 31: Not sure why these aren't enabled by default in dev mode. Possibly people are running dev nodes in production and geth is protecting them? 🤯

I got these flags from https://geth.ethereum.org/docs/getting-started/dev-mode. They might not actually be needed for my use-case, as it turns out. But I think they're worth including anyway.

jacque006 (approved):

PR #273, opened 2022-07-19
Add network switching

### LINKED ISSUE #252: Add network switching

voltrevo:
# What is this PR doing?

Mostly the title, but it's a bit involved.

[Demo video](https://www.youtube.com/watch?v=GYnd15N8M1M)

## In particular, this has meant making the keyring aware of different networks

When you generate a wallet you have a stable private key and public key hash (and public key), but the resulting address is per-network. Hopefully we can get stable addresses in the future and simplify this, but I think it's appropriate for now. I also expect there will be some per-network tracking in the keyring because gateway upgrades are a thing and those upgrades are per-network, so you need to keep track of which gateway to use for each wallet (when we implement gateway upgrades).

# Other things that have changed

## Config is now in `config.json` (and `config.example.json`) instead of environment variables
- This allows for more structure in our config and checking that structure with io-ts (environment variables are also 🤮 imo anyway, curious whether this is shared in the team, though it may be an unnecessary can of worms 😉)
- The builtin networks is now part of this config, which kinda relies on this added structure (alternatively, putting json strings into environment variables would be pretty messy (or worse - flattening all these fields onto the top level with the network name in each key (requiring code changes when you wanna add a new network)))
- There's also `blsNetworksConfig.json` which works in a similar way
  - It's separate to allow continuity with our `../contracts/networks/*.json` thing
  - This is generated in `buildBlsNetworksConfig.js` which uses each file basename (which is the network name) as a key and the content as a value
    - I think it would be better to just use this format at the source, but I didn't want to complicate things with cross-project changes right now
  - `CurrencyConversionCell` is now able to get its config from the config file rather than being hardcoded

## Lots of new debug variables in the console of the extension page

Screen Shot 2022-07-19 at 11 34 31 am

## Using `wallet.Nonce()` has been restored

If you're running a local network with the extension, make sure you use contracts that are compatible with the `bls-wallet-clients` module. ([Discord message with more detail](https://discord.com/channels/943612659163602974/963404009266376774/997053021324312626)).

## Improved change detection in cells

`ethers.providers.Web3Provider` does not allow network switching. To solve this, we need a new instance when the network is switched, which led me to make an `IReadableCell<ethers.providers.Provider>`.

Problem is, cells need to know when the underlying value has actually changed, and the default for that, until now, has been the rather primitive `jsonHasChanged` (which won't work for `ethers.providers.Web3Provider` since different instances will stringify the same way).

I've upgraded this default to `mixtureHasChanged`, which actually comes from [a little side project where I'm experimenting and iterating on cells](https://sidekick-bice.vercel.app/), so I was able to make use of this without booking any EF time for it, thanks to the MIT license :).

Open to naming improvements on `mixtureHasChanged`, but the 'mixture' refers to the way the input can be a mixture of plain objects and classes. It works by doing a plain data comparison for all plain data, allows for a custom `hasChanged` on the class if the two values have the same prototype, and ultimately falls back to referential equality.

This also allows for constructions like `ICell<ICell<T>[]>`, which I think is the preferred future of how our transaction history should be modelled (this kind of construction is already being used with success in that side project 😎).

There's also `mixtureCopy` which works similarly and is needed for similar reasons.

See demo video at the top.

Resolves #252.

# Checklist

# Guidelines


extension/buildBlsNetworksConfig.js
```diff
@@ -0,0 +1,27 @@
...
+
+  blsNetworksConfig[filename.slice(0, -'.json'.length)] = JSON.parse(
+    fs.readFileSync(path.join(networksConfigDir, filename), 'utf-8'),
+  );
+}
+
+fs.writeFileSync(
+  path.join(__dirname, 'build', 'blsNetworksConfig.json'),
+  JSON.stringify(blsNetworksConfig, null, 2),
+);
```
jacque006: Create issue to have `../contracts/scripts/deploy_all.ts` do this by default and update all other component usage points to use the new format. We can likely just start with an empty JSON file (`"{}"`) and just overwrite entries when the script is run.

We may also want to have a local version that is git ignored still for local dev flow.

extension/config.release.json
```diff
@@ -0,0 +1,76 @@
...
+      "rpcTarget": "https://mainnet.optimism.io",
+      "chainCurrencyName": "Ether",
+      "chainCurrency": "ETH",
+      "aggregatorUrl": "https://optimism.blswallet.org",
+      "networkKey": "optimism"
+    }
+  },
+  "currencyConversion": {
+    "api": "https://min-api.cryptocompare.com/data/price",
+    "apiKey": "injected-from-github-actions-environment",
```
jacque006:
```suggestion
    "apiKey": "CRYPTO_COMPARE_API_KEY",
```

Then we can inject with something like `envsubst` during release in GH actions. I will look into tweaking that park of the build process.

extension/config.release.json
```diff
@@ -0,0 +1,76 @@
...
+      "blockExplorerUrl": "http://etherscan.io/",
+      "chainId": "0x1",
+      "displayName": "Mainnet",
+      "logo": "",
+      "rpcTarget": "https://mainnet.infura.io/v3/d0f317090d6645b6b494ddc6f1cce5ad",
+      "chainCurrencyName": "Ethereum",
+      "chainCurrency": "ETH",
+      "aggregatorUrl": "https://mainnet.blswallet.org",
+      "networkKey": "mainnet"
+    },
```
jacque006: @jzaki I assume we're never going to have mainnet deploy, so we should just remove this as an option to reduce future confusion?

extension/scripts/addFileStubsIfNeeded.sh
```diff
@@ -0,0 +1,15 @@
...
+PROJECT_DIR="$SCRIPT_DIR/.."
+
+if [ ! -f "$PROJECT_DIR/config.json" ]; then
+  echo {} >"$PROJECT_DIR/config.json"
+fi
+
+if [ ! -f "$PROJECT_DIR/build/blsNetworksConfig.json" ]; then
+  mkdir -p "$PROJECT_DIR/build"
+  echo {} >"$PROJECT_DIR/build/blsNetworksConfig.json"
+fi
```
jacque006: I think this would be better as a NodeJS script like `buildBlsNetworksConfig.js` for consistency with other scripts. Makes it easier for non-Bash/Shell devs to change & maintain as well. Thoughts?

extension/source/BlsNetworksConfig.ts
```diff
@@ -0,0 +1,43 @@
...
+type BlsNetworksConfig = io.TypeOf<typeof BlsNetworksConfig>;
+
+assertType(blsNetworksConfigJson, BlsNetworksConfig);
+
+export function loadBlsNetworksConfig(): BlsNetworksConfig {
+  assertType(blsNetworksConfigJson, BlsNetworksConfig);
+  return blsNetworksConfigJson;
+}
+
+export default BlsNetworksConfig;
```
jacque006: Since we eventually want to use this across all the components in this repo, would it be better to put this in `bls-wallet-clients`? Can keep the current config there as well for backwards compatibility until we move everything over.

extension/source/background/KeyringController.ts
```diff
@@ -101,11 +159,62 @@ export default class KeyringController {
...
+        if (networkDataForWallet === undefined) {
+          networkDataForWallet = {
+            originalGateway: blsNetworkConfig.addresses.verificationGateway,
+            address: await (
+              await this.BlsWalletWrapper(wallet.privateKey)
+            ).address,
+          };
+
+          // eslint-disable-next-line no-param-reassign
+          wallet.networks[network.networkKey] = networkDataForWallet;
```
jacque006: Is there a reason you need to directly assign this here versus using a write/update method on the keyring itself? Seems like there could be some unintended side effects.

extension/source/background/PreferencesController.ts
```diff
@@ -89,14 +116,14 @@ export default class PreferencesController {
     const $preferences = await this.preferences.read();
```
jacque006: #nitpick Not from this changeset, but would prefer a different const name than one that starts with `$`.

extension/source/background/QuillController.ts
```diff
@@ -151,7 +157,21 @@ export default class QuillController {
...
+        const wallet = keyring.wallets.find(
+          (w) => w.networks[network.networkKey]?.address === address,
+        );
+
+        assert(wallet !== undefined);
+
+        this.preferencesController.createUser(
+          wallet.publicKeyHash,
+          'USD',
+          'light',
```
jacque006, line 172: Looks like you can detect this on init, as well as respond to changes if the user switches it in their browser/OS https://stackoverflow.com/a/57795495/648624 #outofscope

extension/source/cells/mixtureHasChanged.ts
```diff
@@ -0,0 +1,77 @@
+export default function mixtureHasChanged<T>(
+  previous: T | undefined,
+  latest: T,
+): boolean {
+  return hasChanged(previous, latest);
```
jacque006, line 5: Could you use [_.isEqual](https://lodash.com/docs/4.17.15#isEqual) here instead on needing custom logic?

jacque006 (commented):
Overall looks good. I think there are some new issues where we can document followup work and some of the larger changes for the new network config format.

I will commit some changes to inject the `CRYPTO_COMPARE_API` in the build process for releases, likely next week.

extension/config.release.json
jzaki: I think it's worth keeping the config there, but concealed from the UI unless explicitly activated by the user (under advanced, with sufficient information)?
For v2 of the contracts will be eip4337 compliant (once the standard settles), and that is intended for any network including mainnet.

jzaki: Regarding `rpcTarget`, we should consider nodes other than infura since that's the default for MM.
That is, consider/evaluate: Alchemy, Moralis, ...

extension/buildBlsNetworksConfig.js
jacque006: https://github.com/web3well/bls-wallet/issues/299

extension/scripts/addFileStubsIfNeeded.sh
voltrevo: Yeah sounds good to me. TypeScript with ts-node might be worth considering too.

extension/source/BlsNetworksConfig.ts
voltrevo: Yeah definitely. I was only keeping it to the extension for now for the simplicity of the change.

extension/source/background/KeyringController.ts
voltrevo: Yeah sorta. In order to do a write/update you'd need to also do a fresh read beforehand as well. Otherwise when updating another wallet the rest of the keyring will have the old values.

It makes a bit more sense to me this way but I don't feel strongly, happy if you'd rather have the update committed for each wallet.

extension/source/background/QuillController.ts
voltrevo: Oooh sounds like a good #good-first-issue.

extension/source/cells/mixtureHasChanged.ts
voltrevo: Hmm that's tricky. I think we'd be fine right now but the problem is that cells care about internal mutability, and `Cell<Cell>` is a valid use-case. `mixtureHasChanged` provides a mechanism for this by using referential equality to distinguish two cells that currently have the same value but are referentially different.

Perhaps that's only a problem for `MemoryCell` 🤔. I have to think about this some more.

I'd prefer to keep it but I'd be ok just using `_.isEqual` for now and revisiting this later if it feels off to you.

extension/source/background/QuillController.ts
voltrevo: Created: https://github.com/web3well/bls-wallet/issues/302.

extension/source/cells/mixtureHasChanged.ts
jacque006: It sounds like there may be some cases that `_.isEqual` or a similar function would not handle as we want, so I'm fine leaving it as is. In general, I like using external lib code for these kinds of things as they are more battle tested, handle edge cases we wouldn't think of, and reduce the code footprint we have to maintain.

Let's leave as is for this PR and consider revisiting in the future when/where it makes sense.

extension/scripts/addFileStubsIfNeeded.sh
jacque006: This will no longer be needed with https://github.com/web3well/bls-wallet/issues/299 , so we can just remove there. Don't expect this will need to be maintained medium/long term.

jacque006:
Remaining tasks:
 - [x] Fix extension build error due to ~bad network name key~ empty `local.json` from CI build.
 - [x] Regression test extension with changes

aggregator/.env.local.example
```diff
@@ -6,8 +6,8 @@ ORIGIN=http://localhost:3000
 PORT=3000
 
 NETWORK_CONFIG_PATH=../contracts/networks/local.json
-PRIVATE_KEY_AGG=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
-PRIVATE_KEY_ADMIN=0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d
+PRIVATE_KEY_AGG=0x6ed6982f65033352b8c6a821a00d462b3774290569fde98331cd828da7ccf6fb
+PRIVATE_KEY_ADMIN=0xafb3afac09efb080cb5f925ba956af68146804e4b892e5060b8c6e96b1ecdf77
```
jacque006: FYI @voltrevo @kautukkundan @JohnGuilding @blakecduncan it looks like the local hardhat network is now pulling accounts from `hardhat.config.ts`, where we use `MAIN_MNEMONIC` from the `.env` file in contracts rather than the pre-loaded ones it used to.

contracts/hardhat.config.ts
```diff
@@ -22,6 +22,28 @@ task("accounts", "Prints the list of accounts", async (_taskArgs, hre) => {
...
+    for (let i = 0; i < accounts.count; i++) {
+      const wallet = hre.ethers.Wallet.fromMnemonic(
+        accounts.mnemonic,
+        `m/44'/60'/0'/0/${i}`,
+      );
+      console.log(`${i}: ${wallet.address}`);
+      console.log(wallet.privateKey);
+      console.log(separator);
+    }
+  });
```
jacque006, line 45: @jzaki This was useful for getting pks for use in local dev, but if you think it might accidentally expose too much I can remove.

jacque006 (approved):

aggregator/.env.local.example
jacque006: Turn out ^ only applies locally, in CI I'm still seeing the old values be needed. Unsure why.

contracts/hardhat.config.ts
jzaki: As long as that's not auto-run :) Good `--force true` requirement :+1:

PR #274, opened 2022-07-19
Update markup and guidelines in PR template

voltrevo:
# What is this PR doing?

## Uses top-level markdown headings

This allows subheadings to be more prominent, including underlines, as demonstrated here (otherwise probably overkill - dot points are still good).

## Added guidelines

`Please be careful when handling disagreements and stay mindful of others`.

I don't think we've run into this in our team, but I've definitely seen it at other companies and feel something like this is very much worth including.

`Prefer to seek followup instead of requiring changes, especially if the PR is already a clear net improvement`.

I think requiring changes unnecessarily is a very common mistake that creates unnecessary delays. This is trying to address that so that we keep the ball rolling rather than perfecting things before they're added.

Of course, sometimes it is appropriate to require changes instead. Things like bugs and change requests that reduce future merge issues come to mind.

`If appropriate, consider implementing followup changes yourself instead of waiting for them from the author and open a new PR against this one`.

When it makes sense to do this, I really feel this is the ideal way to keep things moving. Like all these things, sometimes it makes less sense ofc, eg changes for the benefit of training, changes that you can't (easily) do yourself, and changes that you don't have time to do yourself.

Only by merging this PR 😉.

Nope.

# Checklist
- ~I have manually tested these changes~

# Guidelines
- Sometimes PRs are a source of conflict, please be mindful of your colleagues
- Prefer to seek followup instead of requiring changes, especially if the PR is already a clear net improvement
- If appropriate, consider implementing followup changes yourself instead of waiting for them from the author

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -12,7 +12,7 @@
 - [ ] Post a link to the PR in the group chat
 
 # Guidelines
-- Sometimes PRs are a source of conflict, please be mindful of your colleagues
+- Please be careful when handling disagreements and stay mindful of others
 - If your PR is not ready, mark it as a draft
 - Prefer to seek followup instead of requiring changes, especially if the PR is already a clear net improvement
 - If appropriate, consider implementing followup changes yourself instead of waiting for them from the author and open a new PR against this one.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,17 +1,22 @@
-## What is this PR doing?
+# What is this PR doing?
 
 
-## How can these changes be manually tested?
+# How can these changes be manually tested?
 
 
-## Does this PR resolve or contribute to any issues?
+# Does this PR resolve or contribute to any issues?
 
 
-## Checklist
+# Checklist
 - [ ] I have manually tested these changes
 - [ ] Post a link to the PR in the group chat
 
-## Guidelines
+# Guidelines
+- Sometimes PRs are a source of conflict, please be mindful of your colleagues
 - If your PR is not ready, mark it as a draft
+- Prefer to seek followup instead of requiring changes, especially if the PR is
+  already clear improvement overall
+- If appropriate, consider implementing followup changes yourself instead of
+  waiting for them from the original author
 - The `resolve conversation` button is for reviewers, not authors
   - (But add a 'done' comment or similar)
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -15,6 +15,6 @@
 - Sometimes PRs are a source of conflict, please be mindful of your colleagues
 - If your PR is not ready, mark it as a draft
 - Prefer to seek followup instead of requiring changes, especially if the PR is already a clear net improvement
-- If appropriate, consider implementing followup changes yourself instead of waiting for them from the author
+- If appropriate, consider implementing followup changes yourself instead of waiting for them from the author and open a new PR against this one.
 - The `resolve conversation` button is for reviewers, not authors
   - (But add a 'done' comment or similar)
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -14,9 +14,7 @@
 # Guidelines
 - Sometimes PRs are a source of conflict, please be mindful of your colleagues
 - If your PR is not ready, mark it as a draft
-- Prefer to seek followup instead of requiring changes, especially if the PR is
-  already clear improvement overall
-- If appropriate, consider implementing followup changes yourself instead of
-  waiting for them from the original author
+- Prefer to seek followup instead of requiring changes, especially if the PR is already clear improvement overall
+- If appropriate, consider implementing followup changes yourself instead of waiting for them from the original author
 - The `resolve conversation` button is for reviewers, not authors
   - (But add a 'done' comment or similar)
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -14,7 +14,7 @@
 # Guidelines
 - Sometimes PRs are a source of conflict, please be mindful of your colleagues
 - If your PR is not ready, mark it as a draft
-- Prefer to seek followup instead of requiring changes, especially if the PR is already clear improvement overall
-- If appropriate, consider implementing followup changes yourself instead of waiting for them from the original author
+- Prefer to seek followup instead of requiring changes, especially if the PR is already a clear net improvement
+- If appropriate, consider implementing followup changes yourself instead of waiting for them from the author
 - The `resolve conversation` button is for reviewers, not authors
   - (But add a 'done' comment or similar)


.github/pull_request_template.md
```diff
@@ -1,17 +1,20 @@
...
 
 
-## Checklist
+# Checklist
 - [ ] I have manually tested these changes
 - [ ] Post a link to the PR in the group chat
 
-## Guidelines
+# Guidelines
+- Sometimes PRs are a source of conflict, please be mindful of your colleagues
```
jacque006: Since we would like to accept/be open to external contributions asides from the core team, I would word this differently.

```suggestion
- Please be careful when handling disagreements and stay mindful of others
```

.github/pull_request_template.md
```diff
@@ -1,17 +1,20 @@
...
 - If your PR is not ready, mark it as a draft
+- Prefer to seek followup instead of requiring changes, especially if the PR is already a clear net improvement
```
jacque006, line 17: I personally think it is better to resolve suggestions in place to improve code quality, make sure tech debt is addressed and not lost, and to have more context around changes.

That being said, I do think it is important to delineate between suggestions and requested changes. I can do a better job of this by splitting my comments up into those two sets in future PRs.

.github/pull_request_template.md
```diff
@@ -1,17 +1,20 @@
...
+- If appropriate, consider implementing followup changes yourself instead of waiting for them from the author
```
jacque006: Unless you have direct permission/acknowledgement from the author, I think direct commits into a PR can at best be rude and at worst lead to conflicts/data loss, such as in some of my workflows where I rebase/force push to limit commits.

Instead, perhaps:
```suggestion
- If appropriate, consider implementing followup changes yourself instead of waiting for them from the author and open a new PR against this one.
```

.github/pull_request_template.md
voltrevo: Oh I meant interpersonal conflict. Might be worth clarifying that too.

.github/pull_request_template.md
voltrevo: Agreed, suggestion applied

.github/pull_request_template.md
voltrevo: Yeah it depends. I think it's especially true for us because we're so asynchronous - every round of changes is often another 24h (or much more) because you're not online at the same time.

It also depends on how good we are at doing followup. I always treat followup as a top priority, like your work isn't actually done until the followup is done. I know that other teams are inclined to kick these cans down the road though.

.github/pull_request_template.md
voltrevo: Perhaps this wording:

```
- Please be careful when handling disagreements and stay mindful of others
```

voltrevo:
```suggestion
- Please be careful when handling disagreements and stay mindful of others
```

.github/pull_request_template.md
```diff
@@ -1,17 +1,20 @@
...
+# Checklist
 - [ ] I have manually tested these changes
 - [ ] Post a link to the PR in the group chat
 
-## Guidelines
+# Guidelines
+- Please be careful when handling disagreements and stay mindful of others
 - If your PR is not ready, mark it as a draft
+- Prefer to seek followup instead of requiring changes, especially if the PR is already a clear net improvement
+- If appropriate, consider implementing followup changes yourself instead of waiting for them from the author and open a new PR against this one.
```
kautukkundan, line 18: I disagree on this point, 
doing the followup changes yourself does accelerate the process of merging the PR but the author is left without context about how their own work can be improved next time. The author can indeed go through the changes made afterwards but most of the times they would just let go of things (I am personally guilty of doing this). 

Also it may feel a bit rude as well, giving the impression of mistrust on the author and is the same as pushing more commits to the same PR.

I don't think "rushing" into merging the PR is a good idea. 
This also resonates in this comment
> Prefer to seek followup instead of requiring changes, especially if the PR is already a clear net improvement

I would personally prefer if the PR is "complete in itself" for the feature it is adding rather than having multiple "follow ups" even if it requires a little extra time.

.github/pull_request_template.md
```diff
@@ -1,17 +1,20 @@
...
 - The `resolve conversation` button is for reviewers, not authors
   - (But add a 'done' comment or similar)
```
kautukkundan, line 20: I wanna suggest adding another point -

- Keep the diff of the PR as small as possible, if the PR changes more than around 10 files, further break down the scope of the PR

The PR should only be of a limited scope, if the PR modifies 10+ files it would be better to rethink the scope of the PR and create additional issues and related PRs. (with exceptions ofcourse)
IMHO, it becomes extremely difficult and cumbersome to review a PR which changes 40+ files.

kautukkundan:
I wanna suggest adding another point -

Keep the diff of the PR as small as possible, if the PR changes more than around 10 files, further break down the scope of the PR
The PR should only be of a limited scope, if the PR modifies 10+ files it would be better to rethink the scope of the PR and create additional issues and related PRs. (with exceptions ofcourse)
IMHO, it becomes extremely difficult and cumbersome to review a PR which changes 40+ files.

This would in-turn reduce the turnaround time for a PR review -> merge process.

eg - This PR https://github.com/web3well/bls-wallet/pull/273 can be broken down into atleast 4 different PRs where each PR would be feature complete and easier to review.

kautukkundan:
Also, all the changes under #guidelines should go to a new CONTRIBUTING.md [read more](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors) rather than the PR template itself.

.github/pull_request_template.md
jacque006: Updated suggestion

.github/pull_request_template.md
voltrevo: Hmm. I'm trying to capture this with 'if appropriate', because it's difficult to define when it makes sense.

Having said that, are you suggesting that it's generally not a good idea? Definitely something worth chatting about, especially if you feel there are concrete examples where this has happened.

voltrevo:
> Also, all the changes under #guidelines should go to a new CONTRIBUTING.md [read more](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors) rather than the PR template itself.

I'd like to learn more about why you feel that way.

For me, I don't feel these points would have enough impact in a contributing.md. I think that document is for detail, not a shortlist of important things to remember.

voltrevo:
> I wanna suggest adding another point -
> 
> Keep the diff of the PR as small as possible, if the PR changes more than around 10 files, further break down the scope of the PR The PR should only be of a limited scope, if the PR modifies 10+ files it would be better to rethink the scope of the PR and create additional issues and related PRs. (with exceptions ofcourse) IMHO, it becomes extremely difficult and cumbersome to review a PR which changes 40+ files.
> 
> This would in-turn reduce the turnaround time for a PR review -> merge process.
> 
> eg - This PR #273 can be broken down into atleast 4 different PRs where each PR would be feature complete and easier to review.

I can see what you mean about #273, though I don't think it's necessarily too large and prefer to split it up on request rather than doing it proactively. Splitting up changes can get pretty complicated, and if it's not needed it can be wasteful. Having said that, I'd like to get better at predicting this preference so that I can sometimes do it proactively without running into this problem.

How about something like this?

`Prefer smaller PRs and talk to the author about splitting their PR if you think it would help`

I'm just worried that emphasising small PRs too much can lead to 'sweating the small stuff' (same goes for small issues/tickets, btw). Small PRs can be great when they're appropriate because they are fast to review, but a PR is not a unit of progress. Small PRs can also be very inefficient by accomplishing little but nonetheless going through the full PR process and bringing excess attention to small details. I think a disproportionate amount of useful progress happens in larger PRs, not because they were larger than necessary, but because meaningful change often (but not always) demands it.

kautukkundan:
> I'm just worried that emphasising small PRs too much can lead to 'sweating the small stuff' (same goes for small issues/tickets, btw). Small PRs can be great when they're appropriate because they are fast to review, but a PR is not a unit of progress. Small PRs can also be very inefficient by accomplishing little but nonetheless going through the full PR process and bringing excess attention to small details. I think a disproportionate amount of useful progress happens in larger PRs, not because they were larger than necessary, but because meaningful change often (but not always) demands it.

Hard disagree

voltrevo:
Maybe we can try again some other time.

PR #276, opened 2022-07-19
Specify node version

blakecduncan:
## What is this PR doing?

Specifying what node version we use for the project.  Right now the `./setup.ts` script fails if you run it with node 18 because the extension requires node to be `>=16.0.0 <18.0.0`.

Screenshot 2022-07-19 at 13 19 08

I just added instructions to download node 16 and how to do so if using `nvm`.  If you want to test the error you can clone the project and run the setup script with node 18.

No.

## Checklist

## Guidelines


.gitignore
```diff
@@ -1,2 +1,3 @@
 .data
 .DS_Store
+.idea
```
blakecduncan, line 3: I use webstorm so I was hoping to add this in if that is okay?

jacque006: Yep, :+1:

.nvmrc
```diff
@@ -0,0 +1 @@
+16
```
jacque006, line 1: Should we set this to v18?

jacque006:
@voltrevo I think the only reason we didn't allow NodeJS 18 in the extension was issues with Webpack v4. Now that we are on v5 can we switch the `engine` field to `"node": ">=18.0.0"`?

voltrevo:
@jacque006 I gave this a go but it turns out there are other things holding us back. Still the [md4 shenanigans](https://github.com/web3well/bls-wallet/pull/206#discussion_r886291383), looks like [webpack-extension-reloader](https://www.npmjs.com/package/webpack-extension-reloader) might be the culprit, it's a bit niche and hasn't been updated in a few years.

So yeah regrettably node 16 is the way to go for now.

voltrevo (approved):

.nvmrc
blakecduncan: @jacque006 Looks like Andrew is bumping it up in this [PR](https://github.com/web3well/bls-wallet/pull/280) with his update to webpack

PR #282, opened 2022-07-21
Add currency switching and use currency conversions

### LINKED ISSUE #251: Add currency switching

voltrevo:
# What is this PR doing?

[Demo video](https://www.youtube.com/watch?v=kEsPSdqQXRE)

## Autopicks your currency at install time

Uses the [IP API](https://ip-api.com/) instead of just USD.

## Currency switching

- 165 currencies to choose from
- Currencies are ordered by relevance to the current currency (as determined by international trade) by default
- Alphabetical ordering also available

## Applies currency conversions

Screen Shot 2022-07-21 at 3 14 30 pm
Screen Shot 2022-07-21 at 3 15 22 pm
Screen Shot 2022-07-21 at 3 19 01 pm
Screen Shot 2022-07-21 at 3 16 13 pm

These are all updated live too 😎.

## Confirm popup has a nicer table now

When adding the currency conversion in this table, I felt this was also a nice upgrade.

Screen Shot 2022-07-21 at 3 36 52 pm

## `preferences.identities` has been removed

This was discussed in the team on a call previously.

It's something we inherited from the copied code. The idea is that torus/metamask has a multiple identities feature where each wallet/address gets its own preferences.

We're not doing that, so those preferences are just at the top level now.

## `CurrencyConversionCell` now exposes the currencies it's currently converting

This makes it easier to keep things consistent - when it's just a number for the current rate and you switch currencies, that number will be wrong until we get a new response from cryptocompare.com.

Replicate the things in the demo video.

Resolves #251.

# Checklist

# Guidelines


extension/source/Home/Wallet/Settings/GeneralSettings.tsx
```diff
@@ -1,15 +1,61 @@
...
+
 const GeneralSettings: FunctionComponent = () => {
   const { cells } = useQuill();
 
+  const orderPreference = useMemo(
+    () => new MemoryCell<'trade' | 'alphabet'>('trade'),
+    [],
+  );
+
+  const $orderPreference = useCell(orderPreference) ?? 'trade';
```
jacque006: Boo `$`

```suggestion
  const orderPreferenceVal = useCell(orderPreference) ?? 'trade';
```

extension/source/currencies/defaultCurrency.ts
```diff
@@ -0,0 +1,21 @@
+import * as io from 'io-ts';
+
+import assertType from '../cells/assertType';
+
+export default async function defaultCurrency(): Promise<string> {
+  try {
+    const ipBasedCurrencyJson = await fetch(
+      'http://ip-api.com/json/?fields=currency',
+    ).then((res) => res.json());
```
jacque006: I don't think some people are going to like it if we're letting a 3rd party service look them up by IP location, even though this could happen if we decided to log ips on a aggregator or the same on a chain RPC node.

I would recommend doing one of the following instead to bootstrap someone's initial currency:
- Use [navigator.language](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language) and map that to a currency.
- Ask for permission to use `http://ip-api.com/json/?fields=currency'` to detect someone's currency
- Use [navigator.geolocation.getCurrentPosition](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/getCurrentPosition) to allow the user to decide if they want to share their position. Then translate those coordinates to a country. Note if you do this you will also need to update our extension manifest with the ['geolocation' permission](https://developer.chrome.com/docs/extensions/mv3/declare_permissions/)

jacque006 (changes requested):
Think we should do something else to get someones default currency. Otherwise looks good.

extension/source/currencies/defaultCurrency.ts
voltrevo: > letting a 3rd party service look them up by IP location

I think there might be a misunderstanding here. Every website and external service you use gets to see your IP address and can use that IP to look up your location. Services don't ask permission for your IP because it's a pre-requisite of using the internet.

Users wishing to hide their local IP need to use a VPN or similar. I don't really see the benefit of avoiding ip-api.com.

Perhaps we should chat about this further?

extension/source/Home/Wallet/Settings/GeneralSettings.tsx
voltrevo: Fixed

extension/source/currencies/defaultCurrency.ts
voltrevo: Updated to just use USD as discussed.

jacque006 (approved):

Issue #277, opened 2022-07-19
Store wallet unlock credentials


kautukkundan:
![Screenshot 2022-07-19 at 10 26 02 PM](https://user-images.githubusercontent.com/23727056/179806974-572d2834-e467-4965-afdf-6c4fc8d3126d.png)
Step 1 of the onboarding process requires user to create a password to unlock their wallet. However, it is just a functional placeholder and the password isn't stored anywhere for further use. 
The relevant code can be found in [PasswordCreationPanel.tsx](https://github.com/web3well/bls-wallet/blob/main/extension/source/Home/Onboarding/PasswordCreationPanel.tsx) 

The credentials can be stored in the KeyringController itself. 

for reference checkout [Matamask's controllers ](https://github.com/MetaMask/controllers)

PR #283, opened 2022-07-21
Adding an encrypted vault to local storage 

blakecduncan:
## What is this PR doing?

Adding a vault to local storage.  The vault is an encrypted string (encrypted with the users pw) that contains the `HDPhrase`, `HDPath`, and the `numberOfAccounts`.  My understanding from looking into stuff the past couple days is if you have the three of those you can always recreate the Private and Public keys.

This PR is in draft because I want to confirm that this is the direction we want to go with this.  Some next steps I could think of if we go this way.

1.  Update the onboarding flow so we don't have to temporarily store the password in local storage.
2. Remove all private keys from local storage.  They should not be persisted and should only be in memory or the user must re-enter their password.
3. Update existing logic to keep the vault up to date with edits to a users wallet (like adding a new address).

Yes.  There should be no visible changes.  After creating the account you should be able to see the vault in local storage.

[277](https://github.com/web3well/bls-wallet/issues/277)

## Checklist

## Guidelines

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -37,6 +37,7 @@
     "async-mutex": "^0.3.2",
     "axios": "^0.27.2",
     "bls-wallet-clients": "0.6.0",
+    "browser-passworder": "^2.0.3",
     "bs58check": "^2.1.2",
     "crypto-browserify": "^3.12.0",
     "emoji-log": "^1.0.2",
--- a/extension/source/Home/Onboarding/ReviewSecretPhrasePanel.tsx
+++ b/extension/source/Home/Onboarding/ReviewSecretPhrasePanel.tsx
@@ -94,10 +94,16 @@ const ReviewSecretPhrasePanel: FunctionComponent<{
   const navigate = useNavigate();
 
   const setHDWalletPhrase = async () => {
+    // Temporary solution to get the password after the phrase
+    // is created. We could create the HDPhrase at the point the
+    // password is created to avoid this.
+    const password = (await cells.onboarding.read()).tempPassword;
+
     await rpc.setHDPhrase(secretPhrase.join(' '));
     const address = await rpc.addHDAccount();
     await rpc.setSelectedAddress(address);
-    await cells.onboarding.update({ completed: true });
+    await rpc.createNewVault(password);
+    await cells.onboarding.update({ completed: true, tempPassword: '' });
 
     navigate('/wallets');
   };
--- a/extension/source/background/KeyringController.ts
+++ b/extension/source/background/KeyringController.ts
@@ -1,5 +1,7 @@
 import { BlsWalletWrapper } from 'bls-wallet-clients';
 import { ethers } from 'ethers';
+import encryptor from 'browser-passworder';
+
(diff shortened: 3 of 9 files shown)


extension/package.json
```diff
@@ -37,6 +37,7 @@
     "async-mutex": "^0.3.2",
     "axios": "^0.27.2",
     "bls-wallet-clients": "0.6.0",
+    "browser-passworder": "^2.0.3",
```
blakecduncan, line 40: This is the package metamask uses.  I can look into if there are others if we want to follow this approach

extension/source/Home/Onboarding/ReviewSecretPhrasePanel.tsx
```diff
@@ -94,10 +94,16 @@ const ReviewSecretPhrasePanel: FunctionComponent<{
...
+    // is created. We could create the HDPhrase at the point the
+    // password is created to avoid this.
+    const password = (await cells.onboarding.read()).tempPassword;
+
     await rpc.setHDPhrase(secretPhrase.join(' '));
     const address = await rpc.addHDAccount();
     await rpc.setSelectedAddress(address);
-    await cells.onboarding.update({ completed: true });
+    await rpc.createNewVault(password);
+    await cells.onboarding.update({ completed: true, tempPassword: '' });
```
blakecduncan, line 106: Part of my temporary solution - I'm clearing the password from local storage after the vault is created.

extension/source/background/KeyringController.ts
```diff
@@ -99,6 +101,21 @@ export default class KeyringController {
 
       await this.keyring.update({ wallets: newWallets });
     },
+
+    createNewVault: async ({ params: [password] }) => {
```
blakecduncan, line 105: We will need additional functions to decrypt the vault, edit the vault, etc.

kautukkundan:
> My understanding from looking into stuff the past couple days is if you have the three of those you can always recreate the Private and Public keys.

That is correct!
There are some limitations to the private keys generated by the HD wallet method to be compatible with BLS keys but its not relevant right now, just putting it here as an fyi.

Overall It looks perfectly fine and I am onboard with the next steps you mentioned 👍🏼

extension/package.json
kautukkundan: this works

blakecduncan:
> That is correct! There are some limitations to the private keys generated by the HD wallet method to be compatible with BLS keys but its not relevant right now, just putting it here as an fyi.
> 
> Overall It looks perfectly fine and I am onboard with the next steps you mentioned 👍🏼

Thanks for the review @kautukkundan.  Just to give you a heads up, I paired up with @voltrevo this morning and we discussed an approach of encrypting everything rather than just the three values I did.  I think that approach is more consistent with how this extension is currently set up and will make it easier in the long run.  I'm going to close this PR and take a stab at that

blakecduncan:
Closing this pull request as I will be putting up a new PR soon with a different strategy for encrypting data

blakecduncan:
Closing this pull request as I will be putting up a new PR soon with a different strategy for encrypting data

PR #286, opened 2022-07-26
Encrypting necessary local storage data

blakecduncan:
## What is this PR doing?
Adding a local cell collection that encrypts the `values` before persisting to local storage

Yes this was tested.  There should be no functional UX changes to the app.

[277](https://github.com/web3well/bls-wallet/issues/277)

## Checklist

## Guidelines


extension/source/cells/encryptedLocalCellCollection.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const events = new EventEmitter() as IAsyncStorage['events'];
+
+Browser.storage.onChanged.addListener((changes) => {
+  events.emit('change', Object.keys(changes));
+});
+
+// TODO: Store users key that is generated from the password in memory so we
+// don't keep a copy of the password
+const PASSWORD = 'TEMP_PASSWORD';
```
blakecduncan, line 19: I thought for a first PR it would be good to just get the encryption working.  I figure as a separate PR I'll store the pass key in memory.

If anyone has any recommendations on how I should do that, let me know.  I know that metamask uses [obs-store](https://www.npmjs.com/package/obs-store) to keep the password in memory.

extension/source/QuillStorageCells.ts
```diff
@@ -32,7 +35,7 @@ function QuillStorageCells(storage: CellCollection) {
         completed: false,
       }),
     ),
-    keyring: storage.Cell(
+    keyring: encryptedStorage.Cell(
```
blakecduncan, line 38: This was the cleanest way I could think of to allow both encrypted and decrypted storage.  If anyone has alternative ideas let me know.

This approach allows us to only encrypt specific cells.  So for example, we can keep the preferences decrypted

extension/source/cells/encryptedLocalCellCollection.ts
```diff
@@ -0,0 +1,62 @@
...
+  async read<T>(key: string, type: io.Type<T>): Promise<T | undefined> {
+    const readResult = (await Browser.storage.local.get(key))[key];
+
+    if (readResult === undefined) {
+      return undefined;
+    }
+
+    const payload = JSON.parse(readResult);
+    const { salt } = payload;
+    const passwordKey = await encryptor.keyFromPassword(PASSWORD, salt);
```
blakecduncan, line 31: This key from password part will be moved somewhere else once we are keeping the key in memory

extension/source/QuillStorageCells.ts
kautukkundan: This is perfect! I was wishing for exactly the same setup. Just change the storage type and your collection is good to go.

extension/source/QuillStorageCells.ts
```diff
@@ -19,7 +19,10 @@ import { QuillTransaction } from './types/Rpc';
 // fields that always have concrete values incrementally without breaking
 // existing clients.
 
-function QuillStorageCells(storage: CellCollection) {
+function QuillStorageCells(
+  storage: CellCollection,
```
kautukkundan: lets make this change to be more explicit about the storage type
```diff
- storage
+ standardStorage
```
A few more places within the file

extension/source/background/QuillController.ts
```diff
@@ -53,9 +53,10 @@ export default class QuillController {
 
   constructor(
     public storage: CellCollection,
+    public encryptedStorage: CellCollection,
     public currencyConversionConfig: CurrencyConversionConfig,
   ) {
-    this.cells = QuillStorageCells(storage);
+    this.cells = QuillStorageCells(storage, encryptedStorage);
```
kautukkundan: considering storage and encryptedStorage are types of the same entity, we can group these and pass as an object instead of 2 different constructor arguments 

lets create a new type 
```ts
export type StorageConfig = {
  standardStorage: CellCollection;
  encryptedStorage: CellCollection;
};
```

and modify the constructor to accept a single storage object
```ts
public storage: StorageConfig,
```
and access standardStorage and encryptedStorage separately

```diff
- this.cells = QuillStorageCells(storage, encryptedStorage);
+ this.cells = QuillStorageCells(storage.standardStorage, storage.encryptedStorage);
```

extension/source/QuillStorageCells.ts
```diff
@@ -19,7 +19,10 @@ import { QuillTransaction } from './types/Rpc';
 // fields that always have concrete values incrementally without breaking
 // existing clients.
 
-function QuillStorageCells(storage: CellCollection) {
+function QuillStorageCells(
```
kautukkundan, line 22: the separate arguments are fine here since this function only handles storage 👍🏼

extension/source/QuillContext.tsx
```diff
@@ -66,9 +71,10 @@ export function QuillContextProvider({ children }: Props) {
 
 function QuillContextCells(
   storage: CellCollection,
+  encryptedStorage: CellCollection,
```
kautukkundan: Same changes as QuillController's constructor can be made here by importing the `StorageConfig` type since both arguments are part of same entity.

extension/source/background/index.ts
```diff
@@ -15,6 +16,7 @@ import { assertConfig } from '../helpers/assert';
 
   const quillController = new QuillController(
     extensionLocalCellCollection,
+    encryptedLocalCellCollection,
```
kautukkundan: this would be converted to a single object as per changes mentioned above.

extension/source/cells/encryptedLocalCellCollection.ts
kautukkundan: works for now 👍🏼

kautukkundan (changes requested):
The changes are good! Exactly what I was hoping for. Left some minor comments related to grouping of similar items.

kautukkundan (approved):
LGTM 👍🏼

PR #307, opened 2022-08-18
Update private key

JohnGuilding:
## What is this PR doing?
This PR updates the aggregator private key: `PRIVATE_KEY_AGG` in `./aggregator/.env.example`. Previously, this private key belonged to the first account that the hardhat node provides. This has been updated to the private key from the 3rd account in the account list that hardhat provides.

These changes have been tested by running bls wallet locally, attempting to send transactions, and observing there is no nonce too low errors.

A periodic nonce mismatch issue occurs when the aggregator tries to submit a batch. This is because the private key used by the aggregator is the same as the default contract deployer. This change swaps this aggregator private key for another one so there is no longer a clash.

Error:
image

## Checklist

## Guidelines


JohnGuilding:
Tests failing, looking into this now

blakecduncan (approved):
lgtm!  Glad it was a simple fix 💯

PR #308, opened 2022-08-19
New testnet deployments - Arb Rinkeby and Arb Goerli

kautukkundan:
## What is this PR doing?
deployed a new set of contracts on arbitrum testnet

tested the deployment by manually running the aggregator and extension and sending ETH from wallet

in extension/.env
add this line
```
NETWORK_CONFIG='../contracts/networks/arbitrum-testnet.json'
```
or
```
NETWORK_CONFIG='../contracts/networks/arbitrum-testnet-goerli.json'
```
- remove the installed extension from browser
- run `yarn run dev:chrome`  
- re-import extension

in aggregator/.env
make these changes
```
RPC_URL=https://rinkeby.arbitrum.io/rpc or https://goerli-rollup.arbitrum.io/rpc
PRIVATE_KEY_AGG=<any_pvt_key_with_rinkarbyETH>
PRIVATE_KEY_ADMIN=<any_pvt_key_with_rinkarbyETH>
```

clear any existing docker container and image of postgres instance
- run `docker-compose up -d postgres`
- run `./programs/aggregator.ts`

resolves the issue of aggregator not being able to process txs from original deployment. 

## Checklist

## Guidelines


blakecduncan (approved):
Tested this out with my burner wallet and it worked for me 💯

kautukkundan:
part of #310

voltrevo (changes requested):
I don't think this is quite right.

The current arbitrum rinkeby contract should already have this version, what's the reason for the change? (perhaps to use the `create2Deployer` as described below?)

This uses the `create2Deployer` at `0x036d...`, which is derived from the publicly exposed mnemonic in `.env.example`. While these are testnets, I believe the idea is to get consistent addresses across networks, which requires keeping the account that creates `create2Deployer` secret. Otherwise a griefer can make a different transaction with that account before us, which permanently removes our ability to deploy to the intended address.

I believe that the account behind `0xc132...` is already set up for this purpose, but you'll need @jzaki (or maybe @jacque006?) to make the `create2Deployer` on arbitrum goerli. New issue to track this: https://github.com/web3well/bls-wallet/issues/331.

jzaki:
@voltrevo lets use -> https://github.com/gnosis/safe-singleton-factory.
Will need to create issues requesting Arbitrum and Optimism testnets if the network id is not yet in `./artifacts/`.
EDIT: https://github.com/safe-global/safe-singleton-factory/issues/64

voltrevo:
@jzaki Nice! I was wondering whether something like that might exist. Actually I really hope this gets built in at the protocol level so no one needs to rely on this kind of thing.

voltrevo:
In the interest of time, we're just going to use that hardcoded phrase for this version. More contract updates are definitely coming anyway 🤷‍♂️. CC @jzaki.

voltrevo (approved):

PR #339, opened 2022-09-14
Added wrapper to return bls signer without connecting to VG

### LINKED ISSUE #337: Generate BLS keypair without creating deployable wallet

kautukkundan:
## What is this PR doing?
adds method `generateStaticBlsSigner` on BlsWalletWrapper class to generate a signer without connecting to Verification Gateway

closes #337 

## Checklist

## Guidelines

### DIFF
--- a/contracts/clients/src/BlsWalletWrapper.ts
+++ b/contracts/clients/src/BlsWalletWrapper.ts
@@ -26,8 +26,8 @@ export default class BlsWalletWrapper {
   private constructor(
     public blsWalletSigner: BlsWalletSigner,
     public privateKey: string,
-    public address: string,
-    public walletContract: BLSWallet,
+    public address?: string,
+    public walletContract?: BLSWallet,
   ) {}
 
   /** Get the wallet contract address for the given key, if it exists. */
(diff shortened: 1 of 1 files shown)


voltrevo:
I think it would make more sense to simply use `BlsWalletSigner` directly for this.

ie, instead of:

```ts
const wallet = await BlsWalletWrapper.generateStaticBlsSigner(privateKey, provider);
const publicKey = wallet.getPublicKey();
```

you could just write:

```ts
const network = await provider.getNetwork();
const signer = await initBlsWalletSigner({ chainId: network.chainId });
const publicKey = signer.getPublicKey(privateKey);
```

It's a line longer, but even if we went with `generateStaticBlsSigner` I'd also want it to take `chainId` instead of `provider`, since the only reason it's taking a provider is to get the `chainId` inside. Indeed, this is the reason why `initBlsWalletProvider` takes a `chainId` and not a `provider`. So, it's actually 3 lines either way.

Does that make sense to you? Is there another reason to add this that I might be missing?

voltrevo (changes requested):
See comment above

kautukkundan:
>  Is there another reason to add this that I might be missing?

Yeah kinda, I thought it maybe worthwhile to keep the functionality of connecting to the VG afterwards in case the dev wants to but it feels like instantiating another object with the same details is not too difficult anyway. 
Alright, I'll go over it again, might as well close this PR as the suggested method is already implemented.

voltrevo:
Closing as per suggestion (feel free to re-open if I'm mistaken)

Issue #140, opened 2022-02-15
Automate gas cost comparison script to run in GH actions


jacque006:
Depends on https://github.com/jzaki/bls-wallet/pull/138

After https://github.com/jzaki/bls-wallet/pull/138 is merged, add a Github Actions pipeline for automatically running and uploading the gas cost comparisons on contract changes so people can see the gas differences when using BLS wallets.

### Considerations
- Probably need to create a markdown template to inject measurement results into.
- Run tests on any commit where `./contracts` is changed.
- Use Arbitrum Testnet to run tests by default, but see if it is possible to setup and run a local node instead as it will reduce the amount of time to run the script and not require constant funding of the address on testnet.
- Upload the report to somewhere and link on the main `README.md` file to the latest run for `main` branch. This could be uploaded to somewhere on Github, IFPS, AWS S3, etc.

PR #341, opened 2022-09-17
Update gas measurement script for Arbitrum Nitro

### LINKED ISSUE #311: Update and run gas measurement script
Script likely needs updating, and when run against #310, should be able to do more than 29 txs.
Array [here](https://github.com/web3well/bls-wallet/blob/main/contracts/scripts/test/measure_gas.ts#L19) can be the number of UserOperations to try in a bundle, eg: `[20, 40 , 80, 120]`, it will fail once the bundle is too large for the chain's transaction.

Previuosly on Arbitrum testnet there was a limit to the number of curve operations per tx, this limit is not present in Nitro.

jacque006:
## What is this PR doing?

Update gas measurement script for Arbitrum Nitro

- Add alea lib for deterministic random values
- Move gas scripts to new folder
- Move to config based approach for measurement script
- Fix bls wallet lazy creation in fixture so nonce will be correct for larger amount of wallets
- Fix fixture initialization of bls wallet contract from swallowing errors & attempting to initialize when not needed.
- Get arbitrum specific gas field
- Add address indexing optimization to expander, better typing on contracts fixture
- Use @ethereumjs/tx to get raw txn size
- Add local CI test to make sure measurement script is working
- Update measurements to be able to handle multiple transactions
- Update `gethDev` to allow for longer http timeouts on longer running operations
- I somewhat duplicated a number of `Fixture/TokenHelper` features as I needed more fine grained control and was running into unexpected side effects. We can likely consolidate these in the future.

### local
Run `yarn hardhat run scripts/measure_gas/run.ts --network gethDev`. An automated test has been added to run 

### arbitrum goerli
Run `yarn hardhat run scripts/measure_gas/run.ts --network arbitrum_goerli`. Make sure there are funds in Arbitrum Goerli at account index 0 for `env.MAIN_MNEMONIC`.

Resolves https://github.com/web3well/bls-wallet/issues/311

## Checklist

## Guidelines

## TODO

- [x] Add measurement task for non-`BLSExpander` transfers
- [x] Add Arb calldata costs (`gasUsedForL1`) https://medium.com/offchainlabs/understanding-arbitrum-2-dimensional-fees-fd1d582596c9
~Include gas price for L1 & L2, total & per transaction ETH costs~ Moved to future work
~(Strech) automate running & publishing of results on comits/merges to `main`.~ https://github.com/web3well/bls-wallet/issues/140
~Add clients README entry for `connectToContracts`~
~Add integration tests for `connectToContracts`~ Moved to https://github.com/web3well/bls-wallet/issues/503

# Follow Up

https://github.com/web3well/bls-wallet/milestone/9


contracts/scripts/test/measure_gas.ts
```diff
@@ -16,7 +16,7 @@ async function main() {
 }
 
 async function logGasForTransfers() {
-  const transferCounts = [31];
+  const transferCounts = [241, 242, 243, 244, 245, 246, 247, 248, 249, 250];
```
jacque006: @jacque006 Update this to more staggered numbers. Maybe add failure row if txn fails/is rejected.

jacque006:
@jacque006 Use `address` instead of pubkey hash for wallet lookup in indexing test/measurement

contracts/contracts/BLSExpander.sol
```diff
@@ -145,4 +154,42 @@ contract BLSExpander {
...
+        }
+
+        // Use them to re-create bundle
+        VerificationGateway.Bundle memory bundle;
+        bundle.signature = addressBundle.signature;
+        bundle.senderPublicKeys = senderPublicKeys;
+        bundle.operations = addressBundle.operations;
+
+        return verificationGateway.processBundle(bundle);
+    }
```
jacque006, line 194: @jzaki I assume these changes are ok to go in since the expander is not a core contract. Imagine we will carry these optimizations further in the future.

jacque006: Confirmed.

jacque006:
Moving back to draft, making additional changes to optimize.

voltrevo (changes requested):
Ran into a few problems manually testing:

1. Testing locally with geth (`yarn start`), the script crashes with `execution reverted` on an early `estimateGas` call:

```
<ref *1> Error: cannot estimate gas; transaction may fail or may require manual gas limit [ See: https://links.ethers.org/v5-errors-UNPREDICTABLE_GAS_LIMIT ] (error={"reason":"execution reverted","code":"UNPREDICTABLE_GAS_LIMIT","method":"estimateGas","transaction":{"from":"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266","maxPriorityFeePerGas":{"type":"BigNumber","hex":"0x59682f00"},"maxFeePerGas":{"type":"BigNumber","hex":"0x61db5578"},"to":"0x689A095B4507Bfa302eef8551F90fB322B3451c6","data":"0xc65cd28900000000000000000000000000000000000000000000000000000000000000201cb60f8e17fdb4a185143dcda69af9b107358c59c98a39d0e4298ec405aa40552706e3373b7fd40fa227f91fb8aa45b6b646ea
```

Double checked this on a fresh environment using EC2.

Seems to work with `yarn start-hardhat`. Not sure what's going on there.

2. When running against arbitrum goerli, I spent some time troubleshooting before realizing there's a hardcoded `networkConfigName: "local",` in `run.ts` which needs to be changed to `"arbitrum-goerli"`. We should be able to detect the `--network` setting instead (`process.argv` if necessary). Alternatively, there at least needs to be a clear instruction about this.

3. After fixing (2) above, the first 3 configs work but the 4th fails with:

```
running transfer, mode blsExpanderAddress. # of transactions: 10
registering public keys...
<ref *1> Error: cannot estimate gas; transaction may fail or may require manual gas limit [ See: https://links.ethers.org/v5-errors-UNPREDICTABLE_GAS_LIMIT ] (error={"reason":"execution reverted","code":"UNPREDICTABLE_GAS_LIMIT","method":"estimateGas","transaction":{"from":"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266","maxPriorityFeePerGas":{"type":"BigNumber","hex":"0x59682f00"},"maxFeePerGas":{"type":"BigNumber","hex":"0x6553f100"},"to":"0x4473e39a5F33A83B81387bb5F816354F04E724a3","data":"0x194efdd1000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000001a0000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000008a3925f17d23d3415d898f8236dd017ccf16b98800000000000000000000000015964751601f28ed...
```

voltrevo:
Sadly, I'm getting `"gasUsedForL1": 0` for arbitrum-goerli, which means that it isn't possible to measure L1 gas at the moment.

I believe this is because the gas price on goerli's L1 is almost zero at the moment.

voltrevo:
> Include gas price for L1 & L2, total & per transaction ETH costs

I don't see the L1 gas price or any per-transaction measurements. Outputs I'm getting look like this:

```
        {
            "transactions": {
                "count": 10,
                "mode": "bls",
                "type": "transfer",
                "hashes": [
                    "0xae921946d6f303ee0c30b83a692930fe6914514c792739051e5f08bb0aa3d617"
                ],
                "totalSizeBytes": 5466
            },
            "gas": {
                "used": 1593641,
                "price": 100000000,
                "arbitrum": {
                    "gasUsedForL1": 0
                }
            }
        },
```

voltrevo:
> "type": "transfer",

To me, a 'transfer' without any qualification should mean ETH transfer, but these are ERC20 transfers. Can we rename this to `erc20Transfer`?

voltrevo:
> Calculate new calldata costs

Where is this in the output?

voltrevo:
Let's omit `"arbitrum": { "gasUsedForL1": 0 }` when not testing on arbitrum.

.github/workflows/contracts.yml
```diff
@@ -32,3 +32,13 @@ jobs:
...
+
+  # ensure gas measurement script runs
+  test-gas-measurements:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - uses: ./.github/actions/setup-contracts-clients
+    - uses: ./.github/actions/local-contract-deploy
+    - run: yarn hardhat run ./scripts/measure_gas/run.ts --network gethDev
```
voltrevo, line 44: The results of this get written to a file which we can't see after CI runs. We should upload it as an artifact or add a `--stdout` option so that we can see the measurement results on CI.

Could be done as followup.

contracts/scripts/measure_gas/token.ts
```diff
@@ -0,0 +1,68 @@
...
+export const createTransferAction = (
+  ctx: GasMeasurementContext,
+  to: string,
+): ActionData => ({
+  contractAddress: ctx.contracts.testToken.address,
+  encodedFunction: ctx.contracts.testToken.interface.encodeFunctionData(
+    "transfer",
+    [to, getTransferAmount(ctx)],
+  ),
+  ethValue: 0,
```
voltrevo, line 24: This does not include a `tx.origin` payment, which is essential.

Makes sense to split it out though: https://github.com/web3well/bls-wallet/issues/508.

jacque006: Since these measurements go directly through the VG contract, I don't think it's relevant at this level as the VG doesn't care about `tx.origin`. If we decided in the future to do a more full stack measurement using an agg instance, we can add that payment in that script.

.github/workflows/contracts.yml
jacque006: This is mainly here as a smoke test to let someone know (such as myself when developing this) if their changes break this script. See https://github.com/web3well/bls-wallet/issues/140 for work which will upload the results.

jacque006:
> Testing locally with geth (yarn start), the script crashes with execution reverted on an early estimateGas call

I'm seeing this as well, not sure what the cause is. If you manually specify a gas limit it will still fail for unknown reasons. Broke off into https://github.com/web3well/bls-wallet/issues/515. Don't think it will be high priority as most people will run this against a remote network or hardhat node to debug.

>     When running against arbitrum goerli, I spent some time troubleshooting before realizing there's a hardcoded networkConfigName: "local", in run.ts which needs to be changed to "arbitrum-goerli". We should be able to detect the --network setting instead (process.argv if necessary). Alternatively, there at least needs to be a clear instruction about this.

Configuration is mentioned in the [README](https://github.com/web3well/bls-wallet/pull/341/files#diff-ac4b5a05a6426e2ff4c0b120dbd12660d2367803a1775a026e2282718f4c8730R9-R11), but it makes sense to make this path default to the network it is being run on. Fixed.

> 
> After fixing (2) above, the first 3 configs work but the 4th fails with:
> 
> running transfer, mode blsExpanderAddress. # of transactions: 10
> registering public keys...
> <ref *1> Error: cannot estimate gas; transaction may fail or may require manual gas limit [ See: https://links.e

I was unable to reproduce this, was able to run on local hardhat node fine. May need your configuration to reproduce. I've tended to see this 4th step fail if the `BLSExpander` contract is not initialized with the correct VG contract, usually when I have multiple contract deploys remotely.

> I don't see the L1 gas price or any per-transaction measurements

L1 stats were broken off into https://github.com/web3well/bls-wallet/issues/502 . The per-transaction measurements can be calculated by just taking the total cost(s) and dividing by the batch size. I tried to keep the information in the JSON output limited but enough that a person or machine could write a report based on just that info, esp the txn hashes.

> To me, a 'transfer' without any qualification should mean ETH transfer, but these are ERC20 transfers. Can we rename this to erc20Transfer?

If we add in ETH transfers, I would be fine with updating that name. My gut feeling is that `IERC20.transfer`'s are much more frequent than ETH transfers (such as with stablecoins) and therefore are more important to measure.

> Calculate new calldata costs
> 
> Where is this in the output?

For now, this is just `gasUsedForL1`. Can dig deeper into "real" cost in https://github.com/web3well/bls-wallet/issues/502

> Let's omit "arbitrum": { "gasUsedForL1": 0 } when not testing on arbitrum.

Fixed

contracts/scripts/measure_gas/token.ts
voltrevo: I'm not sure I understand your point. The VG is a contract, it doesn't care about gas anyway. We care about gas because of the way we want to use it, and the way we want to use it requires that `tx.origin` is compensated.

I can see how ops that don't pay `tx.origin` are useful for internal comparisons (e.g. is the gas usage going up or down), but I think that's a relatively low priority.

The priority is whether we're beating non-bls transactions, and by how much. For this purpose, ops that don't compensate `tx.origin` are not relevant.

voltrevo:
> My gut feeling is that IERC20.transfer's are much more frequent than ETH transfers (such as with stablecoins) and therefore are more important to measure.

This may be different on L2s, but on mainnet they're about the same, possibly ETH transfers are slightly more common:

Screen Shot 2023-02-09 at 12 08 35 pm

Screen Shot 2023-02-09 at 12 11 54 pm

(Continuation of [Script Used to Analyze Transfer Amounts](https://hackmd.io/0q7H3Ad0Su-I4RWWK8wQPA#Script-Used-to-Analyze-Transfer-Amounts).)

voltrevo:
> > Testing locally with geth (yarn start), the script crashes with execution reverted on an early estimateGas call
> 
> I'm seeing this as well, not sure what the cause is. If you manually specify a gas limit it will still fail for unknown reasons. Broke off into #515. Don't think it will be high priority as most people will run this against a remote network or hardhat node to debug.
>

👍

> > ```
> > When running against arbitrum goerli, I spent some time troubleshooting before realizing there's a hardcoded networkConfigName: "local", in run.ts which needs to be changed to "arbitrum-goerli". We should be able to detect the --network setting instead (process.argv if necessary). Alternatively, there at least needs to be a clear instruction about this.
> > ```
> 
> Configuration is mentioned in the [README](https://github.com/web3well/bls-wallet/pull/341/files#diff-ac4b5a05a6426e2ff4c0b120dbd12660d2367803a1775a026e2282718f4c8730R9-R11), but it makes sense to make this path default to the network it is being run on. Fixed.
>

Ah my mistake. Cheers for the network defaulting though, big improvement 👍.

> > After fixing (2) above, the first 3 configs work but the 4th fails with:
> > running transfer, mode blsExpanderAddress. # of transactions: 10
> > registering public keys...
> > <ref *1> Error: cannot estimate gas; transaction may fail or may require manual gas limit [ See: https://links.e
> 
> I was unable to reproduce this, was able to run on local hardhat node fine. May need your configuration to reproduce. I've tended to see this 4th step fail if the `BLSExpander` contract is not initialized with the correct VG contract, usually when I have multiple contract deploys remotely.
>

Hmm darn it. I don't think this is worth chasing down right now, happy to just see if it resurfaces or not 👍.

> > I don't see the L1 gas price or any per-transaction measurements
> 
> L1 stats were broken off into #502 . The per-transaction measurements can be calculated by just taking the total cost(s) and dividing by the batch size. I tried to keep the information in the JSON output limited but enough that a person or machine could write a report based on just that info, esp the txn hashes.
>

Ok, happy for that result, I'm just confused by your PR description:

> - [x] Include gas price for L1 & L2, total & per transaction ETH costs

This part is wrong, no?

> > Calculate new calldata costs
> > Where is this in the output?
> 
> For now, this is just `gasUsedForL1`. Can dig deeper into "real" cost in #502
>

Sounds good, just thrown off by the phrasing in the PR description.

> > Let's omit "arbitrum": { "gasUsedForL1": 0 } when not testing on arbitrum.
> 
> Fixed

🙌

contracts/scripts/measure_gas/token.ts
voltrevo: (Let's continue this outside of the current PR, since it's non-blocking.)

contracts/scripts/measure_gas/README.md
```diff
@@ -0,0 +1,35 @@
+# Gas Measurments Script
+
+This script runs a variety of gas measurements on both normal EVM transactions as well as their BLS Wallet equivilents to aid in analyzing the costs and benefits of each. It specifically targets optimisitic rollups but should be runnable on any EVM network/chain.
```
voltrevo:
```suggestion
This script runs a variety of gas measurements on both normal EVM transactions as well as their BLS Wallet equivalents to aid in analyzing the costs and benefits of each. It specifically targets optimisitic rollups but should be runnable on any EVM network/chain.
```

voltrevo (approved):

jacque006:
@voltrevo Updated PR TODO list to more accurately reflect what was done in this work.

Issue #281, opened 2022-07-20
Add recovery to extension


voltrevo:
Using this mechanism:
https://github.com/web3well/bls-wallet/blob/main/contracts/contracts/VerificationGateway.sol#L196

Let users know that multisig may be a good combo when doing this.

Probably some overlap with:
https://github.com/web3well/bls-wallet/issues/220

kautukkundan:
related #220

blakecduncan:
I caught up with James on this ticket and will explain the expected UX below

In the Instant wallet
- The users sets the recovery hash using an address in the Quill extension (ex: wallet 0) and arbitrary salt
- After the hash is set the user can click to copy the salt and the Instant wallet smart contract address. These can then be used in the quill extension to recover the wallet

In the quill extension
- The user must input the salt and the smart contract address from above
- We will use the smart contract address to look up the bls key hash (this is the hash of the instant wallet) and the quill extension will generate a new bls key.  With the hash, salt, new bls key and signature from the existing wallet (ex: wallet 0) in the quill extension we can call the `recoverWallet` function

@jzaki just want to confirm this is correct?
Also, FYI @kautukkundan

jzaki:
To nest it a little more:

1. User installs quill extension (eg first-time web3 users upgrading their instant wallet to a more secure one)
  - first bls wallet address in quill to be used as recovery address
  - this must be funded with some eth to recover the instant wallet

2. In the Instant wallet
  - The user sets the recovery hash using an address in the Quill extension (ex: wallet 0) and arbitrary salt
  - After the hash is set (regular transaction toast) the user can click a button in the recovery menu to copy the salt and the Instant wallet smart contract address (convenience for recovery params)
  - These can then be used in the quill extension to recover the wallet

3. In the quill extension
  - Instead of "add" wallet, user clicks button to "recover a wallet" a bls wallet
  - The user must input (just paste?) the salt and the smart contract address copied from above
  - On user confirmation, the extension will:
    - use the smart contract address to look up the bls key hash (**_param2_**) (this is the hash of the instant wallet's bls key)
    - generate a new bls key to replace that of the instant wallet (**_param 4_**)
    - sign the instant wallet address with the new bls key (signatures is **_param 1_**)
    - With the 4 params: signature, hash, salt (**_param 3_**), and new bls key; call `recoverWallet` from wallet 0
      - (ie create the transaction, sign with wallet 0's bls key, then submit signed tx to aggregator)

[...]

Issue #220, opened 2022-06-06
Add recovery & upgrade wallet functions to clients


jacque006:
We should add client functionality to `bls-wallet-clients` to allow consumers to call wallet recovery & upgrade functionality for their wallet such that they can:

- Set a new BLS public key
- Set a new Verification Gateway (Trusted gateway)
- Set a new wallet implementation (Proxy Admin Function Hash)
- Set a recovery hash
- Commit any pending updates
- Execute a wallet recovery in case of attack

We may also want to add helpful functions to track the above, such as:

- Query if there are any pending updates to the BLS PubKey, Trusted Gateway, PA Func hash, or recovery hash 
- Query how long if will be until those pending updates can be committed

So that wallets/services can clearly alert/communicate with users when updates are pending.

See https://github.com/web3well/bls-wallet/blob/main/contracts/test/recovery-test.ts for some examples of this can be done.

PR #342, opened 2022-09-20
Wallet recovery

blakecduncan:
Merging is blocked by https://github.com/web3well/bls-wallet/pull/367

## What is this PR doing?

- Adding an addRecoveryWallet method to the RPC
- UI changes to allow the user to recover

Video demo of recovery
https://www.loom.com/share/34e84b23cd8647058dc92afc43a9ed28

- You can test by recovering from an instant wallet to the Quill extension (as seen in the demo video above)
- This won't currently work on Goerli because there is a bug fix in progress.  If you want to test it locally you need to use `bls-wallet-clients@0.8.0-2d28b7b`

https://github.com/web3well/bls-wallet/issues/281

## Checklist

## Guidelines


extension/source/background/KeyringController.ts
```diff
@@ -154,6 +158,76 @@ export default class KeyringController {
...
+          [
+            addressSignature,
+            recoveryWalletHash,
+            recoverySaltHash,
+            newWalletWrapper.PublicKey(),
+          ],
+        ),
+      };
+
+      this.InternalRpc().eth_sendTransaction({
```
blakecduncan: I think calling this function brings up the transaction approval modal.  I'm not sure if we want that to happen or if I should just call the method directly doing something like

```
  const bundle = signerWallet.sign({
    nonce: await signerWallet.Nonce(),
    actions: [
      {
        ethValue: 0,
        contractAddress: verificationGatewayContract.address,
        encodedFunction:
          verificationGatewayContract.interface.encodeFunctionData(
            'recoverWallet',
            [
              addressSignature,
              instantWalletHash,
              recoverySaltHash,
              newWallet.PublicKey(),
            ],
          ),
      },
    ],
  });
```

extension/source/background/KeyringController.ts
```diff
@@ -154,6 +158,76 @@ export default class KeyringController {
...
+        ...tx,
+        gas: undefined,
+        gasPrice: undefined,
+        data: undefined,
+      });
+
+      // Add new private key
+      await this.InternalRpc().addAccount(newPrivateKey);
```
blakecduncan: This is related to the above comment.  I need to test this more but I might need to figure out a way to get this to wait until the transaction is approved by the user before calling the addAccount function.

extension/source/background/KeyringController.ts
```diff
@@ -154,6 +159,30 @@ export default class KeyringController {
...
+     * @param recoveryWalletAddress Smart contract address
+     * of wallet being recovered
+     * @param recoverySaltHash Salt used to set the recovery
+     * hash on the wallet that is being recovered
+     * @param signerWalletPrivateKey the private key of the wallet that is used
+     * to generate the recovery hash. This wallet will sign the 'recoverWallet'
+     * request.
+     */
+    addRecoveryWallet: async ({
+      params: [recoveryWalletAddress, recoverySaltHash, signerWalletPrivateKey],
```
blakecduncan, line 189: @kautukkundan There's one thing I wanted to raise that I'm doing differently than what we discussed on our call today.

I'm now accepting a `signerWalletPrivatekey` as a param here.  That's because, based off our convo earlier, you are going to generate a wallet on the fly to be used for the recovery.  This way, you can just pass that wallet's PK to this function.

extension/source/background/KeyringController.ts
```diff
@@ -214,4 +243,93 @@ export default class KeyringController {
...
+      signerWallet.PublicKey(),
+      netCfg.addresses.verificationGateway,
+      await this.ethersProvider.read(),
+    );
+
+    // Thought about using this.InternalRpc().eth_sendTransaction() here.
+    // However since we are generating a wallet on the fly and not using
+    // an existing wallet in the keyring I am calling creating a bundle
+    // manually and submitting it to the aggregator.
+    const bundle = signerWallet.sign({
```
blakecduncan, line 342: I also didn't change this to use the `eth_sendTransaction` function because that function uses the selected wallet in the quill extension.  If we are generating a wallet on the fly, I think we will need to create the bundle ourselves here.  Fyi @kautukkundan

extension/source/Home/Wallet/Wallets/Recovery/RecoverWalletModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+        }}
+      >
+        <div className="p-8 h-full w-100 flex flex-col">
+          <WorkflowNumbers max={3} current={pageIndex} />
+          <div className="mt-8 h-100 flex-grow">
+            {
+              [
+                <StepOneInfo
+                  key={1}
+                  onComplete={() => {
```
JohnGuilding, line 75: I accidentally clicked next on the second page without copying the address and found that I couldn't go back. I instead had to close to model or refresh the page (can't remember which) to start the process again and access the address.

Wondering if there is an easy way to go back a step?

extension/source/Home/Wallet/Wallets/Recovery/RecoverWalletModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+        </div>
+      ))}
+    </div>
+  );
+};
+
+const RecoverWalletModal = () => {
+  const [modalIsOpen, setIsOpen] = useState(false);
+  const [pageIndex, setPageIndex] = useState(0);
+  const [walletPrivateKey, setWalletPrivateKey] = useState<string>('');
```
JohnGuilding, line 37: Do we need to explicitly add the type here? Can see that we're letting typescript infer the type for the above useState's

extension/source/Home/Wallet/Wallets/Recovery/StepOneInfo.tsx
```diff
@@ -0,0 +1,38 @@
...
+const StepOneInfo: FunctionComponent<{
+  onComplete: () => void;
+}> = ({ onComplete }) => {
+  return (
+    <div className="flex flex-col gap-4 h-full">
+      <div className="flex-grow">
+        <div className="text-[14pt]">Recover existing wallet in Quill</div>
+        <div className="text-[10pt] text-grey-700 leading-loose">
+          You can recover existing instant BLS wallets into Quill. This is a
+          simple 2 step process which requires you to copy-paste some stuff from
```
JohnGuilding: Thoughts on a slightly different wording here instead of 'stuff'? Maybe something like:

1. "...requires you to copy-paste two values from..." or 
2. "...requires you to copy-paste two pieces of information from..."

extension/source/Home/Wallet/Wallets/Recovery/StepOneInfo.tsx
```diff
@@ -0,0 +1,38 @@
...
+          Quill to the instant wallet and then again from instant wallet to
+          Quill
+        </div>
+        <br />
+        <div className="text-[10pt] text-grey-700 leading-loose font-bold mt-2">
+          Do not close this modal until you have completed all the steps that
+          follows else you will lose access to your original keys!
```
JohnGuilding: Thoughts on:

"...steps that follow, else you..."

extension/source/Home/Wallet/Wallets/Recovery/StepThreeRecover.tsx
```diff
@@ -0,0 +1,92 @@
...
+import Button from '../../../../components/Button';
+import { useQuill } from '../../../../QuillContext';
+
+const StepThreeRecover: FunctionComponent<{
+  onComplete: () => void;
+  walletPrivateKey: string;
+}> = ({ onComplete, walletPrivateKey }) => {
+  const { rpc } = useQuill();
+
+  const [salt, setSalt] = useState<string>('');
```
JohnGuilding, line 14: Similar point here and line below, do we want to explicitly state the type?

extension/source/Home/Wallet/Wallets/Recovery/StepThreeRecover.tsx
```diff
@@ -0,0 +1,92 @@
...
+          wallet will be visible inside Quill.
+          <br />
+          <div
+            className={[
+              'text-[10pt]',
+              'text-grey-700',
+              'leading-loose',
+              'font-bold',
+              'mt-2',
+            ].join(' ')}
```
JohnGuilding, line 41: Haven't seen this way of adding classNames before - nice. Might be a silly question but is there a reason why you would pick this way over say the syntax on line 26?

extension/source/Home/Wallet/Wallets/Recovery/StepTwoWalletCreation.tsx
```diff
@@ -0,0 +1,76 @@
...
+          This is the wallet address which will be used to recover your instant
+          wallet.
+        </div>
+        <br />
+        <div className="text-[10pt] text-grey-700 leading-loose font-bold mt-2">
+          Copy the address and paste it in the instant wallet.
+        </div>
+        <div className="mt-4 bg-grey-900 bg-opacity-25 rounded-md p-5">
+          {loading ? (
+            <div className="flex gap-4 text-[12pt] items-center ">
```
JohnGuilding: Tiny bit of whitespace :)

extension/source/background/KeyringController.ts
```diff
@@ -154,8 +172,53 @@ export default class KeyringController {
...
     },
+
+    /**
+     * Recovers an existing BLS wallet and adds the new
+     * recovered wallet to the keyring
+     * @param recoveryWalletAddress Smart contract address
+     * of wallet being recovered
+     * @param recoverySaltHash Salt used to set the recovery
+     * hash on the wallet that is being recovered
+     * @param signerWalletPrivateKey the private key of the wallet that is used
```
JohnGuilding: Nitpicky but just asking for consistency's sake, should the word "the" after `signerWalletPrivateKey` be capitalised here? Or should the capitalised words after the param names above this be uncapitalised?

extension/source/background/KeyringController.ts
```diff
@@ -214,4 +277,93 @@ export default class KeyringController {
...
+    const network = await this.network.read();
+    const netCfg = getNetworkConfig(network, this.multiNetworkConfig);
+
+    // Create new private key for the wallet we are recovering to.
+    const newPrivateKey = generateRandomHex(256);
+
+    const addressSignature = await this.signWalletAddress(
+      recoveryWalletAddress,
+      newPrivateKey,
+    );
```
JohnGuilding, line 314: Might not be fully understanding here, my two questions are:

1. Is this creating a signature using a randomly generated private key and an already generated contract address? 

2. If so, how is this linking an already created address to a randomly generated private key? My knowledge of ECDSA cryptography is that the address must be derived from the public key, which is derived from the private key. But I admit my knowledge may be lacking here as these are smart contract wallets and we're not using ECDSA? And I don't fully understand how they are instantiated.

extension/source/background/KeyringController.ts
```diff
@@ -110,6 +115,19 @@ export default class KeyringController {
       return networkData.address;
     },
 
+    createTempAccount: async (_request) => {
```
JohnGuilding, line 119: Same question for `addRecoveryWallet()`, do we need to add unit tests for this new logic? I'm assuming as this is the extension which isn't meant to be production code, we are happy with not adding tests?

JohnGuilding (commented):
Tested locally and works well. Was able to recover a wallet successfully from the instant wallet into Quill.

I'm not too confident reviewing the changes in the `KeyringController.ts` as it's quite a lot of new stuff for me, so would be good to get a second opinion on that specific file :)

extension/source/Home/Wallet/Wallets/Recovery/StepThreeRecover.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const [salt, setSalt] = useState<string>('');
+  const [instantWalletAddress, setInstantWalletAddress] = useState<string>('');
+
+  const handleRecover = async () => {
+    await rpc.addRecoveryWallet(
+      instantWalletAddress,
+      ethers.utils.formatBytes32String(salt),
+      walletPrivateKey,
+    );
```
jacque006, line 22: Do we need an error state/handling if this RPC call fails?

extension/source/Home/Wallet/Wallets/Recovery/StepTwoWalletCreation.tsx
```diff
@@ -0,0 +1,76 @@
...
+    const createWallet = async () => {
+      setLoading(true);
+      const { address, privateKey } = await rpc.createTempAccount();
+      setWalletAddress(address);
+      setWalletToParent(privateKey);
+      setLoading(false);
+    };
+
+    createWallet();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
jacque006: Why are deps disabled here? Should just be `rpc`.

extension/source/Home/Wallet/Wallets/Recovery/StepTwoWalletCreation.tsx
```diff
@@ -0,0 +1,76 @@
...
+            <div className="flex gap-4 text-[12pt] items-center ">
+              Generating wallet
+              <div className="animate-spin relative">
+                <CircleNotch size={30} />
+              </div>
+            </div>
+          ) : (
+            <div className="flex justify-between">
+              <div className="font-mono text-[11pt]">{walletAddress}</div>
+              {/* eslint-disable-next-line */}
```
jacque006, line 60: What rule was disabled here? Better to disable a specific rule than all linting rules.

extension/source/background/KeyringController.ts
jacque006: @JohnGuilding We generally don't have tests for the extension as they would be complicated to setup & manage. However, as extension gets more stable might make sense to add some e2e tests for more common paths. Would save us some regression testing as well.

extension/source/background/KeyringController.ts
```diff
@@ -110,6 +115,19 @@ export default class KeyringController {
...
+      const pKey = generateRandomHex(256);
```
jacque006: Use a `BN254` spaced key instead of just a random hex. See https://github.com/thehubbleproject/hubble-bls/blob/master/src/mcl.ts#L186 . We should expose that via the signer in the client library somewhere.

extension/source/background/KeyringController.ts
```diff
@@ -214,4 +277,93 @@ export default class KeyringController {
...
+  async recoverWallet(
+    recoveryWalletAddress: string,
+    recoverySaltHash: string,
+    signerWalletPrivateKey: string,
+  ): Promise<string> {
+    const network = await this.network.read();
+    const netCfg = getNetworkConfig(network, this.multiNetworkConfig);
+
+    // Create new private key for the wallet we are recovering to.
+    const newPrivateKey = generateRandomHex(256);
```
jacque006: See above comment on BLS private key gen.

jacque006 (commented):
It might also be beneficial to move some of the functionality into the clients lib so that other wallets/dApps can take advantage of this functionality, similar to https://github.com/web3well/bls-wallet/issues/220 . But I would be okay if that's follow up work.

extension/source/background/KeyringController.ts
blakecduncan: @jacque006  Ahh interesting we do `generateRandomHex` a couple of other places in the keyring controller which is where I copied this from.  Should I also switch those?

Also, just curious, what is the benefit of `BN254` over a random hex?

blakecduncan: Just want to double check that this is the right way to call that function?

```
randFr().serializeToHexStr()
```

extension/source/background/KeyringController.ts
blakecduncan: @JohnGuilding good question

1. Yeah it is.  It's the smart contract address of the 'instant wallet'.  So the wallet we are recovering.  Signed by the newly generated private key that will become the new private key for the address we are recovering.

3. I may not be the best at explaining this but this is my understanding. This signature is basically how the trusted wallet (the one that was set in the recovery hash) says that this private key now has permission to act on behalf of the instant wallet address.  When the recovery happens this private key will create a new public key but instead of deterministically generating an address the public key will map to the existing instant wallet address.  This signature is a way of verifying and setting this PK as the PK to the old address without revealing the private key.

extension/source/Home/Wallet/Wallets/Recovery/StepThreeRecover.tsx
kautukkundan: due to tailwind-css the line length for classes sometimes exceeds 80 which makes the linter unhappy.

extension/source/Home/Wallet/Wallets/Recovery/RecoverWalletModal.tsx
kautukkundan: My personal preference is to have explicit typing everywhere. Copied the above 2 lines from somewhere else in the codebase so missed adding type to it.

extension/source/Home/Wallet/Wallets/Recovery/StepTwoWalletCreation.tsx
kautukkundan: it wanted to also have a "keyboard input" for the copy button which didn't make much sense to me.

extension/source/Home/Wallet/Wallets/Recovery/StepTwoWalletCreation.tsx
kautukkundan: I sometimes don't understand these deps. Like it's not possible to change some deps without explicit reload which is gonna rerender anyway.

But yeah, let's add rpc here.

extension/source/Home/Wallet/Wallets/Recovery/StepThreeRecover.tsx
kautukkundan: error-handling is not great across the extension's codebase. I think we should have a separate task for improving errors/fails altogether. That's why left it simple here as well.

extension/source/Home/Wallet/Wallets/Recovery/RecoverWalletModal.tsx
JohnGuilding: I may lean towards letting typescript infer the type but don't have a strong opinion either way, so happy with either option as long as we're being consistent :)

extension/source/Home/Wallet/Wallets/Recovery/StepTwoWalletCreation.tsx
JohnGuilding: Useful comment on this issue with links for further reading from Dan Abramov
https://github.com/facebook/create-react-app/issues/6880#issuecomment-485912528

extension/source/background/KeyringController.ts
JohnGuilding: Thanks for the answer, have a better understanding now. Will make sure to re-read your recovery article too

extension/source/background/KeyringController.ts
kautukkundan: yeah makes sense 

> what is the benefit of BN254 over a random hex?

If I understand correctly, The random number generated might sometimes lie outside the range of the BN curve we are using. This function ensures it's in range.

jacque006: @kautukkundan correct. If the key is out of range, we just hash it into a key that is in range. I have a bucklist item/change that will be more explicit about this in the future https://github.com/thehubbleproject/hubble-bls/issues/30

extension/source/Home/Wallet/Wallets/Recovery/RecoverWalletModal.tsx
blakecduncan: Fixed!

extension/source/Home/Wallet/Wallets/Recovery/RecoverWalletModal.tsx
voltrevo: @kautukkundan Are you sure that's your preference in this case where TypeScript is correctly inferring the type? Would you also add `<boolean>` and `<number>` to the previous lines?

kautukkundan: @voltrevo Fine with me as is in this _particular_ case. 

The problem starts happening when using complex types. eg - let's say a new `Address` type is defined as a string that always starts with "0x". In this case initializing with default '' would also work as it'll infer type string, but may break code later when things get complex. This had been the case earlier when using sol.G1 and sol.G2 types in the HubbleBLS library.

So looking at broader picture and anticipating future I prefer being explicit even with simpler types to maintain consistency, If that makes sense.

kautukkundan (approved):
LGTM 🚀

Nerotool:
SEND A DM TO NEROTOOL_ ON INSTAGRAM FOR HACKING

PR #343, opened 2022-09-20
4337 exploration

voltrevo:
## What is this PR doing?

*These are early ideas and not intended for merging at this stage.*

This is a draft of how we might integrate [eth-infinitism's version of 4337](https://github.com/eth-infinitism/account-abstraction/blob/develop/eip/EIPS/eip-4337.md). I'm referencing this version specifically because the [official EIP](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4337.md) doesn't currently have the signature aggregation support we need.

A wallet integrating us as a 4337 wallet would still need to understand how to rewrite a transaction as (or otherwise generate) a [`UserOperation`](https://github.com/eth-infinitism/account-abstraction/blob/42b643c/contracts/interfaces/UserOperation.sol#L19-L32) (aka [`UserOperation4337`](https://github.com/web3well/bls-wallet/blob/2e68505/contracts/contracts/interfaces/UserOperation4337.sol)). Critically, this means knowing how to do a BLS signature.

A `UserOperation` looks like this:

```solidity
    struct UserOperation {

        address sender;
        uint256 nonce;
        bytes initCode;
        bytes callData;
        uint256 callGasLimit;
        uint256 verificationGasLimit;
        uint256 preVerificationGas;
        uint256 maxFeePerGas;
        uint256 maxPriorityFeePerGas;
        bytes paymasterAndData;
        bytes signature;
    }
```
https://github.com/eth-infinitism/account-abstraction/blob/42b643c/contracts/interfaces/UserOperation.sol#L19-L32

Gas is a big thing we're doing differently. Currently our idea is still to reward `tx.origin` in an unspecified way (ie `tx.origin` just detects that they'll get paid by simulating the tx), so we'd be opting out of the 4337 gas mechanisms by setting gas limits high and fee per gas fields to zero.

`initCode` can be left as an empty byte array, but it means doing the creation step explicitly.

`callData` is an encoded call to [`BLSWallet#performOperation`](https://github.com/web3well/bls-wallet/blob/2e68505/contracts/contracts/BLSWallet.sol#L148) which has been changed to allow calls from the 4337 entry point as well as the verification gateway.

`signature` is the result of signing the `UserOperation` with an empty `signature` field (TODO: there's a bit of a wrinkle here, need to update based on `requestId`). (Note: This means signing for a 4337 `UserOperation` is different for signing an equivalent `Bundle`. It might be possible to make these signatures equal but it would mean doing more data wrangling on-chain which means more gas and it doesn't seem to have much benefit (might need to prevent collisions by using a different bls domain though).)

As it stands, when these `UserOperation`s go into the mempool they won't work out-of-the-box with clients designed for 4337 for two reasons:
- We're not explicitly offering any gas compensation
- 4337 relies on on-chain code for defining how the off-chain signature aggregation works ([ref](https://github.com/eth-infinitism/account-abstraction/blob/8083c53/eip/EIPS/eip-4337.md?plain=1#L126-L128)), and we're not supplying that on-chain code right now (it's missing from blsLib but can hopefully be added)

This means that we either need to:
- convince 4337 implementers (in clients aka nodes) to also implement our alternative reward mechanism and specifically add BLS signature aggregation
- OR add independent nodes that find our `UserOperation`s and aggregate+submit them (ie - something like the aggregator we have now)
- OR pivot to 4337's gas compensation (and just being a 4337 wallet in general?) and add the on-chain BLS signature aggregation code ([ref](https://github.com/eth-infinitism/account-abstraction/blob/8083c53/eip/EIPS/eip-4337.md?plain=1#L126-L128))

NA

Nope.

## Checklist
- ~[ ] I have manually tested these changes~

## Guidelines

### DIFF
--- a/contracts/clients/src/BlsWalletWrapper.ts
+++ b/contracts/clients/src/BlsWalletWrapper.ts
@@ -57,6 +57,7 @@ export default class BlsWalletWrapper {
     const initFunctionParams =
       BLSWallet__factory.createInterface().encodeFunctionData("initialize", [
         verificationGatewayAddress,
+        ethers.constants.AddressZero, // TODO: 4337 EntryPoint address
       ]);
 
     return ethers.utils.getCreate2Address(
--- a/contracts/contracts/BLSWallet.sol
+++ b/contracts/contracts/BLSWallet.sol
@@ -6,7 +6,7 @@ pragma abicoder v2;
 //To avoid constructor params having forbidden evm bytecodes on Optimism
 import "@openzeppelin/contracts/proxy/utils/Initializable.sol";
 import "./interfaces/IWallet.sol";
-
+import "./interfaces/UserOperation4337.sol";
 
 /** Minimal upgradable smart contract wallet.
     Generic calls can only be requested by its trusted gateway.
@@ -23,6 +23,7 @@ contract BLSWallet is Initializable, IWallet
 
     // BLS variables
     address public trustedBLSGateway;
+    address public trustedEntryPoint;
     address pendingBLSGateway;
     uint256 pendingGatewayTime;
 
@@ -49,10 +50,12 @@ contract BLSWallet is Initializable, IWallet
     );
 
     function initialize(
-        address blsGateway
+        address blsGateway,
+        address entryPoint
     ) external initializer {
         nonce = 0;
         trustedBLSGateway = blsGateway;
+        trustedEntryPoint = entryPoint;
     }
 
     receive() external payable {}
@@ -144,7 +147,7 @@ contract BLSWallet is Initializable, IWallet
      */
     function performOperation(
         IWallet.Operation calldata op
-    ) public payable onlyTrustedGateway thisNonce(op.nonce) returns (
+    ) public payable onlyTrusted thisNonce(op.nonce) returns (
         bool success,
         bytes[] memory results
     ) {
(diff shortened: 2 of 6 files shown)


voltrevo:
@jzaki I tried fixing getInitializeData but it didn't resolve the gas limit issue. It does seem to be about the actual contract size. I tested this by pruning the recovery feature and got the remaining tests working (which includes deploying the downsized `VerificationGateway`). There's a branch for this [here](https://github.com/web3well/bls-wallet/tree/4337-exploration-prune-test). We can also reproduce the same issue by [artificially bloating the contract to make it too big again](https://github.com/web3well/bls-wallet/tree/4337-exploration-prune-bloat-test).

Anyway, I don't think it has much bearing on our decision process here - we can chase it up later.

jzaki:
> @jzaki I tried fixing getInitializeData but it didn't resolve the gas limit issue. 

The changes to the constructor params means the initializeData bytes need to be updated, otherwise the tx reverts (out of gas).

You _may_ legitimately be out of gas when deploying the VerificationGateway if its bytecode increased by a non-small amount.
By importing UserOperation.sol you're importing definitions (UserOperation), but also the bytecode, into to VerificationGateway.

voltrevo:
> The changes to the constructor params means the initializeData bytes need to be updated, otherwise the tx reverts (out of gas).

I think you might have misread that I fixed it? Let me know if [this](https://github.com/web3well/bls-wallet/pull/343/commits/92f6ec846734aab41f630d232c28563e909987b0#diff-b48b32d4b9972fa32985b2f1dfcf06f7a35cc8f8d77c549b9332dc651d55fcbaR361) looks wrong to you.

> You may legitimately be out of gas when deploying the VerificationGateway if its bytecode increased by a non-small amount.
By importing UserOperation.sol you're importing definitions (UserOperation), but also the bytecode, into to VerificationGateway.

Yep this is what's going on. Just trying to make things clear ([see this previously linked demonstration](https://github.com/web3well/bls-wallet/tree/4337-exploration-prune-bloat-test) (see most recent commit in that branch)).

This means there's either a hardhat issue here that needs fixing or there is a genuine hard limit that forces us to use a separate contract. Possibly the limit does not apply to arbitrum and maybe other L2s. Do you have any thoughts on this?

jacque006:
@voltrevo @jzaki Can this PR be closed? (archived, branch & PR may be useful for later)

PR #344, opened 2022-09-21
Improve clients docs

### LINKED ISSUE #330: Improvements to use_bls_wallet_clients.md
- `yarn install` -> `yarn add`
  - "You will also need to have ethers installed." misleading: ethers is a dependency of bls-wallet-clients
  - Typo in example: `BlsWallerWrapper` -> `BLSWalletWrapper`
  - Could use some more guidance for eg specifying addresses, getting private key

voltrevo:
## What is this PR doing?

A variety of improvements to make the guide easier to follow, particularly fixes to the example code.

Try the example code.

Resolves #330.

## Checklist

## Guidelines

### DIFF
--- a/docs/use_bls_wallet_clients.md
+++ b/docs/use_bls_wallet_clients.md
@@ -12,8 +12,6 @@ yarn add bls-wallet-clients
 # deno in example further below
 ```
 
-You will also need to have [ethers](https://docs.ethers.io) installed.
-
 ## Import
 
 ```typescript
--- a/docs/use_bls_wallet_clients.md
+++ b/docs/use_bls_wallet_clients.md
@@ -16,7 +16,7 @@ yarn add bls-wallet-clients
 
 ```typescript
 import { providers } from "ethers";
-import { Aggregator, BLSWalletWrapper, getConfig } from "bls-wallet-clients";
+import { Aggregator, BlsWalletWrapper, getConfig } from "bls-wallet-clients";
 ```
 
 ### Deno
@@ -25,7 +25,7 @@ You can use [esm.sh](https://esm.sh/) or a similar service to get Deno compatibl
 
 ```typescript
 import { providers } from "https://esm.sh/ethers@latest";
-import { Aggregator, BLSWalletWrapper, getConfig } from "https://esm.sh/bls-wallet-clients@latest";
+import { Aggregator, BlsWalletWrapper, getConfig } from "https://esm.sh/bls-wallet-clients@latest";
 ```
 
 ## Get Deployed Contract Addresses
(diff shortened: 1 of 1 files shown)


docs/use_bls_wallet_clients.md
```diff
@@ -39,55 +37,91 @@ If you would like to deploy to a remote network, see [Remote development](./remo
...
-  netCfg.contracts.verificationGateway,
+  netCfg.addresses.verificationGateway,
   provider
 );
 
-const erc20Address = "0x...";
+const erc20Address = netCfg.addresses.testToken; // Or some other ERC20 token
 const erc20Abi = [
-    "function transfer(address to, uint amount) returns (bool)",
+    'function mint(address to, uint amount) returns (bool)',
```
jacque006, line 76: Although we include `mint` on our test token, and `_mint` is a function in the OpenZeppelin ERC20 implementation it is not part of the standard ERC20 interface. I would recommend keeping this as the more common `transfer` here and below.

```suggestion
    'function transfer(address to, uint amount) returns (bool)',
```

docs/use_bls_wallet_clients.md
```diff
@@ -39,55 +37,91 @@ If you would like to deploy to a remote network, see [Remote development](./remo
...
 
-const aggregator = new Aggregator("https://rinkarby.blswallet.org");
-await aggregator.add(bundle);
+const aggregator = new Aggregator('https://arbitrum-goerli.blswallet.org');
+
+console.log('Sending bundle to the aggregator');
+const addResult = await aggregator.add(bundle);
+
+if ('failures' in addResult) {
+  throw addResult.failures.join('\n');
```
jacque006: I would not recommend throwing non-Error object or primitives as no stack information will be included, making it harder to track down.

```suggestion
  throw new Error(addResult.failures.join('\n'));
```

docs/use_bls_wallet_clients.md
```diff
@@ -39,55 +37,91 @@ If you would like to deploy to a remote network, see [Remote development](./remo
...
+}
+
+console.log('Bundle hash:', addResult.hash);
+
+const checkConfirmation = async () => {
+  console.log('Checking for confirmation')
+  const maybeReceipt = await aggregator.lookupReceipt(addResult.hash);
```
jacque006, line 114: As a future improvement, we could add a `wait` function to the return value from `aggregator.add` that does this under the covers. This would also help partially match the familiarity of a ethers.js Transaction.

```typescript
const addResult = await aggregator.add(bundle);
await addResult.wait();
```

jacque006 (approved):

docs/use_bls_wallet_clients.md
voltrevo: I see what you mean, but the good thing about `mint` is it allows this to work without any other set up (sending tokens to the contract wallet some other way first).

What do you think we should do instead?

docs/use_bls_wallet_clients.md
voltrevo: Ah good point. Fixed.

docs/use_bls_wallet_clients.md
voltrevo: Interesting idea. We'd definitely benefit from the aggregator being able to let us know when the transaction has been processed instead of polling it as new blocks come in. In particular, we often hit the aggregator just before it receives its own confirmation this way, which means we wait for a whole extra unrelated block.

In terms of the api, we're currently providing a direct translation of the http api, so I'd rather add something like `waitForReceipt` rather than shoehorning an exception like this. Reworking the API to be more standard/like ethers could be a good idea too though.

docs/use_bls_wallet_clients.md
jacque006: I originally wrote this with the assumption that it could be any arbitrary token. I think if we the example to use our config for populating the contract we could include both and just leave a comment on the `mint`.

```typescript
const erc20Abi = [
    'function transfer(address to, uint amount) returns (bool)',
    // ... other ERC20 functions
    // Included on BLS Wallet test token for convenience, remove if using another token.
    'function mint(address to, uint amount) returns (bool)',
];
```

PR #349, opened 2022-10-02
update config for extension

kautukkundan:
## What is this PR doing?
For unknown reason extension wasn't able to resolve localhost to 127.0.0.1 so updated the config to use the IP by default

import the extension in browser and test creating wallet

none

## Checklist

## Guidelines

### DIFF
--- a/extension/config.example.json
+++ b/extension/config.example.json
@@ -18,7 +18,7 @@
       "chainId": "0x7a69",
       "displayName": "Local Network",
       "logo": "",
-      "rpcTarget": "http://localhost:8545",
+      "rpcTarget": "http://127.0.0.1:8545",
       "chainCurrencyName": "Ethereum",
       "chainCurrency": "ETH",
       "aggregatorUrl": "http://localhost:3000",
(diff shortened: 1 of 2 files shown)


JohnGuilding:
Was there a specific error when localhost wouldn't resolve to 127.0.0.1? 

Just tried to recreate the error in `main` and can create a wallet locally. Are those the correct reproduction steps?

kautukkundan:
The wallet is unable to reach the hardhat server because localhost doesn't resolve the correct local IP. It maybe specific to my system (considering metamask is also unable to resolve localhost to connect to hardhat), I may have borked some network settings. However, setting it to the exact IP will make sure no one else face the same issue as it was a little hard to debug.

image

JohnGuilding (approved):
Gotcha, strange MetaMask isn't working for you either. Was reading about similar issues online and a few stackoverflow answers suggested clearing your browser cache. https://stackoverflow.com/questions/33524826/localhost-not-working-in-chrome-127-0-0-1-does-work

Tested in your branch with no issues - approved

PR #353, opened 2022-10-16
updated inputDecode to parse params

kautukkundan:
## What is this PR doing?
- Update useInputDecode hook to use samczun's Eth signature database - https://sig.eth.samczsun.com. It has much better support for different kinds of function signatures and extends parity's onchain registry
-  Updated Confirmation box to decode the function params

Before             |  After
:-------------------------:|:-------------------------:
Screenshot_2022-10-11_at_10 40 34_AM  |  Screenshot 2022-10-16 at 6 27 56 PM

Pull latest version and build 

## Checklist

## Guidelines

### DIFF
--- a/extension/source/Confirm/TransactionCard.tsx
+++ b/extension/source/Confirm/TransactionCard.tsx
@@ -40,10 +40,11 @@ const TransactionCard: React.FC<SendTransactionParams> = ({
           <div className="text-[9pt] mt-2 font-mono">
             {loading
               ? 'loading params...'
-              : args.map((arg, i) => (
-                  <div key={i}>
-                  {i + 1}. {arg.toString()}
-                </div>
+              : // prettier-ignore
+                args.map((arg, i) => (
+                  <div key={arg}>
+                    {i + 1}. {arg.toString()}
+                  </div>
                 ))}
           </div>
         </div>
--- a/extension/source/Confirm/TransactionCard.tsx
+++ b/extension/source/Confirm/TransactionCard.tsx
@@ -14,7 +14,7 @@ const TransactionCard: React.FC<SendTransactionParams> = ({
   gas,
   gasPrice,
 }) => {
-  const { loading, method } = useInputDecode(data || '0x', to);
+  const { loading, method, args } = useInputDecode(data || '0x');
 
   return (
     <div className="bg-white rounded-md p-4 border border-blue-400">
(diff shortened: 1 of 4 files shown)


extension/source/Confirm/TransactionCard.tsx
```diff
@@ -37,7 +37,16 @@ const TransactionCard: React.FC<SendTransactionParams> = ({
         <div className="break-all">
           details:{' '}
           <span className="font-bold">{loading ? 'loading...' : method}</span>
-          <div className="text-[9pt] mt-2 font-normal">{data}</div>
+          <div className="text-[9pt] mt-2 font-mono">
+            {loading
+              ? 'loading params...'
+              : // prettier-ignore
```
kautukkundan, line 43: was getting too strict with number of spaces to leave at each line hence ignored

jacque006:
Another decoding source to consider in the future: https://sourcify.dev/

extension/source/hooks/useInputDecode.tsx
```diff
@@ -2,100 +2,52 @@ import { ethers } from 'ethers';
...
     };
 
     if (functionData) {
       getMethod();
     }
-  }, [functionData, to, quill]);
+  }, [functionData, quill]);
 
-  return { loading, method };
+  return { loading, method, args: allArgs };
```
jacque006, line 52: It might be a good idea to memoize this to prevent unnecessary renders in consuming components.

jacque006 (approved):

extension/source/hooks/useInputDecode.tsx
kautukkundan: This one's a bit tricky. This hook is called inside TransactionCard component at the top level and the component is only consumed in [Confirm](https://github.com/web3well/bls-wallet/blob/main/extension/source/Confirm/Confirm.tsx#L111) component. 

The way it's setup is that when a multi-action tx is called, it renders a a new TransactionCard with the current tx calldata as "props". When user uses the arrows to move to the next tx the props given to TransactionCard change which triggers a re-render. Re-rendering will always take place because the props change and since the value changes are handled in a custom hook, memoizing the values won't be helpful here - [reference](https://reactjs.org/docs/react-api.html#reactmemo). 

The only way to get some performance benefit is by caching the API call itself. Can go this route if you think it's useful. The current thought is to use session storage for storing the calldata-functionSig pair. Shouldn't be big.

kautukkundan:
> Another decoding source to consider in the future: https://sourcify.dev/

Yeah! There's a small catch though, sourcify will only work if the contracts are already verified
from the website -
> Goodbye YOLO signing 👋. Decode contract interactions with the verified contract's ABI and [NatSpec comments](https://docs.soliditylang.org/en/develop/natspec-format.html) . Show wallet users meaningful information instead of hex strings. 

samczun's database should work for _most_ unverified contracts as well

extension/source/hooks/useInputDecode.tsx
jacque006: Ah ok then yes, we can leave as is. Caching would make sense for more frequent interactions, but is likely overkill at this point. Can do that as separate work if needed.

PR #355, opened 2022-10-16
added UI for entering recovery hash

kautukkundan:
## What is this PR doing?

Add UI for entering recovery hash.

### Step 1:
Screenshot 2022-10-24 at 3 51 34 AM

### Step 2:
Screenshot 2022-10-24 at 3 51 40 AM

Screenshot 2022-10-24 at 3 51 46 AM

### Step 3:
Screenshot 2022-10-24 at 3 51 54 AM

1. Create an instant wallet at http://instant.blswallet.org
2. open quill (after pulling and importing latest changes) and click "Import" on the menu
3. Follow the instructions presented

helps #342 

## Checklist

## Guidelines


jacque006 (commented):
I would prefer if the full recovery functionality was part of this, otherwise developers using Quill might be confused when this does nothing.

I'd also be ok with hiding this new modal and merging dark until the recovery functionality is implemented.

blakecduncan:
> I would prefer if the full recovery functionality was part of this, otherwise developers using Quill might be confused when this does nothing.
> 
> I'd also be ok with hiding this new modal and merging dark until the recovery functionality is implemented.

@kautukkundan regarding this, you could make this a PR into my branch? `wallet-recovery`

kautukkundan:
Ah okay my bad, accidentally pointed to main. @blakecduncan could you rebase/merge main into your branch so I can point it there. Lots of changes after your PR.

blakecduncan:
@kautukkundan Thanks for putting this together!  I had a couple of questions I wanted to ask.

- I think when we talked we mentioned having a description for the user that said what address they should use to set the recovery hash in the instant wallet.  Do we still want to do that?
- After the recovery hash is set in the instant wallet there are two things we need to copy back to Quill.  The salt and the address of the instant wallet.  We don't need to copy the recovery hash like the example in your screenshot.

Let me know if this doesn't make sense!

kautukkundan:
Yes, for sure. 
Just a placeholder right now. Was meaning to sync with you before finalizing it.

kautukkundan:
@blakecduncan I think this is ready for review now. I have implemented the functions but it looks like we may have to sync a couple of times before shipping it. There are a few questions I have plus a few minor todos might come up. But I would suggest you pulling these changes and trying it out.

extension/source/Home/Wallet/Wallets/Recovery/StepThreeRecover.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const { rpc } = useQuill();
+
+  const [salt, setSalt] = useState<string>("");
+  const [instantWalletAddress, setInstantWalletAddress] = useState<string>("");
+
+  const handleRecover = async () => {
+    const pk = await rpc.lookupPrivateKey(walletAddress);
+    await rpc.addRecoveryWallet(
+      instantWalletAddress,
+      ethers.utils.hexlify(ethers.utils.toUtf8Bytes(salt)),
```
kautukkundan: is this the right way?

blakecduncan: Just tested this out and I think that produces a string that's the wrong length.

I switched it to this and tested it out locally and it seemed to work
```
ethers.utils.formatBytes32String(salt)
```

kautukkundan: gotcha! I was confused too. updated in da2264c

blakecduncan:
> @blakecduncan I think this is ready for review now. I have implemented the functions but it looks like we may have to sync a couple of times before shipping it. There are a few questions I have plus a few minor todos might come up. But I would suggest you pulling these changes and trying it out.

@kautukkundan Nice, this looks really good!  I made the change locally to the salt (from my previous comment) and tested it out and it looks like the recovery worked!!

The UI doesn't currently reflect the recovery though.  This is because the `addRecoveryWallet` function I wrote calls
```
await this.InternalRpc().addAccount(privateKey);
```
This function executes before the recovery is finished onchain.  So when it saves the address to local storage it saves the determinist address instead of the recovered address.  And we don't ever update the local storage value.

I'm not sure the best way off the top of my head to address this.  Would you want to sync on that?

kautukkundan:
@blakecduncan In the last 2 commits I have implemented the discussed points:
1. use a temp wallet instead of adding it to keychain
2. once the recovery is complete, swap the address to new address

In my testing on goerli, the recovery and swap works fine but after adding the wallet to quill, it's unable to do any tx. Could you please check that out on local as well?

blakecduncan (approved):
Tested this out locally with `bls-wallet-clients@0.8.0-2d28b7b` and it worked!  I was able to recover the wallet to Quill and then send a transaction from the wallet in Quill.

PR #356, opened 2022-10-16
Import pre-existing seed phrase during Onboarding

### LINKED ISSUE #318: Recover wallet from secret phrase
ie vanilla recovery of your accounts. Not to be confused with our recovery feature that allows you to change your public key.

kautukkundan:
## What is this PR doing?
Adds ability to use existing seed phrase during wallet generation
Screenshot 2022-10-16 at 9 01 14 PM

pull latest version, remove older extension, import again, use any seed

closes #318 

## Checklist

## Guidelines

### DIFF
--- a/extension/source/Home/Onboarding/SecretPhrasePanel.tsx
+++ b/extension/source/Home/Onboarding/SecretPhrasePanel.tsx
@@ -17,8 +17,8 @@ const SecretPhrasePanel: FunctionComponent<{
     setMnemonic(mnemonicPhrase.split(' '));
   };
 
-  const validateAndSetMnemonic = (mnemonic: string) => {
-    const split = mnemonic.split(' ');
+  const validateAndSetMnemonic = (m: string) => {
+    const split = m.split(' ');
     if (split.length === 12) {
       setMnemonic(split);
     }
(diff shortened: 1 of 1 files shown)


extension/source/Home/Onboarding/SecretPhrasePanel.tsx
```diff
@@ -8,14 +10,69 @@ const SecretPhrasePanel: FunctionComponent<{
...
     const mnemonicPhrase = ethers.Wallet.createRandom().mnemonic.phrase;
     setMnemonic(mnemonicPhrase.split(' '));
-  }, []);
+  };
+
+  const validateAndSetMnemonic = (m: string) => {
+    const split = m.split(' ');
+    if (split.length === 12) {
+      setMnemonic(split);
+    }
```
jacque006: 1. We should display some sort of error message in the UI if the mnemonic is not valid, and not allow the user to set it
2. You should be able to use [ethers.Wallet.fromMnemonic](https://docs.ethers.io/v5/api/signer/#Wallet.fromMnemonic) to validate it is correct (I assume it throws an error), rather than just doing the length check. This will ensure it has the correct words from the language list (english for now) as well.

kautukkundan: updated in 8f17aa9

The set mnemonic button won't show and an error message will pop-up if the words aren't compatible with BIP-39 standard.

Screenshot 2022-10-23 at 11 22 36 PM

jacque006 (approved):

PR #366, opened 2022-10-27
Add tests to decode error results

### LINKED ISSUE #365: Decode errors from action results
**As a** developer,
**I want** a client function that decodes the results of actions in an executed operation when they did not succeed
**So that** I can see the errors that occurred in failed actions

jacque006:
## What is this PR doing?
- ~Add `getOperationResults` client function to more easily parse operation results & errors.~ Done in earlier PR
- ~Add `typemoq` to clients to aid in mocking for unit tests.~ Removed
- Add test coverage for `OperationResults.ts` logic.
- Add `nyc` test coverage instrumentation to identify gaps in our client tests.

`cd ./contracts/clients && yarn test`

Resolves https://github.com/web3well/bls-wallet/issues/365

## Checklist

## Guidelines

## TODO
-  [x] Add coverage for remaining new code branches
-  [x] Add coverage for `decodeError` code branches
-  [x] ~Enforce client test coverage https://www.npmjs.com/package/nyc#coverage-thresholds~ Will be slightly more involved than originally expected, broke off into https://github.com/web3well/bls-wallet/issues/436


contracts/clients/package.json
```diff
@@ -33,6 +33,7 @@
     "mocha": "^10.0.0",
     "source-map-support": "^0.5.21",
     "ts-node": "^10.9.1",
+    "typemoq": "^2.1.0",
```
jacque006: If anyone knows of any other TypeScript friendly mocking libs that would be preferable, let me know.

blakecduncan (approved):
lgtm 👍

contracts/clients/src/OperationResults.ts
```diff
@@ -0,0 +1,110 @@
...
+ * @param transactionReceipt Transaction receipt from a VerificationGateway.processBundle transaction
+ * @returns An array of decoded operation results
+ */
+export const getOperationResults = (
+  transactionReceipt: ContractReceipt,
+): OperationResult[] => {
+  if (!transactionReceipt.events || !transactionReceipt.events.length) {
+    throw new Error(
+      `no events found in transaction ${transactionReceipt.transactionHash}`,
+    );
```
voltrevo: What's the idea behind throwing an error instead of just returning an empty array in these cases? Isn't that a valid result?

contracts/clients/test/OpertionResults.test.ts
```diff
@@ -0,0 +1,165 @@
...
+  describe("getOperationResults", () => {
+    it("fails if no events are in transaction", () => {
+      const hash = "0xabc123";
+      const txnReceiptMock = TypeMoq.Mock.ofType<ContractReceipt>();
+      txnReceiptMock.setup((r) => r.transactionHash).returns(() => hash);
+      txnReceiptMock.setup((r) => r.events).returns(() => undefined);
+
+      expect(() => getOperationResults(txnReceiptMock.object)).to.throw(
+        `no events found in transaction ${hash}`,
+      );
```
voltrevo: I'm concerned that `TypeMoq` is unnecessarily complex for our use case.

Why not just write:

```ts
const txnReceiptMoq = {
  transactionHash: hash,
  events: undefined,
} as ContractReceipt;
```

contracts/clients/test/OpertionResults.test.ts
```diff
@@ -0,0 +1,165 @@
...
+      eventMock.setup((e) => e.event).returns(() => "Other");
+
+      const hash = "0x123456";
+      const txnReceiptMock = TypeMoq.Mock.ofType<ContractReceipt>();
+      txnReceiptMock.setup((r) => r.transactionHash).returns(() => hash);
+      txnReceiptMock.setup((r) => r.events).returns(() => [eventMock.object]);
+
+      expect(() => getOperationResults(txnReceiptMock.object)).to.throw(
+        `no WalletOperationProcessed events found in transaction ${hash}`,
+      );
```
voltrevo: Similarly:

```ts
const txnReceiptMock = {
  transactionHash: hash,
  events: [
    { event: "Other" },
  ],
} as ContractReceipt;
```

voltrevo (changes requested):

voltrevo:
Also `OpertionResults.test.ts` -> `OperationResults.test.ts` (note the 'a')

contracts/clients/src/OperationResults.ts
jacque006: It was mainly to indicate to a user they may have passed in the wrong transaction. But I could likely do a better check on the method called and throw an error on that instead and return empty from here.

Checking the VG contract, it looks like you can technically submit a completely empty bundle, but I'm sure it would revert when it attempts to validate the empty sig.

contracts/clients/test/OpertionResults.test.ts
jacque006: I was trying to avoid `as` casts when possible as the underlying object technically fails proper typing. `typemoq` does add a bit of complexity but it will fully stub out the object, and could be useful in the future for testing more complex OO cases.

I will try without `typemoq`.

jacque006:
Should this be updated with https://github.com/web3well/bls-wallet/pull/391 ? Should we wait until `contract-updates` hits main?

contracts/clients/test/OpertionResults.test.ts
voltrevo: > but it will fully stub out the object

That doesn't seem possible... on this line:
```ts
const txnReceiptMock = TypeMoq.Mock.ofType<ContractReceipt>()
```
`ContractReceipt` only exists in the type space, so the actual object produced here must not have any information about `ContractReceipt`.

voltrevo:
@jacque006 I'd say it's easier to wait until `contract-updates` is merged in.

I'm a bit confused about how this branch adds `OperationResults.ts` but that file also exists in `contract-updates`. Is there some backstory here?

jacque006:
@voltrevo I just noticed that as well, forgot to update here. I cherry-picked the core code because I needed the action error decoding for testing the contracts, so I brought over the func.

We can use this branch/PR after `contract-updates` merge to address your comments and add tests.

jacque006:
Wait on https://github.com/web3well/bls-wallet/pull/377 before updating

contracts/clients/package.json
jacque006: Remove typemoq

contracts/clients/test/OpertionResults.test.ts
jacque006: The stub defaults to `undefined` return values. There also is a behavior where you can make it throw if you don't define something. But I think at this point is is overkill so I just removed.

contracts/clients/test/OpertionResults.test.ts
jacque006: Switched

contracts/clients/src/OperationResults.ts
jacque006: Thinking through this some more, we likely can't assume someone only will call `VG.processBundle` as they may go through a `BLSExpander` function instead. I will remove this check. I hope someone doesn't get too frustrated if they pass the wrong txn receipt in...

jacque006:
https://github.com/web3well/bls-wallet/actions/runs/3708751514/jobs/6286645375

image

contracts/clients/test/OperationResults.test.ts
```diff
@@ -0,0 +1,224 @@
...
+        message: `Unexpected error data: ${errorData}`,
+      });
+    });
+  });
+
+  describe("getOperationResults", () => {
+    it("fails if no events are in transaction", () => {
+      const txnReceipt = {
+        transactionHash: "0x111111",
+      } as unknown as ContractReceipt;
```
JohnGuilding: Seen the conversation about removing Typemoq. Not quite following on why we need to cast as unknown before ContractReceipt? I tried running the tests by only casting to `ContractReceipt` and they still pass.

contracts/clients/src/OperationResults.ts
```diff
@@ -107,19 +107,20 @@ const getError = (
...
-      `no events found in transaction ${txnReceipt.transactionHash}`,
-    );
-  }
-
-  const walletOpProcessedEvents = txnReceipt.events.filter(
+  const walletOpProcessedEvents = txnReceipt.events?.filter(
     (e) => e.event === "WalletOperationProcessed",
   );
-  if (!walletOpProcessedEvents.length) {
+  if (!walletOpProcessedEvents || !walletOpProcessedEvents.length) {
```
JohnGuilding: Possibly a bit less readable but nice and concise
```suggestion
  if (!walletOpProcessedEvents?.length) {
```

Also maybe one for retro, but wondered what your thoughts on negative vs positive logic were? I don't have a strong opinion but think it can help readability sometimes - https://www.obsidiansecurity.com/blog/why-to-stop-writing-negative-code/

contracts/clients/test/OperationResults.test.ts
```diff
@@ -0,0 +1,224 @@
...
+        errorSelectors.ActionError,
+        msg,
+      );
+      // There is probably a better way to extract
+      // the exact hex value of the unexpected action error data.
+      const actionErrDataStart = 202;
+      const actionErrorData = errorData.slice(
+        actionErrDataStart,
+        actionErrDataStart + 200,
+      );
```
JohnGuilding: Not quite following this test, what is happening here exactly?

contracts/clients/test/OperationResults.test.ts
```diff
@@ -0,0 +1,224 @@
...
+    });
+
+    it("handles exceptions when parsing error data", () => {
+      const encodedErrorMessage = utils.defaultAbiCoder.encode(
+        ["uint256"],
+        [0],
+      );
+      const errorData = `${
+        errorSelectors.ActionError
+      }${encodedErrorMessage.slice(2)}`;
```
JohnGuilding: Similar question here just for my understanding, what exactly is causing this error?

JohnGuilding (commented):
Lgtm, left a few questions and minor comments. Also wondering whether this logic needed a test? https://github.com/web3well/bls-wallet/blob/e1e985a82cb36983e47ff8a7dfdf4d39cb2a2357/contracts/clients/src/OperationResults.ts#L175-L178

contracts/clients/src/OperationResults.ts
jacque006: I think I somewhat agree with the difficulty of parsing negative logic. However, it is really helpful in reducing nesting in branching statements via early returns. For example:

```typescript
if (isCar) {
    if (isSedan) {
        return "Sedan";
    } else {
        return "Car";
    }
} else {
    return "Unknown";
}
```

vs.

```typescript
if (!isCar) {
    return "Unknown";
}
if (!isSedan) {
    return "Car";
}
return "Sedan";
```

This is a contrived example, but you can imagine how more complicated branching or error checking/validation could get difficult to grok.

contracts/clients/test/OperationResults.test.ts
jacque006: The intent of this code was to extract specific error hex data from a longer hex string. However, like most magic numbers written in the past I couldn't tell you why I had to use these specific values. Will attempt to update to a more human readable process.

contracts/clients/test/OperationResults.test.ts
jacque006: In this case, I am trying to create a malformed error response that triggers the unexpected error data branch. To do this, I add the `ActionError` selector (initial header bytes in data), and then a malformed encoded error message that triggers the `catch` in attempting to decode that error data.

contracts/clients/test/OperationResults.test.ts
jacque006: Removed unnecessary `unknown` cast.

contracts/clients/test/OperationResults.test.ts
jacque006: Switched from black magic to more understandable abi encoding.

.github/workflows/clients.yml
```diff
@@ -26,7 +26,15 @@ jobs:
...
-  test:
+  test-unit:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - uses: ./.github/actions/setup-contracts-clients
+    - run: yarn test
+
+  test-integration:
```
jacque006, line 37: @JohnGuilding I split this into separate steps to make it easier to identify unit vs. integration test failures. It will also make runs slightly faster, especially if we add more unit tests.

jacque006:
Ready for re-review

contracts/clients/src/OperationResults.ts
JohnGuilding: Yeah makes sense regarding nesting. I don't have a strong opinion on it but just wanted to get your thoughts :)

contracts/clients/test/OperationResults.test.ts
JohnGuilding: Thanks for clarifying both of these, makes sense now

JohnGuilding (approved):
Looks good!

JohnGuilding:
> Lgtm, left a few questions and minor comments. Also wondering whether this logic needed a test?
> 
> https://github.com/web3well/bls-wallet/blob/e1e985a82cb36983e47ff8a7dfdf4d39cb2a2357/contracts/clients/src/OperationResults.ts#L175-L178

Just to clarify @jacque006, this doesn't need a test?

jacque006:
> Just to clarify @jacque006, this doesn't need a test?

Missed this, kind of. This 'invariant' is actually run when the file is loaded into the JS runtime: https://github.com/web3well/bls-wallet/blob/9f4064eef9de499b4d959b6462180b5bd885d5f3/contracts/clients/src/OperationResults.ts#L6-L13
And you can see here those 3 calls being run in coverage file:
image

This means that if those `assert`s failed, it would probably just crash the test suite/code loading this in, which is what I believe @voltrevo wanted.

The one thing we are not testing here is what happens when the assert fails. Normally in coverage detection a branch statement like:
```typescript
if (somethingIsWrong) {
    throw new Error("you done goofed");
}
```
Would require you to run the function this is in twice, testing the true/false paths for this statement to get full branch coverage. However, since that selector check is an `assert`, the branching actually happens in https://github.com/web3well/bls-wallet/blob/main/contracts/clients/src/helpers/assert.ts . So from a testing coverage standpoint you would test the two paths in `assert` and things are fine as is.

We could export `calculateAndCheckSelector` and directly test it to make sure it throws when there is the mismatch, but I don't think there is a ton of value in that ATM. So going to leave as is.

jacque006:
Going to merge w/o @voltrevo 's approval, believe all their comments are addressed. Can follow up with additional work if other issues need to be addressed.

Issue #329, opened 2022-09-06
Fix placeholder (use real value) for tx hash in transactions table


voltrevo:
![Image](https://user-images.githubusercontent.com/9291586/188549557-b5fba5aa-1d45-4d7a-8599-c10a6a3e8520.png)

Fixing this should also fix the link so that the block explorer shows something relevant (probably the aggregate tx that the action was included in).

jacque006:
This should already be stored in the `BundleService` on the aggregator https://github.com/web3well/bls-wallet/blob/main/aggregator/src/app/BundleService.ts#L44

And can be added in to the returned data here: https://github.com/web3well/bls-wallet/blob/main/aggregator/src/app/BundleService.ts#L206-L211

PR #369, opened 2022-11-01
Return txhash from aggregator and display in Quill

blakecduncan:
## What is this PR doing?
- Return txhash from aggregator and display in Quill

- Execute a tx in Quill and view transaction hash in the transactions table

https://github.com/web3well/bls-wallet/issues/329

## Checklist

## Guidelines

### DIFF
--- a/aggregator/src/app/BundleService.ts
+++ b/aggregator/src/app/BundleService.ts
@@ -205,6 +205,8 @@ export default class BundleService {
 
     return {
       transactionIndex: receipt.transactionIndex,
+      transactionHash: receipt.transactionHash,
+      bundleHash: hash,
       blockHash: receipt.blockHash,
       blockNumber: receipt.blockNumber,
     };
--- a/extension/source/Home/Wallet/Wallets/TransactionsTable.tsx
+++ b/extension/source/Home/Wallet/Wallets/TransactionsTable.tsx
@@ -78,7 +78,9 @@ export const TransactionsTable: React.FunctionComponent<ITransactionsTable> = ({
   const quill = useQuill();
   const transactions = useCell(quill.cells.transactions);
   const data =
-    transactions?.outgoing.filter((t) => t.from === selectedAddress) || [];
+    transactions?.outgoing.filter(
+      (t) => t.from.toUpperCase() === selectedAddress.toUpperCase(),
+    ) || [];
 
   const columns = React.useMemo<ColumnDef<QuillTransaction>[]>(
     () => [
--- a/extension/source/Home/Wallet/Wallets/TransactionsTable.tsx
+++ b/extension/source/Home/Wallet/Wallets/TransactionsTable.tsx
@@ -66,7 +66,7 @@ const TransactionTableCell = (cell: { getValue: () => string }) => {
         window.open(`https://etherscan.io/tx/${cell.getValue()}`),
       )}
     >
-      {cell.getValue()}
+      {formatCompactAddress(cell.getValue())}
       <ArrowUpRight />
     </div>
   );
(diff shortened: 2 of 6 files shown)


extension/source/background/AggregatorController.ts
```diff
@@ -91,6 +91,11 @@ export default class AggregatorController {
         aggregatorUrl,
       };
 
+      await this.InternalRpc().updateTransactionBundleHash(
```
blakecduncan, line 94: Need to set the bundle hash on the tx so that when we later call get bundle receipt we can link the transaction with the receipt

extension/source/background/AggregatorController.ts
```diff
@@ -130,15 +135,23 @@ export default class AggregatorController {
...
+        this.InternalRpc().updateTransactionHashByBundleHash(
+          hash,
+          bundleReceipt.transactionHash,
+        );
+      }
+
       return (
         bundleReceipt && {
-          transactionHash: hash,
+          transactionHash: bundleReceipt.transactionHash,
```
blakecduncan, line 154: do we want to add a `bundleHash` value here?

aggregator/src/app/BundleService.ts
```diff
@@ -205,6 +205,8 @@ export default class BundleService {
 
     return {
       transactionIndex: receipt.transactionIndex,
+      transactionHash: receipt.transactionHash,
+      bundleHash: hash,
```
blakecduncan, line 209: My understanding is that the `transactionHash` is the hash of the transaction on chain.  I also wanted to return `bundleHash` because that is the hash we use to identify the bundle in the aggregator db.  Feel free to let me know if my understanding is wrong or if I should change this

extension/source/background/AggregatorController.ts
voltrevo: Maybe, but I wouldn't do it proactively since `eth_getTransactionReceipt` has a standard structure

aggregator/src/app/BundleService.ts
voltrevo: Sounds right. Notably, the 'transaction' of `transactionHash` is the `processBundle` call rather than the user's transaction (aka operation) inside.

extension/source/Home/Wallet/Wallets/TransactionsTable.tsx
```diff
@@ -66,7 +66,7 @@ const TransactionTableCell = (cell: { getValue: () => string }) => {
         window.open(`https://etherscan.io/tx/${cell.getValue()}`),
```
voltrevo, line 66: Followup: https://github.com/web3well/bls-wallet/issues/372

voltrevo (approved):

kautukkundan (changes requested):
Tested on Goerli - 
1. Doing single action tx from a dapp did not show up in the transactions table even though the entries were created in the localstorage ❌
2. Doing a multi action tx correctly populates the transactions table ✅
3. Sending ETH from within the extension correctly populates the transactions table ✅
4. Did not test recovery ❓

Screenshot 2022-11-03 at 12 42 18 PM


used the erc20 demo app for the purpose - 
https://joyful-semifreddo-3b0769.netlify.app/

Could you please verify (and fix if true)

blakecduncan:
> Tested on Goerli -
> 
> 1. Doing single action tx from a dapp did not show up in the transactions table even though the entries were created in the localstorage ❌
> 2. Doing a multi action tx correctly populates the transactions table ✅
> 3. Sending ETH from within the extension correctly populates the transactions table ✅
> 4. Did not test recovery ❓

@kautukkundan sorry I missed the single action.  That should be fixed now.  As Andrew mentioned in the discord channel, the filtering on the transaction table was case-sensitive.  I haven't dug in enough to know why sometimes the addresses have mixed case and other times they don't.  Maybe it has something to do with the network?  I'll take a deeper look but I don't think the mixed case is related to this PR.

For recovery, that code isn't merged into main yet so I haven't tested it.  However, I don't think the way we set it up it would show up in the transactions table.  Main reason being is that we are using a temporary wallet to execute the recovery user op.  So there's nowhere for us to visualize the tx.

kautukkundan (approved):
Gotcha, maybe something to look into in the future. 
could use checksum -
https://docs.ethers.io/v5/api/utils/address/#utils-getAddress

but LGTM for now 🚢

Issue #364, opened 2022-10-26
ABI Decode results from actions in an operation


jacque006:
**As a** developer,
**I want** a client function that decodes the results of actions in an executed operation
**So that** I can extract results data in a structured, human readable way

## Considerations
 - Results should be decodable into a TS/JS struct (maybe with ts-io validation?) so they can be easily usable in client consumers.

voltrevo:
@jacque006 Is this about decoding the return value of `verificationGateway.callStatic.processBundle(...)`?

If so, I feel like a guide might be better than supplying a new function - using that function could be more complicated than the code inside it.

The way to do this is to use `contract.interface.decodeFunctionResult` with the appropriate `contract`(s) used (which we don't have in the bundle, so they need to be passed in anyway).

Curiously, typechain doesn't provide accurate types for `decodeFunctionResult`, e.g.:

```ts
// MockERC20.d.ts
// ...
export interface Result extends ReadonlyArray<any> {
    readonly [key: string]: any;
}
// ...
  decodeFunctionResult(functionFragment: "allowance", data: BytesLike): Result;
  decodeFunctionResult(functionFragment: "approve", data: BytesLike): Result;
  decodeFunctionResult(functionFragment: "balanceOf", data: BytesLike): Result; // <--- Not BigNumber? :(
  decodeFunctionResult(functionFragment: "decimals", data: BytesLike): Result;
```

We could help out by providing a utility that fills in that gap, similar to what we're doing internally here:

```ts
type DecodeReturnType<
  Contract extends BaseContract,
  Method extends keyof Contract["callStatic"],
> = EnforceArray<AsyncReturnType<Contract["callStatic"][Method]>>;
```

- https://github.com/web3well/bls-wallet/blob/a8534c5/aggregator/src/app/EthereumService.ts#L62-L65

Preferably, it would be nice to contribute improved type generation for [typechain](https://github.com/dethcrypto/TypeChain). (In addition to the guide/functionality for decoding `processBundle`, since it would simplify that work.)

jacque006:
@voltrevo The idea was to do it off the event data, similarly to https://github.com/web3well/bls-wallet/pull/366/files#diff-674eb8bdeb284c750de53a656e2bf4afba3c0a81c9d9024dcab7617150c6557cR77 . That way people could get the results of the action function calls.

[...]

PR #377, opened 2022-11-04
Bundle error logging

### LINKED ISSUE #301: Add better failure messaging for bundles
As an aggregator operator (and maybe consumer), I would like to know why a bundle has not been submitted so I can more easily troubleshoot issues with my aggregator configuration or bundles I submit.

### Possible solution(s)
- Log out when a bundle fails to be included in the next submission and why it was not included.
- Return an error/failure status and the reason for the error/failure when querying `/bundleReceipt/:hash`.

blakecduncan:
https://loom.com/share/2a2039a17af04e23838561a08b0c6aaf

## What is this PR doing?
- Adding a submit error column to the DB
- Updating the bundle receipt API call to return a 202 if the bundle exists in the DB
- New response object for bundle receipt API call (note: this will break demos.  It's a simple fix but I want to make sure it is okay with everyone)

Additional work
- We would need to update examples and the extension because right now, they expect the `/bundleReceipt` API call to return a 404 when a bundle is pending.

Future improvements
- Add a status column to the bundles table so that we don't delete bundles when the are successful or fail (we are currently storing the successful bundles in memory)

You can run the zeroAddressError script to trigger a bundle error and log out a message

Resolves https://github.com/web3well/bls-wallet/issues/301

## Checklist

## Guidelines


aggregator/manualTests/zeroAddressError.ts
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env -S deno run --allow-net --allow-env --allow-read --allow-write --unstable
```
blakecduncan, line 1: This is just a script to trigger an error in a bundle for testing

aggregator/src/app/AggregationStrategy.ts
```diff
@@ -235,11 +235,16 @@ export default class AggregationStrategy {
       let success: boolean;
 
       if (bundleResult.success) {
-        const [operationResults] = bundleResult.returnValue;
+        const [operationStatuses, results] = bundleResult.returnValue;
```
blakecduncan: When there is a failure this results variable has some data in it.  I haven't been able to decode that value though.  I get an out of bounds error everytime I try.  If anyone knows if I can decode that let me know.  Or if it is even worth trying to decode the value.

See comment in the code below for an example of a value that is returned

aggregator/src/app/EthereumService.ts
```diff
@@ -302,6 +302,19 @@ export default class EthereumService {
...
 
+      try {
+        const result = attemptResult.value.events[0].args.results[0]; // For errors this is "Error(string)"
+        const errorArgBytesString = `0x${result.substring(10)}`; // remove methodId (4bytes after 0x)
+        const errorString = ethers.utils.defaultAbiCoder.decode(
+          ["string"],
+          errorArgBytesString,
+        )[0]; // decoded bytes is a string of the action index that errored.
+
+        console.error(errorString)
```
blakecduncan: I'm not sure if there's anything else we can do here other than just log out the message?  If you have thoughts let me know

aggregator/src/app/AggregationStrategy.ts
```diff
@@ -235,11 +235,16 @@ export default class AggregationStrategy {
...
+
+        // TODO: is it possible to decode this?
+        // Here is an example when I try to send more ETH than an account has:results =
+        // results = [ "0x4e487b710000000000000000000000000000000000000000000000000000000000000011" ]
```
jacque006: ~Would the work in https://github.com/web3well/bls-wallet/pull/366 help with this?~ See https://github.com/web3well/bls-wallet/pull/377/files#r1019631375 (comment above ^)

aggregator/src/app/AggregationStrategy.ts
jacque006: I wouldn't worry about decoding the results for this work as we have https://github.com/web3well/bls-wallet/issues/364 to address this. Would focus on bubbling up human readable errors from submitted bundles.

aggregator/src/app/EthereumService.ts
```diff
@@ -302,6 +302,21 @@ export default class EthereumService {
...
       const attemptResult = await attempt();
 
+      try {
+        if (attemptResult.value.events[0]?.args?.results[0]) {
+          const result = attemptResult.value.events[0].args.results[0]; // For errors this is "Error(string)"
+          const errorArgBytesString = `0x${result.substring(10)}`; // remove methodId (4bytes after 0x)
+          const errorString = ethers.utils.defaultAbiCoder.decode(
+            ["string"],
+            errorArgBytesString,
+          )[0]; // decoded bytes is a string of the action index that errored.
```
jacque006: If https://github.com/web3well/bls-wallet/pull/366 goes in before this, you can just use that lib method.

blakecduncan: Yeah that is a good call.  I'll probably do that

jacque006: Actually available right meow https://github.com/web3well/bls-wallet/blob/main/contracts/clients/README.md#verificationgateway

aggregator/src/app/AggregationStrategy.ts
```diff
@@ -444,3 +461,89 @@ export default class AggregationStrategy {
     return left;
   }
 }
+
+// The below code is temporary till the functions are exposed in the client module
```
blakecduncan: After this is merged and a new version is published, I will delete the below code and use the function from the client module

https://github.com/web3well/bls-wallet/pull/407

aggregator/src/app/AggregationStrategy.ts
```diff
@@ -231,22 +236,34 @@ export default class AggregationStrategy {
...
+            return;
+          }
+        });
+        errorReason = error[0] !== undefined
+          ? error[0]
+          : null;
+      }
 
-      return { success, fee };
+      return { success, fee, errorReason };
```
blakecduncan, line 272: Returning the error reason here so that we can link it to a failed row later when we check the index of the failed bundle

aggregator/src/app/BundleRouter.ts
```diff
@@ -24,6 +24,16 @@ export default function BundleRouter(bundleService: BundleService) {
   router.get(
     "bundleReceipt/:hash",
     async (ctx) => {
+      // check pending
```
blakecduncan: Return 202 error message if the bundle is in the DB.  Also, pass a submit error in the body

aggregator/src/app/AggregationStrategy.ts
```diff
@@ -231,22 +236,36 @@ export default class AggregationStrategy {
...
-
-      if (bundleResult.success) {
-        const [operationResults] = bundleResult.returnValue;
-
-        // We require that at least one operation succeeds, even though
-        // processBundle doesn't revert in this case.
-        success = operationResults.some((opSuccess) => opSuccess === true);
-      } else {
-        success = false;
+      const fee = after.returnValue[0].sub(before.returnValue[0]);
```
jacque006, line 241: Will this break if someone submits a bundle with 0 operations?

aggregator/src/app/AggregationStrategy.ts
```diff
@@ -231,22 +236,36 @@ export default class AggregationStrategy {
...
+      if (!bundleResult.success) {
+        return { success: false, fee, errorReason: null };
```
jacque006: Should the `errorReason` here be the decoded error results form the bundle operations? Or in the case of a txn revert the top level error message?

aggregator/src/app/AggregationStrategy.ts
```diff
@@ -231,22 +236,36 @@ export default class AggregationStrategy {
...
+      if (!success) {
+        const error = results.map((result: string[]) => {
+          try {
+            if (result[0]) {
+              return decodeError(result[0]);
+            }
+            return;
+          } catch (_err) {
+            return;
+          }
```
jacque006: Would be good to log out the error here. May even be worth adding and emitting as an `AppEvent`.

```suggestion
          } catch (err) {
            console.error(err);
            return;
          }
```

aggregator/src/app/AggregationStrategy.ts
```diff
@@ -231,22 +236,36 @@ export default class AggregationStrategy {
...
+        });
+        errorReason = error[0] !== undefined
+          ? error[0]
+          : null;
```
jacque006:
```suggestion
        const errorReason = error[0] ?? null;
```

aggregator/src/app/BundleRouter.ts
```diff
@@ -24,14 +24,29 @@ export default function BundleRouter(bundleService: BundleService) {
   router.get(
     "bundleReceipt/:hash",
     async (ctx) => {
+      const pendingBundle = await bundleService.lookupBundle(ctx.params.hash!);
+      if (pendingBundle) {
+        ctx.response.status = 202;
+        ctx.response.body = {
+          status: "pending",
+          submitError: pendingBundle.submitError,
```
jacque006: I believe after 5 tries the bundle will be in a permanent failed state. Is there a way to communicate that here? That may be for another issue to address.

jacque006 (commented):
Would be good to have an integration test case(s) for the `HTTP 202` path that ensures the bundle lifecycle & error decoding works as expected.

aggregator/src/app/AggregationStrategy.ts
blakecduncan: hmm I don't think so.  This logic isn't any different than it was before.  I just moved this line up so I could have an early return case to get rid of some of the nested if statements.

aggregator/src/app/AggregationStrategy.ts
blakecduncan: Yeah, that's a good question @jacque006.  I'm really not sure what to put here for the errorReason.  Whenever `bundleResult.success` is `false` there isn't anything returned for me to decode.

aggregator/src/app/BundleRouter.ts
blakecduncan: Looking at the `handleFailedRow()` function in the BundleService.ts, it looks like when a bundle gets into a permanently failed state, we delete the bundle from the table.  With this setup that would mean a permanently failed bundle would return a 404 error.

I think a solution is probably to add a status column in the DB that could have 'pending', 'success', or 'failed'.  This would also fix the issue of storing successful bundles in memory.  I was torn whether to do that as part of this PR or keep it separate.

blakecduncan: I created this issue as a follow up to this comment https://github.com/web3well/bls-wallet/issues/408

blakecduncan:
> Would be good to have an integration test case(s) for the `HTTP 202` path that ensures the bundle lifecycle & error decoding works as expected.

Yeah good call, I'll write some tests for this 👍

aggregator/src/app/AggregationStrategy.ts
jacque006: @blakecduncan You can do that now https://github.com/web3well/bls-wallet/pull/407 . You can use https://github.com/web3well/bls-wallet/blob/main/docs/local_development.md#aggregator-1 as a guide, but feel free to ping me if you need help.

aggregator/src/app/AggregationStrategy.ts
jacque006: Looking through `AggregatorUtilities.sol` and the `EthereumService.ts` i think this makes a bit more sense now. Probably a larger/different discussion on how we are simulating actions.

Until then, I would put in something like `unknown` because that is the extent of what we know.

```suggestion
        return { success: false, fee, errorReason: "unknown" };
```

aggregator/src/app/BundleRouter.ts
jacque006: Follow up seems appropriate, thanks :+1:

voltrevo:
> [Idk how I'm missing these lint errors](https://github.com/web3well/bls-wallet/pull/377/commits/e243212e67e45606a8b00b0cb5c97fc8af437875)

😄. How's your IDE set-up? Are the linting errors showing? Is autofix-on-save working?

aggregator/src/app/AggregationStrategy.ts
```diff
@@ -176,6 +178,10 @@ export default class AggregationStrategy {
 
     if (firstFailureIndex !== nil) {
       const failedRow = includedRows[firstFailureIndex];
+      const errorReason = fees[firstFailureIndex].errorReason;
+      if (errorReason) {
+        failedRow.submitError = errorReason.message || '';
```
voltrevo: Why default to an empty string instead of `undefined`/`nil`?

Also nitpick: Prefer `??` over `||`. (Unless there's a reason for `||` I'm not seeing here?)

aggregator/src/app/AggregationStrategy.ts
```diff
@@ -207,6 +213,7 @@ export default class AggregationStrategy {
   async #measureFees(bundles: Bundle[]): Promise<{
     success: boolean;
     fee: BigNumber;
+    errorReason: OperationResultError | null;
```
voltrevo: We don't really use `null` internally - only when something external needs it. This should probably be documented somewhere. Can you make this `nil`?

aggregator/src/app/AggregationStrategy.ts
```diff
@@ -231,22 +238,38 @@ export default class AggregationStrategy {
...
+        const errorReason: OperationResultError = {
+          message: "Unknown error reason",
+        };
+        return { success: false, fee, errorReason };
       }
 
-      const fee = after.returnValue[0].sub(before.returnValue[0]);
+      const [operationStatuses, results] = bundleResult.returnValue;
+
+      let errorReason: OperationResultError | null = null;
```
voltrevo: `null` -> `nil`

aggregator/src/app/AggregationStrategy.ts
```diff
@@ -231,22 +238,38 @@ export default class AggregationStrategy {
...
+            if (result[0]) {
+              return decodeError(result[0]);
+            }
+            return;
+          } catch (err) {
+            console.error(err);
+            return;
+          }
+        });
+        errorReason = error[0] ?? null;
```
voltrevo: Remove `?? null`

aggregator/src/app/BundleRouter.ts
```diff
@@ -24,14 +24,29 @@ export default function BundleRouter(bundleService: BundleService) {
   router.get(
     "bundleReceipt/:hash",
     async (ctx) => {
+      const pendingBundle = await bundleService.lookupBundle(ctx.params.hash!);
+      if (pendingBundle) {
+        ctx.response.status = 202;
```
voltrevo: I don't think 202 is appropriate here. It's the *bundle* that's been accepted, not the receipt. Also 'accepted' is supposed to refer to the current request, not the resource you're asking about. [More info](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202).

(Possibly 202 is appropriate for the original POST to /bundle, but I'm not actually a fan of worrying about these HTTP nuances.)

On a practical note, if we keep 404 here we can still also provide the `submitError`, and this way it wouldn't be a breaking change. Wdyt?

aggregator/src/app/BundleRouter.ts
```diff
@@ -24,14 +24,29 @@ export default function BundleRouter(bundleService: BundleService) {
...
 
       if (receipt === nil) {
         ctx.response.status = 404;
         return;
       }
 
-      ctx.response.body = receipt;
+      ctx.response.body = {
+        status: "done",
+        submitError: null,
```
voltrevo: `null`

aggregator/src/app/BundleService.ts
```diff
@@ -184,6 +184,7 @@ export default class BundleService {
         bundle,
         eligibleAfter: await this.ethereumService.BlockNumber(),
         nextEligibilityDelay: BigNumber.from(1),
+        submitError: ''
```
voltrevo: Why is there an empty string here? Can we just remove the field?

aggregator/src/app/BundleService.ts
```diff
@@ -200,6 +201,11 @@ export default class BundleService {
     });
   }
 
+  async lookupBundle(hash: string) {
+    const bundleArray = await this.bundleTable.findBundle(hash);
+    return bundleArray[0] ? bundleArray[0] : nil;
```
voltrevo: Why does `findBundle` return an array? Wouldn't `Bundle | nil` be more appropriate? That way you'd also not need to transform the result here.

aggregator/src/app/BundleTable.ts
```diff
@@ -52,6 +54,7 @@ const tableOptions: TableOptions = {
   id: { type: DataType.Serial, constraint: Constraint.PrimaryKey },
   hash: { type: DataType.VarChar },
   bundle: { type: DataType.VarChar },
+  submitError: { type: DataType.VarChar },
```
voltrevo:
```suggestion
  submitError: { type: DataType.VarChar, nullable: true },
```

aggregator/src/app/BundleTable.ts
```diff
@@ -27,6 +27,7 @@ type RawRow = {
   bundle: string;
   eligibleAfter: string;
   nextEligibilityDelay: string;
+  submitError: string;
```
voltrevo:
```suggestion
  submitError: string | null;
```

aggregator/src/app/BundleTable.ts
```diff
@@ -35,6 +36,7 @@ type Row = {
   bundle: Bundle;
   eligibleAfter: BigNumber;
   nextEligibilityDelay: BigNumber;
+  submitError: string;
```
voltrevo:
```suggestion
  submitError?: string;
```

Note that `RawRow` reflects what we get from the database, and therefore needs to use `null` (I think). This means adding the appropriate translation in `toRawRow`, `fromRawRow` (which TypeScript will point out).

aggregator/test/AggregationStrategy.test.ts
```diff
@@ -1,6 +1,6 @@
 import AggregationStrategy from "../src/app/AggregationStrategy.ts";
 import { BundleRow } from "../src/app/BundleTable.ts";
-import { assertEquals, BigNumber } from "./deps.ts";
+import {assertEquals, BigNumber, ethers} from "./deps.ts";
```
voltrevo: Nitpick: Formatting. (Do you have autofix on save enabled?)

aggregator/test/AggregationStrategy.test.ts
```diff
@@ -98,6 +98,7 @@ Fixture.test("includes bundle in aggregation when estimated fee is provided", as
   const bundleRow: BundleRow = {
     id: 0,
     hash: "0x0",
+    submitError: "",
```
voltrevo: (See previous comment about use of empty string)

aggregator/test/AggregationStrategy.test.ts
```diff
@@ -111,3 +112,61 @@ Fixture.test("includes bundle in aggregation when estimated fee is provided", as
...
+          ],
+        ),
+      }
+    ],
+  });
+
+  const bundleRow: BundleRow = {
+    id: 0,
+    hash: "0x0",
+    submitError: "",
```
voltrevo: Empty string

aggregator/test/BundleTable.test.ts
```diff
@@ -49,6 +49,7 @@ const sampleRows: BundleRow[] = [
       ],
       signature: ["0x00", "0x00"],
     },
+    submitError: "",
```
voltrevo: Empty string

contracts/clients/src/Aggregator.ts
```diff
@@ -34,9 +34,17 @@ export type EstimateFeeResponse = {
...
+  transactionHash: string;
+  bundleHash: string;
   blockHash: string;
-  blockNumber: string;
+  blockNumber: number;
+};
+
+export type BundleReceiptResponse = {
+  status: string;
+  submitError: string | null;
```
voltrevo: null

contracts/clients/src/Aggregator.ts
```diff
@@ -34,9 +34,17 @@ export type EstimateFeeResponse = {
...
+  receipt: BundleReceipt | null;
```
voltrevo: null

extension/source/background/AggregatorController.ts
```diff
@@ -126,21 +123,10 @@ export default class AggregatorController {
...
-            transactionHash: io.string,
-            blockHash: io.string,
-            blockNumber: io.number,
-          }),
-        ),
-      );
+      const bundleReceiptResponse = await aggregator.lookupReceipt(hash);
+      // If receipt doesn't exist, set to undefined to be consistent with
+      // the return type
+      const bundleReceipt = bundleReceiptResponse?.receipt || undefined;
```
voltrevo: `|| undefined` is unnecessary (`obj?.field` produces `undefined` when field is not there, even when `obj` is `null`).

Also prefer `??` over `||`.

voltrevo (changes requested):

aggregator/src/app/AggregationStrategy.ts
blakecduncan: @voltrevo Yeah that makes sense I'll switch it.  I was doing this because I didn't make the column nullable in the DB, so it was setting it by default to an empty string.  I updated that per your comment below and was able to make this more consistent with the undefined/nil

extension/source/background/AggregatorController.ts
blakecduncan: Yeah I was getting a typescript error the `BundleReceipt | undefined` does not equal `BundleReceipt | undefined | null`.  I think with the changes you recommended above I can fix this.

aggregator/src/app/BundleRouter.ts
blakecduncan: Yeah, I get what you're saying.  I do like the idea of not doing a breaking change, as there are a good amount of things that have to be updated at this point.

However, I think the `/bundleReceipt` API call is often used as a status check for a bundle.  So my question would be, what should we expect to be returned from this API call?  With the approach you recommended, we would keep a 200 status as is (i.e. returns the bundle receipt) and pass the submit error in a 404 response body.  That would be fine by me - I think the only thing we aren't communicating explicitly is the bundle status. For example, we could communicate a pending or a failed state (ex: even if it failed once, we may still retry it).  Maybe we don't care about that though or it should be a different API endpoint.

@voltrevo What do you think?  I'd be happy to hop on a call tomorrow when I get online if you are still around.  Also, @jacque006  not sure if you have any thoughts?

voltrevo: I'm not sure we're on the same page here. Definitely sounds like a call is in order.

> With the approach you recommended, we would keep a 200 status as is (i.e. returns the bundle receipt) and pass the submit error in a 404 response body.

Did you mean to say 'keep a 200 status'? It seems wrong, and also contradicts the second half of this sentence. At least my reading of it.

To rephrase my suggestion, I think if the bundle is pending, then `/bundleReceipt` should 404, since there isn't a bundle receipt, but we can also put the submitError in the body of this response. Meaning: "No, there's no receipt (404), and here's why (submitError)".

aggregator/src/app/BundleTable.ts
```diff
@@ -54,20 +55,21 @@ const tableOptions: TableOptions = {
...
-  submitError: { type: DataType.VarChar },
+  submitError: { type: DataType.VarChar, nullable: true },
   eligibleAfter: { type: DataType.VarChar },
   nextEligibilityDelay: { type: DataType.VarChar },
 };
 
 function fromRawRow(rawRow: RawRow): Row {
   const parseResult = parseBundleDto(JSON.parse(rawRow.bundle));
-
+  console.log('blake..', parseResult);
```
voltrevo: debug log left here

aggregator/src/app/BundleRouter.ts
blakecduncan: Ahh my response was poorly worded.  We are on the same page.  When I said keep the 200 status, I really meant get rid of my logic to add the 202.  You worded it better than I did - that is what I meant to say.

jacque006: Yep, we can discuss offline. In general, I have seen services go one of two ways:

- Use a variety/specific HTTP codes to communicate the state of a resource (200/201/202/403/404/418)
- Use a limited number of status codes (200/403/404) and use a `status` field or similar to communicate the state of the resource (pending/failed/succeeded). Some Google Cloud APIs go this route

I think either is fine as long as we are consistent with that pattern on our endpoints and document the possible responses in code. Hopefully most consumers use our clients which can abstract a lot of that away.

jacque006 (approved):
:+1: from me. :shipit: once @voltrevo is satisfied

voltrevo (approved):

PR #402, opened 2022-11-16
Bls provider/signer send transaction

### LINKED ISSUE #373: BlsProvider Transaction Methods
We are adding a BlsProvider to `bls-wallet-clients`. This issue is concerned with ensuring the Transaction Methods work with the BlsProvider. This will include overriding methods with BLS logic if needed and testing the other methods work as expected.

Some of the operations include sending BLS transactions, getting a transaction via a hash, getting a transaction receipt via a hash, and estimating the amount of gas a transaction would consume.

[Transactions Methods](https://docs.ethers.io/v5/api/providers/provider/#Provider--transaction-methods) to test/override:

```ts
// Returns the result of executing the transaction, using call. A call does not require any ether,
// but cannot change any state. This is useful for calling getters on Contracts.
provider.call( transaction [ , blockTag = latest ] ) ⇒ Promise< string< DataHexString > >

// Returns an estimate of the amount of gas that would be required to submit transaction to the network.
provider.estimateGas( transaction ) ⇒ Promise< BigNumber >

// Returns the transaction with hash or null if the transaction is unknown.
provider.getTransaction( hash ) ⇒ Promise< TransactionResponse >

// Returns the transaction receipt for hash or null if the transaction has not been mined.
provider.getTransactionReceipt( hash ) ⇒ Promise< TransactionReceipt >

// Submits transaction to the network to be mined. The transaction must be signed, and be valid
// (i.e. the nonce is correct and the account has sufficient balance to pay for the transaction).
provider.sendTransaction( transaction ) ⇒ Promise< TransactionResponse >

// Returns a Promise which will not resolve until transactionHash is mined.
provider.waitForTransaction( hash [ , confirms = 1 [ , timeout ] ] ) ⇒ Promise< TxReceipt >
```

### LINKED ISSUE #378: BlsSigner Signing Methods
We are adding a BlsSigner to `bls-wallet-clients`. This issue is concerned with ensuring the Signing Methods work with the BlsSigner. This will include overriding methods with BLS logic if needed and testing the other methods work as expected.

Some of the operations include signing messages, signing transactions and sending transactions.

[Signing Methods](https://docs.ethers.io/v5/api/signer/#Signer--signing-methods) to test/override:

```ts
// This returns a Promise which resolves to the Raw Signature of message.
signer.signMessage( message ) ⇒ Promise< string< RawSignature > >

// Returns a Promise which resolves to the signed transaction of the transactionRequest. This 
// method does not populate any missing fields.
signer.signTransaction( transactionRequest ) ⇒ Promise< string< DataHexString > >

// This method populates the transactionRequest with missing fields, using populateTransaction and
// returns a Promise which resolves to the transaction.
signer.sendTransaction( transactionRequest ) ⇒ Promise< TransactionResponse >

// Signs the typed data value with types data structure for domain using the EIP-712 specification.
// Note: This is still an experimental feature. If using it, please specify the exact version of ethers 
// you are using (e.g. spcify "5.0.18", not "^5.0.18") as the method name will be renamed from 
// _signTypedData to signTypedData once it has been used in the field a bit.
signer._signTypedData( domain , types , value ) ⇒ Promise< string< RawSignature > >
```

Additional methods not included under the [Signing](https://docs.ethers.io/v5/api/signer/#Signer--signing-methods) section of the docs but are related - see where they came from [here](https://docs.ethers.io/v5/api/signer/#Signer--subclassing).

```ts
// This should return a copy of the transactionRequest, with any properties needed by 
// call, estimateGas and populateTransaction (which is used by sendTransaction). It should also 
// throw an error if any unknown key is specified.
signer.checkTransaction( transactionRequest ) ⇒ TransactionRequest


// This should return a copy of transactionRequest, follow the same procedure as 
// checkTransaction and fill in any properties required for sending a transaction. The result should 
// have all promises resolved; if needed the resolveProperties utility function can be used for this.
signer.populateTransaction( transactionRequest ) ⇒ Promise< TransactionRequest >
```

JohnGuilding:
## What is this PR doing?

This PR adds functionality to send transactions using the BlsProvider & BlsSigner, among other methods. Specifically, this PR constitutes the work required for https://github.com/web3well/bls-wallet/issues/373 and https://github.com/web3well/bls-wallet/issues/378.

The provider and signer are meant to extend the `JsonRpcProvider` & `JsonRpcSigner` classes. You'll notice the signer extends from `Signer` instead of `JsonRpcSigner`. This is because `JsonRpcSigner` has a [constructor guard](https://github.com/ethers-io/ethers.js/blob/c80fcddf50a9023486e9f9acb1848aba4c19f7b6/packages/providers/src.ts/json-rpc-provider.ts#L166-L168) which means you can't instantiate a new instance outside of the `getSigner()` method on `JsonRpcProvider` due to strict equality. Some of the logic from the `BlsSigner` constructor is taken from the `JsonRpcSigner` constructor.

### Provider methods
`call()` - not overridden but tested
`estimateGas()` - overridden & tested
`getTransaction()` - not overridden but tested
`getTransactionReceipt()` - overridden & tested
`sendTransaction()` - overridden & tested
`waitForTransaction()` - overridden & tested

**Additional provider methods added:**
`getSigner()` - overridden and tested. Needed to add this now in order to access the BlsSigner.

### Signer methods
`signMessage()` - overridden & tested
`signTransaction()` - overridden & tested
`sendTransaction()` - overridden and tested
`_signTypedData()` - not implemented as seems to be an ["experimental feature"](https://docs.ethers.io/v5/api/signer/#Signer-signTypedData) from ethers
`checkTransaction()` - not overridden but tested
`populateTransaction()` - not overridden but tested

**Additional signer methods added**:
`initializeWallet()` - tested. Sets a singleton instance of BlsWalletWrapper
`getAddress()` - implemented and tested. Needed access to the account address so added this function.
`constructTransactionResponse()` - tested. Helper function to construct an ethers transaction response type. This is needed to meet the return type of certain overridden methods
`unlock()` - implemented but not tested. This method was required to match the JsonRpcSigner interface. Will look into testing/exploring this in a future PR

There were some additional methods required in order to match the JsonRpcSigner interface. These were not implemented as they would be covered by other Github issues. These include `connect`, `_signTypedData`, `connectUnchecked`, `sendUncheckedTransaction`, `legacySignMessage`.

- Start a local hardhat node and local aggregator, then run `yarn clients-tests` in the `./contracts` directory.
- Also run the clients tests from within the `./clients` directory to run the tests for the polling helper.

https://github.com/web3well/bls-wallet/issues/373 and https://github.com/web3well/bls-wallet/issues/378. 

## Checklist

## Guidelines


contracts/clients/src/Aggregator.ts
```diff
@@ -1,3 +1,5 @@
+import { BigNumber, ContractReceipt } from "ethers";
+import { Log } from "@ethersproject/abstract-provider";
```
kautukkundan: install a new module node-fetch
`yarn add node-fetch@2`
to make fetch compatible outside browser environment

then import in file
`import fetch from "node-fetch";`

make sure to use @2 else it other version has some other issues, this version worked.

PS - I haven't tested in browser, unsure if this module works there as well. Personally I prefer using axios everywhere but I think this should be fine for the time being.

contracts/clients/src/BlsProvider.ts
```diff
@@ -1,3 +1,133 @@
...
+    if (bundleReceipt === undefined) {
+      throw new Error("Could not find bundle receipt");
+    }
+
+    return {
+      to: "0x",
+      from: "0x",
+      contractAddress: "0x",
+      transactionIndex: parseInt(bundleReceipt.transactionIndex),
+      gasUsed: parseEther("0x"),
```
kautukkundan: syntax error. Parse ether takes integer in string form.

```diff
- parseEther("0x")
+ parseEther("0")
```

Also in a couple of more places.

contracts/clients/src/BlsSigner.ts
```diff
@@ -1,3 +1,202 @@
...
+
+  async sendTransaction(
+    transaction: Deferrable<TransactionRequest>,
+  ): Promise<TransactionResponse> {
+    if (!this.wallet) {
+      throw new Error("Call this.init() to initialize the wallet.");
+    }
+    try {
+      // TODO: dynamically add this
+      const verificationGateway = "0x3C17E9cF70B774bCf32C66C8aB83D19661Fc27E2";
```
kautukkundan: Once latest main is merged, VG moves to "0x689A095B4507Bfa302eef8551F90fB322B3451c6" 

also hardcoded in a couple of more places. Ok to dynamically add later. All the config is stored in contracts/networks/<network_name>.json

contracts/test/BlsSigner.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    aggregatorUrl = "http://localhost:3000";
+    rpcUrl = "http://localhost:8545";
+    const network = {
+      name: "localhost",
+      chainId: 0x7a69,
+    };
+
+    // FIXME: Unsure on how to manage the private key!
+    const privateKey = Wallet.createRandom().privateKey;
+    const verificationGateway = "0x3C17E9cF70B774bCf32C66C8aB83D19661Fc27E2";
```
kautukkundan: same as prev. VG will be moved to a new address.

contracts/test/BlsSigner.test.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    blsProvider = new BlsProvider(aggregatorUrl, rpcUrl, network);
+    blsSigner = blsProvider.getSigner();
+    await blsSigner.initWallet(privateKey, verificationGateway, blsProvider);
+
+    regularProvider = new JsonRpcProvider("http://localhost:8545");
+    regularSigner = regularProvider.getSigner();
```
kautukkundan: The main issue was that the newly created bls wallet did not have any funds. That's why the bundle, though created, fails in simulation and a submission attempt is never made.

The aggregator will be more verbose and will be able to log this once #377 is merged.    

add this line before closing the beforeEach hook.

```ts
const fundedWallet = new ethers.Wallet(
      "0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d",
      regularProvider,
    );

    await fundedWallet.sendTransaction({
      to: await blsSigner.getAddress(),
      value: parseEther("1000"),
    });
```

Also as we discussed, the signer that we get by "getSigner" method is from the runtime instance of hardhat, whereas the wallet is created in the local dev node. That's why funding from the signer won't work. One way would be to run the tests with a network flag `--network gethDev` but imo it's easier to just create a wallet instance for the same network considering we use JsonRpcProvider anyway. The above private key is from one of the default hardhat accounts

kautukkundan (commented):
1. Take a pull from main branch before making further progress. 
2. run `./setup.ts` from root
3. While running aggregator add `-r` flag to refresh to the latest client code

contracts/clients/src/Aggregator.ts
kautukkundan: CC - @jacque006 for using "fetch" outside browser. I think it'll be built in by default inside node from v17 onwards.

blakecduncan: Do we need to do something like `global.fetch = fetch` so that imported packages (like bls module) has access to fetch?  I had to do that in some scripts I've written

JohnGuilding: Error in question was `Reference error: fetch is not defined`. Another workaround long-term might be to use node v18 which has the fetch API implemented by default.

https://stackoverflow.com/questions/48433783/referenceerror-fetch-is-not-defined

contracts/clients/src/BlsSigner.ts
JohnGuilding: Updated!

I am right in saying that we shouldn't reference a relative file path outside of `contracts/clients` due to the clients directory being an npm module that is isolated from the rest of our code?

e.g. `contracts/networks/<network_name>.json` would be unaccessible for this reason?

contracts/test/BlsSigner.test.ts
JohnGuilding: Added and tests pass 🎉

contracts/clients/src/Aggregator.ts
jacque006: We should support node versions without fetch until v16 rotates out of LTS support. I would recommend at some init point injecting `node-fetch` if it does not exist, esp if some other package is expecting it.

```
import fetch from "node-fetch";

if (!global.fetch) {
    global.fetch = fetch;
}
```

JohnGuilding:
**These questions and TODOs have now been resolved in a call with Kautuk**

**High level questions**:
1. How closely should we override and test every method? Original aim was to override every method that needed overriding and then testing the end result. 4337 provider and signer is not that many lines of code, and does not seem to override/test everything.
2. Are we happy with tests in this directory and set up like they are? Or should they follow the setup used in our other contracts tests. **Note**: we have to run the aggregator to run the provider/signer tests, can we avoid this?
3. Are we happy with error handling? Do we have a specific pattern to follow when handling errors?

**Code TODOs (_Can find these in their respective files and can be searched for. E.g. search for "TODO: 1" to find TODO 1._)**

`BlsProvider`
 **TODO: 1.** When constructing an Action, should we be stating something is nullable then telling the compiler that a nested property is not? e.g. use of optional chaining (`?.`) followed by non-null assertion operator (`!`) on the same object.

 **TODO: 2.** How do we fill out the rest of the transaction receipt? Could we return additional values from the `ethers.ContractReceipt` in the `BundleService`?

`BlsSigner`
 **TODO: 1.** Are we happy with the init function as it breaks the ethers pattern slightly (Although init pattern is used in the 4337 provider)? 
             - Allows us to have singleton instance of a `BlsWalletWrapper`. Would have to pass the private key in multiple places if we didn't have this, or store the private key on the class. At which point, `BlsSigner` becomes more similar to the ethers `Wallet` class.
 
 **TODO: 2.** When constructing an Action, should we be stating something is nullable then telling the compiler that a nested property is not? e.g. use of optional chaining (`?.`) followed by non-null assertion operator (`!`) on the same object.

 **TODO: 3.** `JsonRpcSigner` does not implement the `signBlsTransaction` method. Issue is we could override the `provider.sendTransaction` method  by updating the parameter types, but we couldn't override `signer.signTransaction` as we would have to update the return type which typescript doesn't seem to allow? End result is we had to create separate methods. Is that ok?

contracts/clients/src/BlsSigner.ts
JohnGuilding: Verification gateway address is currently passed into the constructor to resolve this issue. Plan in a future PR is to add an static method to initialise the class with default config values, similar to `BLSWalletWrapper`.

contracts/package.json
```diff
@@ -12,6 +12,7 @@
     "check-ts": "tsc --noEmit",
     "lint": "eslint . --ext .ts",
     "test": "hardhat test",
+    "test-clients": "hardhat test ./clients-tests/*",
```
JohnGuilding, line 15: Not 100% sure this new directory is the best home for these tests so open to suggestions. Didn't put them in `./clients` as they require hardhat, and the client module isn't a hardhat project.

Would also like to get thoughts on integrating these as automated tests, but unsure on how to do that.

jacque006: Given the dependency the clients have with the contracts, we probably should just setup integration style tests for them in addition to the current unit tests.

For automation, we could split the client GH action [test](https://github.com/web3well/bls-wallet/blob/main/.github/workflows/clients.yml#L26-L32) step into two or just include integration tests in it. See also the gas measurement PR where I [added a local contract deploy action](https://github.com/web3well/bls-wallet/pull/341/files#diff-21998d19187400c87c1b4d6bb4ccbe519f3c9f5fcff783d76f392d7640449de4) that could be used here as well.

contracts/clients-tests/BlsProvider.test.ts
```diff
@@ -0,0 +1,535 @@
...
+      .to.equal(transactionResponse.confirmations);
+    expect(transactionReceipt)
+      .to.have.property("cumulativeGasUsed")
+      .to.equal(parseEther("0"));
+    expect(transactionReceipt)
+      .to.have.property("effectiveGasPrice")
+      .to.equal(parseEther("0"));
+    expect(transactionReceipt).to.have.property("byzantium").to.equal(false);
+    expect(transactionReceipt).to.have.property("type").to.equal(2);
+  });
```
jacque006:
```suggestion
    // Use .to.deep.includes if this is not full receipt object.
    expect(transactionReceipt).to.deep.equal({
      to: "0x",
      from: "0x",
      //... etc.
    );
  });
```
Should be able to the same throughout tests.

contracts/clients/src/helpers/poll.ts
```diff
@@ -0,0 +1,23 @@
...
+    await wait(milliseconds);
+    result = await fn();
+    retryCount++;
+  }
+  return result;
+}
+
+function wait(milliseconds = 1000) {
+  return new Promise((resolve) => {
+    console.log(`Polling every ${milliseconds} milliseconds...`);
```
jacque006: This will print out in browsers/backends with no way for consumers to turn off. I would just remove. If you really need this for debugging, consider using something like https://www.npmjs.com/package/debug
```suggestion
```

contracts/clients-tests/BlsProvider.test.ts
```diff
@@ -0,0 +1,535 @@
...
+    signers = await hardhatEthers.getSigners();
+
+    aggregatorUrl = "http://localhost:3000";
+    verificationGateway = networkConfig.addresses.verificationGateway;
+    rpcUrl = "http://localhost:8545";
+    network = {
+      name: "localhost",
+      chainId: 0x7a69,
+    };
+    privateKey = Wallet.createRandom().privateKey;
```
jacque006: It is generally not a good idea to use purely random values in tests as it can make it hard to reproduce failures with the random value. I would just hard code a random private key here. Could also just use one of the well known Hardhat ones.

contracts/clients-tests/BlsProvider.test.ts
```diff
@@ -0,0 +1,535 @@
...
+    });
+  });
+
+  it("should return a valid signer", async () => {
+    // Arrange & Act
+    const blsSigner = blsProvider.getSigner();
+
+    // Assert
+    /* eslint-disable-next-line no-unused-expressions */
+    expect(blsSigner._isSigner).to.be.true;
```
jacque006: I would just disable this rule in test files. See https://stackoverflow.com/a/37777342

```suggestion
    expect(blsSigner._isSigner).to.be.true;
```

contracts/clients/src/BlsProvider.ts
```diff
@@ -0,0 +1,155 @@
...
+    }
+  }
+
+  override async sendTransaction(
+    signedTransaction: string | Promise<string>,
+  ): Promise<ethers.providers.TransactionResponse> {
+    throw new Error(
+      "sendTransaction() is not implemented. Call 'sendBlsTransaction()' instead.",
+    );
+  }
```
jacque006: Much like we do in the extension, we should parse/decode the signed txn and convert it to a bundle internally.

contracts/clients/src/BlsProvider.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  override async waitForTransaction(
+    transactionHash: string,
+    confirmations?: number,
+    retries?: number,
+  ): Promise<ethers.providers.TransactionReceipt> {
+    return this._getTransactionReceipt(
+      transactionHash,
+      confirmations == null ? 1 : confirmations,
+      retries == null ? 10 : retries,
```
jacque006: Does `null` have special meaning here? Otherwise:
```suggestion
      confirmations ?? 1,
      retries ?? 10,
```

contracts/clients/src/BlsProvider.ts
```diff
@@ -0,0 +1,155 @@
...
+    const bundleExists = (result: BundleReceipt) => !result;
+
+    const bundleReceipt = await poll(
+      getBundleReceipt,
+      bundleExists,
+      retries,
+      2000,
+    );
+
+    if (bundleReceipt === undefined) {
```
jacque006:
```suggestion
    if (!bundleReceipt) {
```

contracts/clients/src/BlsSigner.ts
```diff
@@ -0,0 +1,210 @@
...
+
+export default class BlsSigner extends Signer {
+  override readonly provider: BlsProvider;
+  readonly verificationGatewayAddress!: string;
+  wallet!: BlsWalletWrapper;
+  _index: number;
+  _address: string;
+
+  constructor(
+    constructorGuard: any,
```
jacque006: Prefer better typing over `any`.
```suggestion
    constructorGuard: Record<string, unknown>,
```

contracts/clients/src/BlsSigner.ts
```diff
@@ -0,0 +1,210 @@
...
+    this.provider = provider;
+    this.verificationGatewayAddress = this.provider.verificationGatewayAddress;
+
+    if (constructorGuard !== _constructorGuard) {
+      throw new Error(
+        "do not call the BlsSigner constructor directly; use provider.getSigner.",
+      );
+    }
+
+    if (addressOrIndex == null) {
```
```suggestion
    if (addressOrIndex === null || anddressOrIndex === undefined) {
```

contracts/clients/src/BlsSigner.ts
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  #verifyInit = () => {
+    if (!this.wallet || !this.verificationGatewayAddress) {
+      throw new Error(
+        "To perform this operation, ensure you have instantiated a BlsSigner and have called this.init() to initialize the wallet.",
+      );
+    }
+  };
+}
```
jacque006: Rather than doing this, you can store an init promise in the constructor and `await` its completion in async call points.

```typescript
class BlsSigner {
  readonly initPromise:Promise<void>; 
  
  constructor() {
    this.initPromise = this.init();
  };

  private async init(): Promise<void> {
    return Promise.resolve();
  }

  public async getNumber(): Promise<number> {
    await this.initPromise;
    return 42;
  }
}

const signer = new BlsSigner();
const num = await signer.getNumber();
```

contracts/clients/src/index.ts
```diff
@@ -58,4 +60,6 @@ export {
   // eslint-disable-next-line camelcase
   MockERC20__factory,
   MockERC20,
+  BlsProvider,
+  BlsSigner,
```
jacque006: We may not want to export these for use or maybe use an experimental namespace until they are ready.

jacque006 (commented):
> sendTransaction() - overridden but not implemented due to needing a signed transaction, signer.signTransaction could not be overridden due to having a non-compatible return type with BLS Wallet logic.

> sendBlsTransaction() - tested and used to replace sendTransaction()

> signMessage() - overridden but not implemented as return type was incompatible with BlsWalletWrapper signMessage return type.
signTransaction() - overridden but not implemented as return type was incompatible with BlsWalletWrapper sign return type.

> signBlsMessage() - tested and used in place of signMessage()
signBlsTransaction() - tested and used in place of signTransaction()

In general, we should strive to have exact compatibility with the provider/signer spec so users/consumers can use it their exisiting code with minimal/no chages. There are ways we can convert our args & return types to match what ethers.js is expecting.

> initWallet() - wanted a singleton BlsWalletWrapper and doing so wasn't possible in the constructor, as BlsWalletWrapper.connect() is async and you can't run async code in the constructor.
#verifyInit() - tested. Helper function that throws an error if a certain methods are called and the signer has not been initialized

See my comment on `#verifyInit`, should be able to do this without an extra function call.

jacque006:
This provider work could go in dark to `main` without an integration branch since nothing is current using it. This will make it easier for people to contribute in subsequent work, and make for smaller PR diffs.

Also, at 32 commits I would strongly encourage rebasing to a smaller set or squashing on merge.

contracts/package.json
JohnGuilding: As discussed on a call. Will address this in separate issue
[#414 Setup integration tests for BlsProvider & BlsSigner](https://github.com/web3well/bls-wallet/issues/414)

JohnGuilding:
> In general, we should strive to have exact compatibility with the provider/signer spec so users/consumers can use it their exisiting code with minimal/no chages. There are ways we can convert our args & return types to match what ethers.js is expecting.

Removed `signBlsMessage`, `signBlsTransaction` & `sendBlsTransaction`. Converted/moved the logic to the relevant ethers methods.

contracts/clients-tests/BlsProvider.test.ts
JohnGuilding: Gotcha, changed

contracts/clients-tests/BlsProvider.test.ts
JohnGuilding: Much nicer, thanks

contracts/clients/src/BlsProvider.ts
JohnGuilding: As mentioned in another comment, have removed the extra methods I added for BLS logic, and converted/moved the logic into the correct ethers methods

contracts/clients/src/BlsSigner.ts
JohnGuilding: Updated with this solution

contracts/clients/src/index.ts
JohnGuilding: Used an Experimental namespace

contracts/clients-tests/BlsProvider.test.ts
```diff
@@ -0,0 +1,514 @@
...
+    expect(transactionReceipt).to.include.keys(
+      "transactionIndex",
+      "blockHash",
+      "transactionHash",
+      "blockNumber",
+    );
+
+    // TODO: parseEther() doesn't work when asserting this way ^.
+    // Expects BigNumber { value: "0" } from transactionReceipt,
+    // but this returns { _hex: '0x00', _isBigNumber: true }.
```
JohnGuilding: Got this weird issue where my assertions do not like `parseEther` and `BigNumber.from` now. The tests pass with the below assertions (lines 337-339 at the time of writing), but they do not work with the above assertion using `.includes`. Not sure why this is happening?

To test this, uncomment the `gasUsed`, `cumulativeGasUsed` or `effectiveGasPrice` values above and run the tests.

contracts/clients/src/BlsSigner.ts
```diff
@@ -0,0 +1,215 @@
...
+      this.verificationGatewayAddress,
+      this,
+    );
+
+    const bundle = this.wallet.sign({ nonce, actions: [action] });
+    return JSON.stringify(bundleToDto(bundle));
+  }
+
+  /** Sign a message */
+  override async signMessage(message: Bytes | string): Promise<string> {
```
JohnGuilding, line 181: @jacque006 had a jab at updating the `signMessage` logic but unsure how correct it is. How does this look to you?

contracts/clients-tests/BlsProvider.test.ts
jacque006: That makes sense. Each `BigNumber` (BN) is a separate object instance, meaning two BNs with the same value will not equal, hence why `.eql` is used instead in most places in code with BNs.

Primitives in JavaScript will equal each other, but more complex types like arrays and objects will not. See https://stackoverflow.com/a/359509 for some zany ways they will with `==`. (general best practice is to always use `===`)

```typescript
const a = { a: 1 }
const b = { a: 1 }
a === b // false
a === a // true
a.a === b.a // true

BigNumber.from("1") === BigNumber.from("1") // false
BigNumber.from("1").eql(BigNumber.from("1")) // true
BigNumber.from("1").toNumber() === 1 // true
```

`.includes` should use strict equality (`===`) which means BNs would fail that check.

To resolve this, you can:
- Do your expects outside the `.include` (what you have done)
- Convert the txn receipt props to primitive types (string, number)
- Use a chai pulgin such as [chai-bignumber](https://www.chaijs.com/plugins/chai-bignumber/). `ethers.js` uses its own BN lib so this may not work.

contracts/clients-tests/BlsProvider.test.ts
```diff
@@ -0,0 +1,516 @@
...
+    // Assert
+    // TODO: bls-wallet #412 Update values returned in bundle receipt to more closely match ethers transaction response
+    expect(transactionResponse).to.be.an("object").that.deep.includes({
+      hash: transactionReceipt.transactionHash,
+      to: verificationGateway,
+      // TODO: Why is this different to signer wallet address? Uncomment logs below and run test to see problem.
+      // Will investigate this again when I look at bls-wallet #412.
+      // console.log("blsSigner wallet.address", blsSigner.wallet.address);
+      // console.log("Signer from:            ", expectedTransactionResponse.from);
+      from: "0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC",
```
jacque006: This may be the address of the aggregator that submitted the EOA txn w/ the bundle

contracts/.eslintrc.js
```diff
@@ -38,4 +38,12 @@ module.exports = {
     // TODO (merge-ok) Remove and fix lint error
     "node/no-unpublished-import": ["warn"],
   },
+  overrides: [
+    {
+      files: ["*.test.ts"],
+      rules: {
+        "no-unused-expressions": "off",
+      },
```
jacque006, line 47:
```suggestion
      // chai expect statements
      files: ["*.test.ts"],
      rules: {
        "no-unused-expressions": "off",
      },
```

contracts/clients/src/BlsSigner.ts
jacque006: I'm not 100% sure the `RLP.encode` is correct, but I think this is fine for now (great test btw :+1: ). We can *really* test this once we support [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271), likely in next version of contracts.

jacque006:
Few other minor ideas/suggestions, but I think after deciding on a rebase or squash for commits this is good to go

contracts/clients-tests/BlsProvider.test.ts
JohnGuilding: ~Turns out it's the third hardhat account. I haven't used `signers[2]` anywhere in my code.~

```ts
Account #2: 0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC (10000 ETH)
Private Key: 0x5de4111afa1a4b94908f83103eb1f1706367c2e68ca870fc3fb9a804cdab365a
```

~Could `0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC` be the address of the aggregator then?~

**Edit**: The private key belongs to the aggregator and is taken from the aggregator `.env.example` file as "PRIVATE_KEY_AGG"

contracts/clients-tests/BlsProvider.test.ts
JohnGuilding: Ahh yeah that part makes a lot more sense now. The one part that still confuses me is how `expect(transactionReceipt.gasUsed).to.equal(parseEther("0"));` still passes. 

According to the chai docs, `.equal` "Asserts that the target is strictly (===) equal to the given val." So if BNs fail using strict equality, that assertion shouldn't pass? I tested for a false positive with `expect(transactionReceipt.gasUsed).to.equal(parseEther("1"));` as a sanity check, and that assertion fails so confused how the first assertion passes. Am I understanding correctly?

JohnGuilding: In terms of what I went for in the end:

Tried `chai-bignumber` to no avail. 

Decided against converting the receipt properties to primitive types, as that would mean diverging from the ethers return type of `TransactionReceipt`. 

Decided to go with option 1 and leave the assertions as they are. Think it makes more sense to change the test code to work with the implementation code in this case, rather than change the implementation code to make the tests pass in a cleaner way (unless I'm misunderstanding point 2).

contracts/clients/src/BlsSigner.ts
JohnGuilding: Ok great, will leave for now and make a note to come back to this once EIP-1271 is supported

contracts/clients-tests/BlsProvider.test.ts
jacque006: TODO @JohnGuilding @jacque006 Determine if this from address should be agg address or wallet address, will link here

jacque006 (approved):

PR #407, opened 2022-11-22
Export decode error function from the client module

blakecduncan:
## What is this PR doing?
Splitting out the decode logic for error messages into its own function and exporting it.

No logic changes

No

## Checklist

## Guidelines

### DIFF
--- a/contracts/clients/src/OperationResults.ts
+++ b/contracts/clients/src/OperationResults.ts
@@ -33,7 +33,13 @@ export type OperationResult = {
   error?: OperationResultError;
 };
 
-export const decodeError = (errorData: string) => {
+/**
+ * Checks if a operation result error string is valid and returns
+ * the decoded error.
+ *
+ * @param errorData An error string returned by an operation result.
+ */
+export const decodeError = (errorData: string): OperationResultError => {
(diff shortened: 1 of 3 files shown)


contracts/clients/src/OperationResults.ts
```diff
@@ -28,23 +28,12 @@ export type OperationResult = {
...
-): OperationResultError | undefined => {
-  if (success) {
-    return undefined;
-  }
-
-  // Single event "WalletOperationProcessed(address indexed wallet, uint256 nonce, bool success, bytes[] results)"
-  // Get the first (only) result from "results" argument.
-  const [errorData] = results;
-
+export const decodeError = (errorData: string) => {
```
jacque006: I forgot to do this with `getOperationResults`, but would be nice to add a docstring describing this method.

jacque006 (commented):
Can you also:
- Add `decodeError` to the client's `index.ts` file as an export
- Add an example snippet to the client's `README.md` showing an example of how to use it? This also could make more sense as an example in the `decodeError` docstring.

jacque006 (approved):

PR #418, opened 2022-12-07
Charge accurate fees

### LINKED ISSUE #415: Charge accurate fees in aggregator
The overall goal of this change is to ensure that the aggregator doesn't lose money, with the following two exceptions:

1. Bundles that don't simulate accurately may result in losses
2. The aggregator might be configured to lose money on purpose (`ALLOW_LOSSES_ON_SMALL_BUNDLES`)

The aggregator currently charges a fixed fee per calldata byte and a fixed fee per gas. This approach is flawed for two reasons:

1. If gas prices go up, we're not raising our fees
2. Calldata is not a separate fee, it is included in gas

Focusing on charging calldata as an approximation seemed reasonable in the past since this dominates the cost and gas estimation seemed slow when testing with hardhat. However:

- Non-calldata gas is still relevant, and neglecting it risks getting things wrong when it makes a difference
- Slow gas estimation (eg 10 seconds) is not acceptable, and is expected to mostly be an artifact of using hardhat

Since [learning more about L2 fees](https://medium.com/offchainlabs/understanding-arbitrum-2-dimensional-fees-fd1d582596c9), it makes more sense to simply use `fee = gasPrice * gasLimit`. We do allow a token to be used instead of ETH right now, I'll preserve this by allowing for a fixed configured exchange rate, which can be upgraded to using some external source (dex?) in the future.

So, the simplest answer is to require that submitted bundles pay `gasPrice * gasLimit`. However, this will over-charge users because the entire point of our project is to reduce costs through bundle aggregation.

Therefore, we need to calculate the marginal cost of the submitted bundle by subtracting the fixed fee of processing an empty bundle (mostly the calldata of the signature). Simply charging this marginal cost is probably too low though, since the fixed overhead needs to be paid somehow.

To solve this, I will add a config parameter `BREAKEVEN_OPERATION_COUNT`, and require that bundles *additionally* pay

```
(operationCount / BREAKEVEN_OPERATION_COUNT) * fixedOverhead
```

Most of the time, `operationCount` is 1 (even if you have multiple actions), so e.g. if `BREAKEVEN_OPERATION_COUNT` is 5, then you'll have to cover 20% of `fixedOverhead`.

(I originally considered using gas to determine the contribution share, but realised that it discourages large txs from using BLS wallet, since they could quickly be asked to pay the full `fixedOverhead`, blowing out any savings. Possibly there is a middle-ground here that is better, but for now that complicates things so I'll leave it for future enhancements.)

Finally, a strictly for-profit aggregator would then refuse to submit bundles until they contained more than `BREAKEVEN_OPERATION_COUNT` operations, so your operation will just sit there if there isn't any other activity. To avoid confusion, and as a potential feature for seeding adoption, I'll add another flag `ALLOW_LOSSES_ON_SMALL_BUNDLES`.

(I'll also add `REQUIRE_FEES` which can be set to false, since with this change there won't be constants that can just be set to zero. I think we should try to move away from using free aggregators for testing though - making things that rely on this over-simplification is an obstacle to realistic testing.)

---

Note that it is expected that the required fees after this change may well still be higher than submitting ordinary transactions. There are two reasons for this:

1. BLS public keys are 128 bytes, which is much larger than the (approx) 64 bytes needed for an ECDSA signature (which includes the address)
2. An additional 52+ bytes are needed to include an action to pay `tx.origin`

We're therefore relying on the use of an expander contract to reduce our calldata to achieve cheaper fees. Measuring accurately will help position us to achieve this in subsequent changes.

voltrevo:
## What is this PR doing?

Changes the way we make bundles and charge for them, so that the aggregator has a good handle on the fees it needs to be charging and is able to turn a profit without charging more than necessary.

There's a fair bit of detail in the issue about this: https://github.com/web3well/bls-wallet/issues/415.

Previously `estimateFee` would report the fee required to submit the user's bundle in isolation, which defeats the whole purpose of saving money by aggregating with other users. This has been fixed so that the required fee reported includes only the marginal cost of the bundle and a configurable share of the bundle's overhead (basically the block space for the single signature and validating it).

When forming an aggregate bundle, we used to put all available bundles together and check that it satisfies our `estimateFee` calculation as a whole. Then, if it didn't work, we'd do a binary search to figure out how many of those bundles we could include, exclude the next one, and then try to add more recursively.

The binary search method had the advantage of processing bundles in order, allowing for things like chaining dependent bundles together (e.g. consecutive nonces). However, this relies on the idea that all the bundles are supposed to be self-sufficient (e.g. even just one bundle pays enough to cover itself), which is wrong. I decided to simplify this by evaluating bundles individually to make sure they satisfy `estimateFee`, and then just checking once that the overall bundle is profitable (or just letting it through regardless if `ALLOW_LOSSES` is `true`).

Use `./manualTests/mintNViaAggregator.ts <N>` with different values of N and check the profitability output is acceptable.

For example, here's some output I got from the aggregator with `BREAKEVEN_OPERATION_COUNT=4.5` and `./manualTests/mintNViaAggregator 5`:

```
submission-confirmed {
  hash: "0x3e06b8dd8ee59458d076d6aeb35cac83161ef9018355b2c68fdf71a65515f4a6",
  bundleHashes: [
    "0xc46e7eeae05caf5912980f5ef4e5f2977dda5203ee2571d10f357fb6ea92203a",
    "0xfba88c9e99d96b454204dc1fbe510e0d0e785bacba516e5b382e56269385af94",
    "0xf34eb24ab1f4fdf01ece625f5e80838888b89b8bbb8957aa9b50ce6a193742fb",
    "0xab6d3299ff437e61a8615421e5cf0173099f158eb71822a1cd50662032884ce1",
    "0xd53760964fecba10a3b08a4afeaeebad39367e874e8986bb686a7338ff2acf54"
  ],
  blockNumber: 166,
  profit: "0.00005934235",
  cost: "0.000483594",
  expectedMaxCost: "0.000483594",
  actualFee: "0.00054293635",
  expectedFee: "0.00054293635"
}
```

When `N` is greater than `BREAKEVEN_OPERATION_COUNT`, bundles should be profitable. Note that due to fluctuations, `mintNViaAggregator` is additionally adding 10% to the fee, so smaller values of N may be profitable too.

You can also check that if `ALLOW_LOSSES` is `false`, then the aggregator generally won't submit unprofitable bundles, and will at least not submit bundles it expects to make losses on (`expectedFee` should be higher than `expectedMaxCost` (both of these may deviate from their *actual* values)).

Resolves https://github.com/web3well/bls-wallet/issues/415

## Checklist

## Guidelines


aggregator/src/app/EthereumService.ts
```diff
@@ -100,13 +101,41 @@ export default class EthereumService {
...
+    if (blsNonce.eq(0)) {
+      if (!env.AUTO_CREATE_INTERNAL_BLS_WALLET) {
+        throw new Error([
+          "Required internal bls wallet does not exist. Either enable",
+          "AUTO_CREATE_INTERNAL_BLS_WALLET or run",
+          "./programs/createInternalBlsWallet.ts",
+        ].join(" "));
+      }
+
+      await (await VerificationGateway__factory.connect(
```
blakecduncan, line 122: nit but this double `await` with a `.wait()` at the end is confusing to read.  Could we break this up into multiple variables/lines?

aggregator/src/app/EthereumService.ts
```diff
@@ -331,6 +363,21 @@ export default class EthereumService {
...
 
+  async GasConfig() {
+    const block = await this.wallet.provider.getBlock("latest");
+    const previousBaseFee = block.baseFeePerGas;
+    assert(previousBaseFee !== null && previousBaseFee !== nil);
+
+    return {
+      maxFeePerGas: previousBaseFee
+        .add(
+          previousBaseFee.mul(env.PREVIOUS_BASE_FEE_PERCENT_INCREASE).div(100),
```
blakecduncan, line 374: Might be nice to give this a variable for readability. ex:

```
const feeIncrease = previousBaseFee.mul(env.PREVIOUS_BASE_FEE_PERCENT_INCREASE).div(100);
```

aggregator/src/app/EthereumService.ts
```diff
@@ -331,6 +363,21 @@ export default class EthereumService {
     throw new Error("Expected return or throw from attempt loop");
   }
 
+  async GasConfig() {
+    const block = await this.wallet.provider.getBlock("latest");
+    const previousBaseFee = block.baseFeePerGas;
+    assert(previousBaseFee !== null && previousBaseFee !== nil);
+
+    return {
+      maxFeePerGas: previousBaseFee
```
blakecduncan, line 372: Could be useful to have a comment that mentions why we set maxFeePerGas to the previousBaseFee + a fee increase + a priority fee per gas.

aggregator/src/app/AggregationStrategy.ts
```diff
@@ -2,216 +2,431 @@ import {
...
+      const bestRow = await this.#pickBest(rowGroup);
+      const conflictingRows = rowGroup.filter((r) => r !== bestRow);
+
+      filteredRows = filteredRows.filter((r) => !conflictingRows.includes(r));
+    }
+
+    return filteredRows;
+  }
+
+  async #pickBest(rows: BundleRow[]) {
```
blakecduncan, line 317: Could we add a comment here explaining how we pick the best?

aggregator/src/app/AggregationStrategy.ts
```diff
@@ -2,216 +2,431 @@ import {
...
 
-    return {
-      aggregateBundle: aggregateBundle.operations.length > 0
-        ? aggregateBundle
-        : nil,
+    if (includedRows.length === 0) {
+      return {
+        aggregateBundle: nil,
+        includedRows: [],
+        bundleOverheadCost: bundleOverheadGas.mul(
```
blakecduncan, line 135: My brain is moving a little slow 😅 why is this the bundleOverheadGas * maxFeePerGas ?

aggregator/src/app/AggregationStrategy.ts
```diff
@@ -2,216 +2,431 @@ import {
...
+      if (excessFees[i].gt(bestExcessFee)) {
+        bestExcessFee = excessFees[i];
+        bestExcessFeeIndex = i;
+      }
+    }
+
+    return rows[bestExcessFeeIndex];
+  }
+
   async #augmentAggregateBundle(
```
blakecduncan, line 341: I know you didn't edit this, but it might be nice to note what are we augmenting

blakecduncan (approved):
Tested out your script and looks like everything is working.  I also really like the events you added 💯 .

The comments I added are non-blocking, they are just around being more descriptive.  I think some added comments describing some of the logic and maybe breaking some things out into variables could be useful for readability.

blakecduncan:
Also, I really like your description of the PR and the description in the linked issue.  Do you think we should add any of that explanation to the Readme?

aggregator/src/app/AggregationStrategy.ts
voltrevo: It's just the basic gas mechanism - transactions consume gas, and you pay a fee per unit of gas, therefore the cost is the amount of gas times the fee paid per gas.

aggregator/src/app/EthereumService.ts
voltrevo: This is a pretty common pattern, so my vote is to keep it compact like this. Might be worth putting this to the team though.

aggregator/src/app/EthereumService.ts
voltrevo: Fixed in https://github.com/web3well/bls-wallet/pull/438

aggregator/src/app/EthereumService.ts
voltrevo: Fixed in https://github.com/web3well/bls-wallet/pull/438

aggregator/src/app/AggregationStrategy.ts
voltrevo: Fixed in https://github.com/web3well/bls-wallet/pull/438

aggregator/src/app/AggregationStrategy.ts
voltrevo: Fixed in https://github.com/web3well/bls-wallet/pull/438

voltrevo:
> Also, I really like your description of the PR and the description in the linked issue. Do you think we should add any of that explanation to the Readme?

Cheers. Fixed in https://github.com/web3well/bls-wallet/pull/438.

aggregator/src/app/EthereumService.ts
blakecduncan: No worries, if it's common to do it like this I'm cool with it 👍

PR #422, opened 2022-12-09
Bls provider json rpc provider methods

### LINKED ISSUE #381: BlsProvider JsonRpcProvider Methods
We are adding a BlsProvider to `bls-wallet-clients`. This issue is concerned with ensuring the JsonRpcProvider Methods work with the BlsProvider. This will include overriding methods with BLS logic if needed and testing the other methods work as expected.

Some of the operations include getting the signer, listing accounts and sending raw messages.

[JsonRpcProvider Methods](https://docs.ethers.io/v5/api/providers/jsonrpc-provider/) to test/override:

```ts
// Connect to a JSON-RPC HTTP API using the URL or ConnectionInfo urlOrConnectionInfo 
// connected to the networkish network.
new ethers.providers.JsonRpcProvider( [ urlOrConnectionInfo [ , networkish ] ] )

// The fully formed ConnectionInfo the Provider is connected to.
jsonRpcProvider.connection ⇒ ConnectionInfo

// Returns a JsonRpcSigner which is managed by this Ethereum node, at addressOrIndex. If no 
// addressOrIndex is provided, the first account (account #0) is used.
jsonRpcProvider.getSigner( [ addressOrIndex ] ) ⇒ JsonRpcSigner

// No description given in docs
jsonRpcProvider.getUncheckedSigner( [ addressOrIndex ] ) ⇒ JsonRpcUncheckedSigner

// Returns a list of all account addresses managed by this provider.
jsonRpcProvider.listAccounts( ) ⇒ Promise< Array< string > >

// Allows sending raw messages to the provider.
jsonRpcProvider.send( method , params ) ⇒ Promise< any >
```

JohnGuilding:
## What is this PR doing?
This PR is concerned with adding/testing the [JsonRpcProvider methods](https://docs.ethers.io/v5/api/providers/jsonrpc-provider/#JsonRpcProvider).

Relevant methods:

```ts
jsonRpcProvider.connection
getSigner()
getUncheckedSigner()
listAccounts()
send()
```

**Notes:**

**getSigner()** - was needed in previous work, so implementation and tests have already been added.

**getUncheckedSigner()** - in order to test this method fully, I had to add several additional methods to the signer along with tests to ensure the unchecked signer worked as expected. This includes adding the `UncheckedBlsSigner` class itself. Methods added/changed:

- `signer.connectBlsUnchecked()` - see [comment](https://github.com/web3well/bls-wallet/pull/422#discussion_r1044535148) on why I added this
- `signer.sendUncheckedTransaction()`
- `uncheckedBlsSigner.sendTransaction()`

For more information on what the unchecked signer is and why it was added to ethers see this [part of the docs](https://docs.ethers.io/v5/api/providers/jsonrpc-provider/#UncheckedJsonRpcSigner) and this [github issue](https://github.com/ethers-io/ethers.js/issues/340)

**send()** - added some tests here to ensure we called the underlying ethers logic, and got the expected results for some common rpc requests. See [this comment](https://github.com/web3well/bls-wallet/pull/422/files#r1044535148) for further discussion.

1. run local hardhat node
2. Spin up a local aggregator
3. in the `./contracts` directory, run `yarn test-clients`

#381  

## Checklist

## Guidelines

### DIFF
--- a/contracts/clients-tests/BlsProvider.test.ts
+++ b/contracts/clients-tests/BlsProvider.test.ts
@@ -461,6 +461,38 @@ describe("BlsProvider", () => {
     // Assert
     expect(accounts).to.deep.equal(expectedAccounts);
   });
+
+  it("should send an rpc request to the provider", async () => {
+    // Arrange
+    const expectedBlockNumber = await regularProvider.send(
+      "eth_blockNumber",
+      [],
+    );
+    const expectedChainId = await regularProvider.send("eth_chainId", []);
+    const expectedAccounts = await regularProvider.send("eth_accounts", []);
+
+    const recipient = signers[1].address;
+    const transactionAmount = parseEther("1");
+    const hexTx = ethers.providers.JsonRpcProvider.hexlifyTransaction({
+      to: recipient,
+      value: transactionAmount,
+    });
+    const balanceBefore = await regularProvider.getBalance(recipient);
+
+    // Act
+    const blockNumber = await blsProvider.send("eth_blockNumber", []);
+    const chainId = await blsProvider.send("eth_chainId", []);
+    const accounts = await blsProvider.send("eth_accounts", []);
+    await blsProvider.send("eth_sendTransaction", [hexTx]);
+
+    // Assert
+    expect(blockNumber).to.equal(expectedBlockNumber);
+    expect(chainId).to.equal(expectedChainId);
+    expect(accounts).to.deep.equal(expectedAccounts);
+    expect(
+      (await regularProvider.getBalance(signers[1].address)).sub(balanceBefore),
+    ).to.equal(transactionAmount);
+  });
 });
 
 describe("JsonRpcProvider", () => {
--- a/contracts/clients-tests/BlsProvider.test.ts
+++ b/contracts/clients-tests/BlsProvider.test.ts
@@ -2,7 +2,7 @@
 import { ethers as hardhatEthers } from "hardhat";
 import chai, { expect } from "chai";
 import spies from "chai-spies";
-import { ethers } from "ethers";
+import { ethers, BigNumber } from "ethers";
 import { parseEther, formatEther, id } from "ethers/lib/utils";
 import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
 
@@ -13,6 +13,7 @@ import {
   NetworkConfig,
 } from "../clients/src";
 import getNetworkConfig from "../shared/helpers/getNetworkConfig";
+import BlsSigner, { UncheckedBlsSigner } from "../clients/src/BlsSigner";
 
 chai.use(spies);
 
@@ -76,6 +77,16 @@ describe("BlsProvider", () => {
 
     // Assert
     expect(blsSigner._isSigner).to.be.true;
+    expect(blsSigner).to.be.instanceOf(BlsSigner);
+  });
+
+  it("should return a valid unchecked bls signer", async () => {
+    // Arrange & Act
+    const uncheckedBlsSigner = blsProvider.getUncheckedSigner(privateKey);
+
+    // Assert
+    expect(uncheckedBlsSigner._isSigner).to.be.true;
+    expect(uncheckedBlsSigner).to.be.instanceOf(UncheckedBlsSigner);
   });
 
   it("should return a new signer if one has not been instantiated", async () => {
@@ -324,9 +335,9 @@ describe("BlsProvider", () => {
       type: 2,
     });
 
-    expect(transactionReceipt.gasUsed).to.equal(parseEther("0"));
-    expect(transactionReceipt.cumulativeGasUsed).to.equal(parseEther("0"));
-    expect(transactionReceipt.effectiveGasPrice).to.equal(parseEther("0"));
+    expect(transactionReceipt.gasUsed).to.equal(BigNumber.from("0"));
+    expect(transactionReceipt.cumulativeGasUsed).to.equal(BigNumber.from("0"));
+    expect(transactionReceipt.effectiveGasPrice).to.equal(BigNumber.from("0"));
 
     expect(transactionReceipt).to.include.keys(
       "transactionIndex",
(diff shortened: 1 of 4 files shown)


contracts/clients-tests/BlsProvider.test.ts
```diff
@@ -428,6 +439,60 @@ describe("BlsProvider", () => {
...
+    const expectedAccounts = await regularProvider.listAccounts();
+
+    // Act
+    const accounts = await blsProvider.listAccounts();
+
+    // Assert
+    expect(accounts).to.deep.equal(expectedAccounts);
+  });
+
+  it("should send an rpc request to the provider", async () => {
```
JohnGuilding, line 468: Undecided on further changes here and whether we should override/extend (not sure what correct terminology is here) the rpc calls to execute bls logic instead, maybe similar to [how we've done in the QuillProvider](https://github.com/web3well/bls-wallet/blob/4f5b7c75d9cf2baef9e44469f016a89d819df4d2/extension/source/background/AggregatorController.ts#L38-L97)?. The overarching question to answer here is **what should the expected behaviour be when a developer uses `blsProvider.send()`?**

contracts/clients/src/BlsSigner.ts
```diff
@@ -200,17 +200,56 @@ export default class BlsSigner extends Signer {
...
-  connectUnchecked(): ethers.providers.JsonRpcSigner {
-    throw new Error("connectUnchecked() is not implemented.");
+  connectUnchecked(): BlsSigner {
+    throw new Error(
+      "connectUnchecked() is not implemented. Use connectBlsUnchecked instead.",
+    );
+  }
+
+  // When we use regular "connectUnchecked", we need to pass the private key into the method as that is a required
+  // argument for the BlsSigner constructor. This breaks liskov substitution as the subclass method signature is different.
```
JohnGuilding: In a previous PR, there was an agreement that adding custom methods to be used in place of an ethers method was undesirable. So I've created this method apprehensively, but I'm unsure how we could get `connectUnchecked` to work due to needing a different method signature. Open to suggestions on better ways to resolve this.

contracts/clients-tests/BlsProvider.test.ts
jacque006: I would expect this to work similarly to the `QuillProvider`/be BLS Wallet specific when:
- Sending a transaction
- Sending multiple transactions (currently overriding/extending `eth_sendTransaction`, we really should switch to a custom RPC namespace like `blsWallet_sendTransactions`).
- Looking up a receipt from a BLS Wallet txn/bundle

And like a normal provider for other send calls.

We actually can't send/sign normal EOA transactions & messages (ECDSA) as we only have our BLS pub/priv keys. We *may* want to add that in the future if:
- Someone wants to bypass an aggregator and submit straight to chain
- Someone wants to use this provider to write a custom aggregator
- We want to use this in our own aggregator

But I think those are out of scope of this initial implementation and we should focus mostly on use cases where a dApp/script/service want an embedded BLS burner wallet, similar to the wallet @blakecduncan created.

contracts/clients/src/BlsSigner.ts
```diff
@@ -200,17 +200,56 @@ export default class BlsSigner extends Signer {
...
+  connectBlsUnchecked(privateKey: string): BlsSigner {
+    return new UncheckedBlsSigner(
+      _constructorGuard,
+      this.provider,
+      privateKey,
+      this._address || this._index,
+    );
```
```suggestion
  connectUnchecked(): BlsSigner {
    return new UncheckedBlsSigner(
      _constructorGuard,
      this.provider,
      // Modify this arg to take a priv key OR promise that will eventually resolve to one, then do similar initPromise guards in this class.
      this.wallet?.privateKey ?? async () => {
      	await this.initPromise();
      	return this.wallet.privateKey;
      },
      this._address || this._index,
    );
  }
```

contracts/clients/src/BlsSigner.ts
jacque006: https://github.com/web3well/bls-wallet/pull/422/files#r1044853129

contracts/clients/src/BlsSigner.ts
jacque006: This requires modifying the `BlsSigner` constructor to accept a `privateKey` or a promise that will eventually resolve to one. We can discuss offline some cleaner ways to do this.

contracts/clients-tests/BlsProvider.test.ts
JohnGuilding: Thanks for the thoughts. Makes sense regarding expected behaviour. 

Thinking the same that this work is out of scope for initial implementation. Have raised an issue to look at this in the future (out of scope for current milestone): #424

contracts/clients-tests/BlsSigner.test.ts
```diff
@@ -370,6 +371,137 @@ describe("BlsSigner", () => {
...
+
+    // Assert
+    expect(
+      (await blsProvider.getBalance(recipient)).sub(balanceBefore),
+    ).to.equal(transactionAmount);
+  });
+
+  it("should throw an error getting the transaction receipt when using a new provider and connecting to an unchecked bls signer", async () => {
+    // Arrange & Act
+    // This test is identical to the above test except this one uses a new instance of a provider, yet fails to find the tx receipt
```
JohnGuilding: As described in the comment, this test fails to find the tx receipt when instantiating the unchecked signer from a new provider instance. Not sure why this is happening?

contracts/clients/src/BlsSigner.ts
```diff
@@ -200,17 +201,55 @@ export default class BlsSigner extends Signer {
     throw new Error("_signTypedData() is not implemented.");
   }
 
-  connectUnchecked(): ethers.providers.JsonRpcSigner {
-    throw new Error("connectUnchecked() is not implemented.");
+  connectUnchecked(): BlsSigner {
+    return new UncheckedBlsSigner(
+      _constructorGuard,
+      this.provider,
+      // Modify this arg to take a priv key OR promise that will eventually resolve to one, then do similar initPromise guards in this class.
```
jacque006: I included this originally to illustrate my suggestion, not needed in code IMO

```suggestion
```

contracts/clients-tests/BlsSigner.test.ts
jacque006: https://github.com/web3well/bls-wallet/issues/427

jacque006 (approved):

PR #425, opened 2022-12-13
Add provider tests to github workflows

JohnGuilding:
## What is this PR doing?
This PR adds the newly created Bls Provider and Signer tests to the github workflows and tests them against an aggregator instance.

**Other changes:**
- Added aggregator logs artifact (See below workflow run and scroll to the bottom of the page)
- Move chai config to single location
- Tests were failing unexpectedly due to reuse of state such as private keys. Each test now uses a new private key

## How can these changes be manually tested?
1. Go to actions and observe the [following run](https://github.com/web3well/bls-wallet/actions/runs/3831809306)
2. Check all tests pass in the `yarn test-integration` step

## Does this PR resolve or contribute to any issues?
#414 

## Checklist
- [x] I have manually tested these changes
- [x] Post a link to the PR in the group chat

## Guidelines
- If your PR is not ready, mark it as a draft
- The `resolve conversation` button is for reviewers, not authors
  - (But add a 'done' comment or similar)

PR #429, opened 2022-12-15
Correctly bubble up bundle submit errors

jacque006:
## What is this PR doing?
- Correctly TS types and returns aggregator submit errors. This broke in previous PR review changes when we switched the HTTP return code on aggregator submit errors.
- Use local (`globalThis`) implementation of fetch instead of `node-fetch` if available. This resolved issues with transpilation via esm.sh and will eventually allow us to move off `node-fetch` in newer NodeJS versions.
- Update client deps
- Update `Aggregator` client documentation

- Manually run script that submits both valid and invalid bundles, see invalid bundles return error on receipt lookup
- CI tests
- Additional integration tests can be added in followup after https://github.com/web3well/bls-wallet/pull/425 is merged to prevent future regressions.

No

## Checklist

## Guidelines

### DIFF
--- a/contracts/clients/package.json
+++ b/contracts/clients/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bls-wallet-clients",
-  "version": "0.8.0",
+  "version": "0.8.1",
   "description": "Client libraries for interacting with BLS Wallet components",
   "main": "dist/src/index.js",
   "types": "dist/src/index.d.ts",
@@ -23,17 +23,17 @@
   "dependencies": {
     "@thehubbleproject/bls": "^0.5.1",
     "ethers": "5.5.4",
-    "node-fetch": "2"
+    "node-fetch": "^3.3.0"
   },
   "devDependencies": {
-    "@types/chai": "^4.3.3",
+    "@types/chai": "^4.3.4",
     "@types/chai-as-promised": "^7.1.5",
-    "@types/mocha": "^10.0.0",
-    "chai": "^4.3.6",
+    "@types/mocha": "^10.0.1",
+    "chai": "^4.3.7",
     "chai-as-promised": "^7.1.1",
-    "mocha": "^10.1.0",
+    "mocha": "^10.2.0",
     "source-map-support": "^0.5.21",
     "ts-node": "^10.9.1",
-    "typescript": "^4.8.4"
+    "typescript": "^4.9.4"
   }
 }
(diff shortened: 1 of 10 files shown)


contracts/clients/package.json
```diff
@@ -23,17 +23,17 @@
   "dependencies": {
     "@thehubbleproject/bls": "^0.5.1",
     "ethers": "5.5.4",
-    "node-fetch": "2"
+    "node-fetch": "2.6.7"
```
jacque006, line 26: I really wish JSON allowed comments... https://stackoverflow.com/questions/69081410/error-err-require-esm-require-of-es-module-not-supported

extension/source/background/AggregatorController.ts
```diff
@@ -124,6 +124,9 @@ export default class AggregatorController {
       const aggregator = new Aggregator(knownTx.aggregatorUrl);
 
       const bundleReceipt = await aggregator.lookupReceipt(hash);
+      if (bundleReceipt && 'submitError' in bundleReceipt) {
+        throw new Error(bundleReceipt.submitError);
+      }
```
jacque006, line 129: @kautukkundan @blakecduncan Is this the correct behavior (throw error in RPC handler) if there is an error with the aggregator trying to submit a bundle?

contracts/clients/src/Aggregator.ts
```diff
@@ -71,24 +93,36 @@ export default class Aggregator {
...
+   * Looks for a transaction receipt for a Bundle sent to the aggregator.
+   * This will return undefined if the bundle has not yet been submitted by the aggregator.
+   *
+   * @param hash Hash of the bundle to find a transaction receipt for.
+   * @returns The bundle receipt, a submission error if the aggregator was unable to submit the bundle on chain, or undefined if the receipt was not found.
+   */
+  async lookupReceipt(
+    hash: string,
+  ): Promise<BundleReceipt | BundleReceiptError | undefined> {
+    return this.jsonGet<BundleReceipt | BundleReceiptError>(
```
blakecduncan, line 118: Thanks for adding this in 💪

extension/source/background/AggregatorController.ts
blakecduncan: Kautuk and I haven't caught up on this.  I think that is a good behaviour.  In the future, it may be good for us to update the transaction in the transactions table if there is a submit error.  I'd say that's outside the scope of this bug fix though because I don't think we have a good way to display a submit error yet.

blakecduncan (approved):
Makes sense to me.  Thanks for fixing this

Issue #383, opened 2022-11-04
BlsSigner JsonRpcSigner & Signer Methods


JohnGuilding:
We are adding a BlsSigner to `bls-wallet-clients`. This issue is concerned with ensuring the JsonRpcSigner & Signer Methods work with the BlsSigner. This will include overriding methods with BLS logic if needed and testing the other methods work as expected.

Some of the operations include connecting to an [unchecked signer](https://docs.ethers.io/v5/api/providers/jsonrpc-provider/#UncheckedJsonRpcSigner), getting the account address and checking if an object is a signer.

**Note 1**: The term Signer Methods here does not mean every method on the ethers `Signer` class. This refers to the [following section](https://docs.ethers.io/v5/api/signer/#Signer) in their docs that group these methods under no specific heading. Other `Signer` methods are included in other issues.

**Note 2**: There is one method called `getUncheckedSigner()` that returns a `JsonRpcUncheckedSigner`. If we were to implement this method, we would have to ensure the new `BlsJsonRpcUncheckedSigner` works as expected. If this is easy to implement, we should do it as part of this issue. If not, we should raise a separate issue to address this. Using an unchecked signer is also quite a niche requirement so we may also want to decide whether it is worth implementing and testing a `BlsJsonRpcUncheckedSigner`.

[JsonRpcSigner Methods](https://docs.ethers.io/v5/api/providers/jsonrpc-provider/#JsonRpcSigner) to test/override:

```ts
// The provider this signer was established from.
signer.provider ⇒ JsonRpcProvider

// Returns a new Signer object which does not perform additional checks when sending 
// a transaction. See getUncheckedSigner for more details.
signer.connectUnchecked( ) ⇒ JsonRpcUncheckedSignersource

// Sends the transaction and returns a Promise which resolves to the opaque transaction hash.
signer.sendUncheckedTransaction( transaction ) ⇒ Promise< string< DataHexString< 32 > > >source

// Request the node unlock the account (if locked) using password.
signer.unlock( password ) ⇒ Promise< boolean >source
```

[Signer Methods](https://docs.ethers.io/v5/api/signer/#Signer) to test/override:
```ts
// Sub-classes must implement this, however they may simply throw an error if 
// changing providers is not supported.
signer.connect( provider ) ⇒ Signer

// Returns a Promise that resolves to the account address.
signer.getAddress( ) ⇒ Promise< string< Address > >

// Returns true if and only if object is a Signer.
[...]

PR #439, opened 2023-01-06
Bls signer json rpc signer and default signer methods

JohnGuilding:
## What is this PR doing?
This PR completes BlsSigner JsonRpcSigner & Signer Methods #383. The following methods have been added/tested: 

- `signer.provider` - test added
- `signer.connectUnchecked( )` - implementation and tests added as part of  #422 
- `signer.sendUncheckedTransaction()` - implementation and tests added as part of #422 
- `signer.unlock()` - no test or implementation added (see comment)
- `signer.connect()` - test added
- `signer.getAddress()` - test added as part of previous work
- `signer.isSigner()` - test added

- spin up hardhat node and local aggregator
- run `yarn test-integration`

#383 

## Checklist

## Guidelines


JohnGuilding:
The [signer.unlock()](https://github.com/web3well/bls-wallet/blob/2e2b4f3cf22510a70b98b118f02fc671a64f9056/contracts/clients/src/BlsSigner.ts#L143-L153) method was added in a previous PR. I have not added a test as I'm not too sure how I would test it and I don't have a clear understanding of what the difference between a locked account and an unlocked account are. I have copied the ethers implementation which calls `provider.send("personal_unlockAccount, ...",` under the hood.

According to the [geth docs](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-personal), `personal_unlockAccount` "_Decrypts the key with the given address from the key store_".

Open to suggestions on a test to write or whether it's ok to leave it as it is (or remove the method altogether)

jacque006:
@JohnGuilding The `personal` namespace appears to mainly deal with managing the EOA accounts in a node (creating, unlocking encrypted keys, etc). It is also deprecated. From the loose amount I have read, it is generally not enabled on public nodes as anyone could use your unlocked account.

Almost all of the operations there have to deal with managing keys for ECDSA signing. Since we deal with BLS keys and those are not supported natively by execution clients, there are a couple ways we could go about calls that use that namespace:
1. Throw an unimplemented error with some messaging around the different key/sig types
2. Pass through the calls and allow users to manipulate EOA accounts through our provider (What you opt'd to do)
3. Make the calls do the equivalent operation with BLS keys

Of these, I would lean towards what you've done (option 2) or option 1. It may be worth adding a test for and just confirming it does not error and returns the correct boolean value. If the `personal` namespace is not supported by the Hardhat node we use for testing then I would just have a test that confirms the error and leave a note for that test. That can be done as a low priority follow up issue.

jacque006 (approved):

JohnGuilding:
@jacque006 Thanks for the description. Added a test to assert the correct error is thrown and added some comments. New low priority issue: #444

Issue #408, opened 2022-11-23
Add status column to bundles table in the aggregator DB


blakecduncan:
Follow up from this comment https://github.com/web3well/bls-wallet/pull/377#discussion_r1030413936

jacque006:
Completed in https://github.com/web3well/bls-wallet/pull/440

PR #440, opened 2023-01-09
Persist bundles in table

### LINKED ISSUE #190: Store confirmed/historical bundles in db
Currently, confirmed/submitted bundles are not persisted in the aggregator. We should add a db table or column for status which tracks them so they can be looked up in the future, or be available between aggregator restarts.

blakecduncan:
## What is this PR doing?
This update makes it so we no longer delete bundles for the DB
- There is now a status column on the bundles table.  That can be "pending", "failed", "confirmed"
- Also added a receipt column because we were previously storing the transaction receipt in memory for a confirmed bundle

**Note: Tests seem to be hanging when running through GitHub actions. They pass for me locally, I'm looking into what could be the issue running them remotely**

- Stopping and starting the Agg doesn't clear out confirmed bundles anymore.  So hitting the `/bundleReceipt` endpoint after restarting the Agg for a confirmed bundle should still return the bundle receipt

https://github.com/web3well/bls-wallet/issues/408
resolves https://github.com/web3well/bls-wallet/issues/190

## Checklist

## Guidelines


aggregator/src/app/BundleTable.ts
```diff
@@ -82,6 +103,7 @@ function toInsertRawRow(row: InsertRow): InsertRawRow {
     bundle: JSON.stringify(bundleToDto(row.bundle)),
     eligibleAfter: toUint256Hex(row.eligibleAfter),
     nextEligibilityDelay: toUint256Hex(row.nextEligibilityDelay),
+    receipt: JSON.stringify(row.receipt),
```
JohnGuilding, line 106: Assume there is reason for this. Why do we only need to add the receipt property here and not the status as well? Same question for `toRawRow` below

aggregator/src/app/BundleTable.ts
```diff
@@ -24,20 +25,27 @@ import nil from "../helpers/nil.ts";
...
+  status: string;
   hash: string;
   bundle: string;
   eligibleAfter: string;
   nextEligibilityDelay: string;
   submitError: string | null;
+  receipt: string | null;
 };
 
+const BundleStatuses = ["pending", "confirmed", "failed"] as const;
```
JohnGuilding, line 37: Is it possible for this to be an enum? Would that make parsing the type easier? More of an open question than a suggestion

aggregator/test/BundleServiceFees.test.ts
```diff
@@ -115,7 +116,8 @@ Fixture.test("submits bundle with sufficient token fee", async (fx) => {
     approveAndSendTokensToOrigin(fx, await wallet.Nonce(), oneToken),
   );
 
-  assertBundleSucceeds(await bundleService.add(bundle));
+  const bundleResponse: ExplicitAny = await bundleService.add(bundle);
```
JohnGuilding: What was the use case for ExplicitAny here? Would this not be `AddBundleResponse`?

aggregator/src/app/BundleTable.ts
```diff
@@ -53,25 +61,38 @@ export type BundleRow = Row;
...
-  if ("failures" in parseResult) {
-    throw new Error(parseResult.failures.join("\n"));
+  const parseBundleResult = parseBundleDto(JSON.parse(rawRow.bundle));
+  if ("failures" in parseBundleResult) {
+    throw new Error(parseBundleResult.failures.join("\n"));
+  }
+
+  const status = rawRow.status;
+  if (!isValidStatus(status)) {
+    throw new Error(`Not a valid bundle status: ${status}`);
```
JohnGuilding, line 81: Does this error or `isValidStatus` need a test?

JohnGuilding (approved):
Looks good - tested it out and could retrieve a bundleReceipt after restarting the aggregator. Left 1 or 2 minor Qs

aggregator/test/BundleServiceFees.test.ts
blakecduncan: `AddBundleResponse` doesn't have the hash type on it.  I was thinking about changing the type but wanted to keep the changes related to what I was updating.  I look at updating it though

aggregator/src/app/BundleTable.ts
blakecduncan: We could make it an Enum.  I'm definitely open to suggestions but don't think there would be any functional benefit to that though.  I used a string literal because I wanted the type ahead and type checking when someone creates a bundle in the code. We could do the same with an Enum but would need the same checks when getting the string from the db and converting it to an Enum.

aggregator/src/app/BundleTable.ts
blakecduncan: I added `isValidStatus()` more for the typing benefits rather than to check if the DB returns a valid status. The DB is just returning the status as a string and I want it to be typed as a BundleStatus which is why I needed to add this.  I also added the error for typing reasons.  I don't expect this error to be thrown though.  I'd lean towards saying this doesn't need a test but I'd be happy to add one

aggregator/src/app/BundleTable.ts
blakecduncan: Yeah we don't have to have the status here because I'm not transforming the status value like I am the receipt value.  Since the receipt is an object I'm stringifying it to save it in the DB.  Then we have to parse that string when we get it back from the DB into an object.  Since the status is already a string it gets added to the object with the spread of the row object (`...row,`).

Saving the receipt as a string is kind of a 'quick win'. rather than having to do something like create a new `receipt` table where every possible value on the receipt is a column.  We do the same thing with the bundle column in the bundles table.

jacque006:
@blakecduncan The aggregator tests appear to be timing out on whatever test case is running after `removes failing bundle when its eligibility delay is larger than MAX_ELIGIBILITY_DELAY`

aggregator/test/BundleServiceFees.test.ts
blakecduncan: hmm sorry, I guess hash is included in the AddBundleResponse type, Typescript is complaining for some reason when I try to access it inside the response though.

aggregator/src/app/BundleTable.ts
JohnGuilding: Just realised I didn't respond to these - thanks for the detailed answers!

Ok gotcha, sounds good to me.

aggregator/src/app/BundleTable.ts
JohnGuilding: Ok nice, how it is already sounds good

PR #443, opened 2023-01-10
Limit bundles by gas instead of actions

### LINKED ISSUE #416: Limit bundle size using gas instead of actions
Bundles need to fit inside the block gas limit to be valid.

Technically even 2 actions is too many, since we could individually validate both actions but then together they don't fit in a block.

It's therefore essential that we use gas instead.

voltrevo:
## What is this PR doing?

See title / #416.

- Replaces `MAX_AGGREGATION_SIZE` with `MAX_GAS`
- Adds `BUNDLE_CHECKING_CONCURRENCY` (existed previously but it was hardcoded)
- Fixes polling for unconfirmed space (since old method relied on action counting and needed to be replaced anyway)
- The aggregation delay is now bypassed when `BREAKEVEN_OPERATION_COUNT` is exceeded instead of when the action limit is reached (since there no longer is an action limit)

Run the tests.

Resolves #416.

## Checklist

## Guidelines


voltrevo:
~~Note: There seems to be an unrelated issue running the tests on github actions, but it reliably passes locally.~~

aggregator/README.md
```diff
@@ -49,7 +49,7 @@ commands.
...
 | BUNDLE_QUERY_LIMIT                 | 100                                                                | Maximum number of bundles returned from Postgres                                                                                                                                                                                   |
-| MAX_AGGREGATION_SIZE               | 12                                                                 | Maximum number of actions from bundles which will be aggregated together for submission on chain                                                                                                                                   |
+| MAX_GAS                            | 2000000                                                            | Limits the amount of user operations which can be bundled together by using this value as the approximate limit on the amount of gas in an aggregate bundle.                                                                       |
```
jacque006, line 52: Something like `MAX_GAS_PER_BUNDLE` would provide a better initial description.

aggregator/README.md
```diff
@@ -61,6 +61,7 @@ commands.
...
 | PRIORITY_FEE_PER_GAS               | 0                                                                  | The priority fee used when submitting bundles (and passed on as a requirement for user bundles).                                                                                                                                   |
 | PREVIOUS_BASE_FEE_PERCENT_INCREASE | 2                                                                  | Used to determine the max basefee attached to aggregator transaction (and passed on as a requirement for user bundles)s.                                                                                                           |
+| BUNDLE_CHECKING_CONCURRENCY        | 8                                                                  | Limits the concurrency of client requests when checking bundles (getting gas usage, detecting fees, etc).                                                                                                                          |
```
jacque006, line 64: Is `client requests` equivalent to RPC calls? Or more like the number of bundles that are checked concurrently? Might be better worded like:

```suggestion
| BUNDLE_CHECKING_CONCURRENCY        | 8                                                                  | The maximum number of bundles that are checked concurrently (getting gas usage, detecting fees, etc).                                                                                                                          |
```

jacque006 (approved):

aggregator/README.md
voltrevo: #468

aggregator/README.md
voltrevo: > Is `client requests` equivalent to RPC calls? Or more like the number of bundles that are checked concurrently?

The latter... good point 👍
Implemented in #468.

Issue #403, opened 2022-11-22
Aggregator hardening


voltrevo:
Identify & pursue issues that would be barriers to running a non-test aggregator

- [x] https://github.com/web3well/bls-wallet/issues/415
- [x] https://github.com/web3well/bls-wallet/issues/416
- [ ] https://github.com/web3well/bls-wallet/issues/389 - being dealt with separately 
- [x] #453

jacque006:
One thing I know we will need to do is move from storing submitted bundles in memory to the db https://github.com/web3well/bls-wallet/issues/190

The other is bootstrapping/syncing chain state on first start/restarts https://github.com/web3well/bls-wallet/issues/191

jacque006:
Also worth considering a production infra situation where more than one agg instance is running at a time. Some things should work running off the same postgres db, but I imagine some other things would break such as aggs trying to submit the same bundles/ops without being aware the other agg has already done so.

We could also move of of a postgres instance to internal storage such as SQLite or LevelDB so that agg instances can run independently and without operators needing to spin up separate db resources.

voltrevo:
https://github.com/web3well/bls-wallet/issues/415

EDIT: Added to list in description

voltrevo:
@jacque006 

> We could also move of of a postgres instance to internal storage such as SQLite or LevelDB so that agg instances can run independently and without operators needing to spin up separate db resources.

Yeah the storage is local to the aggregator, so I've kinda wanted to just move to SQLite since the early days. That's not a necessary improvement... but I would like to do it. Probably not right now.

I believe the idea for sharing bundles is a future unrelated enhancement to do something similar to the L1 mempool.

voltrevo:
@jacque006

> One thing I know we will need to do is move from storing submitted bundles in memory to the db https://github.com/web3well/bls-wallet/issues/389

I think you pasted the wrong issue link there. Do you know if there's a different one for it? (Also want to talk about this issue.)

jacque006:
Good catch, corrected to https://github.com/web3well/bls-wallet/issues/190

PR #452, opened 2023-01-12
Use geth

voltrevo:
## What is this PR doing?

Switches from `yarn hardhat node` to geth.

Hardhat's node can be very slow, which is a big problem for understanding the performance of the aggregator.

As an added bonus, the aggregator test suite is now about 3x faster 😎.

Additionally, the set up is now all-inclusive by just running `yarn start`. It's completely ephemeral (`ctrl+c` and it's gone and you can immediately start a fresh one), does the `fundDeployer` and deploys the contracts in about 10 seconds.

Since hardhat has other nice features like doing console.log in solidity, the option is still there, and you can do the same kind of thing using `yarn start-hardhat`.

Run the tests.

Contributes to #403.

## Checklist

## Guidelines


contracts/hardhat.config.ts
```diff
@@ -131,7 +131,6 @@ const config: HardhatUserConfig = {
     gethDev: {
       url: `http://localhost:8545`,
       accounts,
-      gasPrice: 0,
```
voltrevo, line 134: Sending transactions hangs with `gasPrice:0` when using geth, probably because geth is unaware of this option.

I believe this was an early hack that we've outgrown anyway. Transactions need to pay gas.

contracts/scripts/fundAccounts.js
```diff
@@ -0,0 +1,11 @@
...
+  "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
+  "0x70997970C51812dc3A010C7d01b50e0d17dc79C8",
+  "0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC",
+  "0x90F79bf6EB2c4f870365E785982E1f101E93b906",
+  "0x15d34AAf54267DB7D7c367839AAf71A00a2C6A65",
+];
+
+for (const to of devAccounts) {
+  eth.sendTransaction({ from: eth.coinbase, to, value: 1e22 });
+}
```
jacque006, line 11: Rather than doing this, I would recommend seeding the geth instance with the Hardhat accounts (https://geth.ethereum.org/docs/fundamentals/backup-restore). This will make it easier to switch between hardhat & geth at will.

contracts/scripts/start-hardhat.sh
```diff
@@ -0,0 +1,20 @@
...
+
+function cleanup {
+  jobs -p | xargs kill
+}
+
+trap cleanup EXIT
+
+yarn hardhat node &
+
+sleep 2
```
jacque006: Node spinup may take longer on older/slower machines.

```suggestion
sleep 5
```

contracts/scripts/start.sh
```diff
@@ -0,0 +1,35 @@
...
+CONTAINER=geth$RANDOM
+
+docker run --name $CONTAINER --rm -p8545:8545 ethereum/client-go:stable \
+  --http \
+  --http.api eth,web3,personal,net \
+  --http.addr=0.0.0.0 \
+  --http.vhosts='*' \
+  --dev \
+  --dev.period=0 \
+  &
```
jacque006, line 24: You could use our [docker-compose](https://github.com/web3well/bls-wallet/blob/main/docker-compose.yml#L20-L35) file to store this info instead.

```suggestion
container=geth${RANDOM} # non-exported bash vars should be lowercase

docker-compose up -d geth
docker rename geth ${container}
```

contracts/scripts/start.sh
```diff
@@ -0,0 +1,35 @@
...
+
+sleep 2
```
jacque006:
```suggestion
sleep 5
```

jacque006 (commented):
- Instead of using bash scripts, you might want to re-write them to be TS scripts run-able from `ts-node`. This will make it easier for other JS/TS devs to modify & maintain them, and it shouldn't be too difficult to still run it all as a background process using child processes.
- It would be good to use `geth` in the client integration tests as well to speed those up. @JohnGuilding can provide additional input on that.

JohnGuilding:
> It would be good to use geth in the client integration tests as well to speed those up. @JohnGuilding can provide additional input on that.

Sounds good to me. FYI @voltrevo we added some [integration tests](https://github.com/web3well/bls-wallet/tree/main/contracts/test-integration) that run off a hardhat and aggregator instance. Initially created for testing the provider but can be expanded beyond that. 

When running locally I've been using `yarn hardhat test ./test-integration/*`. I'm currently working on an issue that will change this to `hardhat test --network gethDev ./test-integration/*` to resolve something unrelated. There is also a GitHub workflow.

So I guess looking at your changes, the [workflow](https://github.com/web3well/bls-wallet/blob/1d35f4e80a5f5316158e7f01e61a721fe62f0103/.github/workflows/clients.yml#L37-L55) and also the [test script](https://github.com/web3well/bls-wallet/blob/1d35f4e80a5f5316158e7f01e61a721fe62f0103/contracts/package.json#L15) would need to change? Correct me if I'm misunderstanding.

contracts/scripts/start-hardhat.sh
voltrevo: `sleep 2` was already a lot slower than was needed in my testing. Any particular reason you feel 5 seconds is the right number?

contracts/scripts/fundAccounts.js
voltrevo: I'm not sure what you're suggesting here, could you be a bit more specific?

Based on the link, it sounds like the suggestion is to get a snapshot with those accounts (somehow? some other method?) and then restore from that snapshot. That seems a lot more complicated, so I'm guessing you meant something else. Perhaps you pasted the wrong link?

contracts/scripts/start.sh
voltrevo: The one in our docker-compose file is a bit different. Most notably, it binds the data directory to persist the state.

This version is ephemeral so it works the same way as `yarn hardhat node`, ie, you get a blank slate each time.

Would you want to change the one in docker-compose to fix this? We'd also need to change `dev.period` which is currently `14` in docker-compose.

contracts/scripts/start.sh
voltrevo: As above

voltrevo:
@JohnGuilding ~These changes don't make it mandatory to use geth - it should continue working as before (unless you ran into some issue?). Those tests may benefit from also switching to geth, but we don't need to do everything in one step.~

EDIT: I glossed over the quote you were responding to... sounds good - https://github.com/web3well/bls-wallet/issues/456.

voltrevo:
> Instead of using bash scripts, you might want to re-write them to be TS scripts run-able from ts-node. This will make it easier for other JS/TS devs to modify & maintain them, and it shouldn't be too difficult to still run it all as a background process using child processes.

@jacque006 I created a PR targeting this branch which implements this: https://github.com/web3well/bls-wallet/pull/457.

Doing this conversion isn't a simple task and I think it's pretty iffy that it is better. I don't feel strongly about it and it's done now, so if you prefer it feel free to apply it. What matters more is the work created by this comment which I really feel wasn't worth it. Not the end of the world ofc, but let's follow-up in retro.

jacque006 (approved):
Discussed comments offline. FYI contract tests are failing

PR #454, opened 2023-01-12
Move config outside docker image

### LINKED ISSUE #453: Move aggregator config outside docker image

voltrevo:
## What is this PR doing?

See title.

Build a docker image and run it using the new `programs/start-docker.sh`.

Resolves #453.

## Checklist

## Guidelines


aggregator/programs/start-docker.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# Path to the network config containing addresses etc. Overrides value from
+# env file.
+# NETWORK_CONFIG_PATH=/home/you/bls-wallet/contracts/networks/local.json
+
+# The ethereum network targeted (used to name the container).
+# NETWORK=local
+
+# Port to use. Overrides value in env file.
+# PORT=3000
```
jacque006: These would great in the aggregator README under existing/new docker section.

aggregator/src/env.ts
```diff
@@ -11,11 +12,26 @@ import nil from "./helpers/nil.ts";
...
+  return requireIntEnv("PORT");
+})();
 
 export const USE_TEST_NET = requireBoolEnv("USE_TEST_NET");
 
-export const NETWORK_CONFIG_PATH = requireEnv("NETWORK_CONFIG_PATH");
+export const NETWORK_CONFIG_PATH = (
+  Deno.env.get("NETWORK_CONFIG_PATH_OVERRIDE") ??
+    requireEnv("NETWORK_CONFIG_PATH")
+);
```
jacque006: Is there a reason we need these overrides? The container should have access to the env vars from the env file, so an agg operator can just update those as need be.

aggregator/programs/start-docker.sh
```diff
@@ -0,0 +1,39 @@
...
+
+docker run \
+  --name aggregator-$VERSION-$NETWORK \
+  -d \
+  -p$PORT:$PORT \
+  --env PORT_OVERRIDE=$PORT \
+  --restart=unless-stopped \
+  --mount type=bind,source="$ENV_PATH",target=/app/.env \
+  --mount type=bind,source="$NETWORK_CONFIG_PATH",target=/app/networkConfig.json \
+  aggregator:git-$VERSION
```
jacque006: Could remove this script and move into our `docker-compose` file.

```yaml
  aggregator:
    container_name: aggregator-${VERSION}-${NETWORK}
    depends_on:
      - postgres
      - geth
    build:
      context: ./aggregator
    env_file: ./aggregator/.env
    ports:
      - ${PORT}:${PORT}
    volumes:
      - ${NETWORK_CONFIG_PATH}:/app/networkConfig.json
```

Note that since the env var file can be set, we shouldn't need to mount the env file.

aggregator/src/env.ts
voltrevo: I can't say we 'need' them, but the reason they're there is to allow you to run the 'same' config in docker. Without these overrides, you need to make a dedicated env file when running docker.

aggregator/programs/start-docker.sh
jacque006: Shelve to new issue

jacque006 (approved):
Discussed dropping overrides & moving script comment to README offline. Other comments can be followed up on as needed.

voltrevo:
@jacque006 Got rid of the overrides, except I still felt and `IS_DOCKER` variable was needed to force `/app/networkConfig.json` in the container. Otherwise you'd be forced to put `NETWORK_CONFIG_PATH=/app/networkConfig.json` in the env file and then specify the host path to your network config separately.

`start-docker.sh` now only requires `VERSION=abc1234` and it's documented in the readme.

Please re-review :).

jacque006:
@voltrevo I think that in the future, we can just get rid of the entire agg build script (and maybe run) and move to a more standardized Docker build pipeline. I am not familiar with best practices for containerizing Deno so may need some research into that.

This work is still a significant improvement and should make it easy for agg operators to run & [configure](https://12factor.net/config). Thanks :+1:

Issue #334, opened 2022-09-12
Deploy ProxyAdmin from within VerificationGateway (within gas limits)


jzaki:
- add a function to VG (eg `initProxyAdmin`) that takes PA bytecode as a param, deploys it, and sets `walletProxyAdmin`
- cant be param in constructor since adds to VG deployment bytecode
- `initProxyAdmin` to be called after VG deployment
- Ensure, wallets can't be deployed when PA is not yet set (should fail internal require of `ERC1967Upgrade._setAdmin`)

Resolves #332

Issue #332, opened 2022-09-12
Bug: ProxyAdmin's separate creation prevents consistent create2 addresses


voltrevo:
Currently we make the ProxyAdmin, then pass its address to VerificationGateway (via create2Deployer), then transfer ownership to the VerificationGateway:
https://github.com/web3well/bls-wallet/blob/0ef2799/contracts/shared/helpers/Fixture.ts#L73-L84

This means the ProxyAdmin address is tied to the EOA account being used, and therefore the VerificationGateway address is also tied to that EOA account. In order to get the same address on multiple networks, it is therefore required to use the same EOA account and also ensure the ProxyAdmin is created with the same nonce on that EOA account. This defeats the whole point of using create2 for us, because, while consistent addresses are possible, operating under these constraints would generate the same addresses even without create2.

ProxyAdmin is [Ownable](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3dac7bb/contracts/access/Ownable.sol), and it needs to be owned by the VerificationGateway. This also means that if we want to deploy multiple VerificationGateways (as updates are made), we also need to deploy multiple ProxyAdmins, which is a headache because ProxyAdmin needs to have a fixed address.

I think the best solution here is for VerificationGateway to deploy the ProxyAdmin. That way it'll inherit its fixed address from the VerificationGateway's fixed address (and fixed creation on startup), and will be owned by the VerificationGateway on creation rather than needing a transfer.

Maybe the reason it's not already like this is due to gas issues? If so, and if that's still a problem, I think we can still solve this but it'll be a bit more complicated.

jzaki:
Good catch. Indeed the VerificationGateway is just under the gas limit, and I look forward consolidating requirements to date into v2, and conform to 4337. In the meantime not having the ProxyAdmin bytecode inside VG made way for other changes.
To fix this deployment issue, a small function can be added to the VG that deploys PA bytecode passed in as a parameter. (create #334 )

Issue #189, opened 2022-04-12
Extract Hubble BLS Solidity to seperate repo


jacque006:
Tracking https://github.com/thehubbleproject/hubble-contracts/issues/663

jacque006:
Moved to https://github.com/getwax/wax/issues/9

PR #464, opened 2023-01-18
Use safe singleton factory

### LINKED ISSUE #394: Update contract deployment to use safe singleton factory
Now that we have safe singleton factory (https://github.com/web3well/bls-wallet/issues/331#issuecomment-1308126246), update our contract deployment code to make use of this, allowing the contracts to be deployed to a deterministic address without relying on special keys.

voltrevo:
## What is this PR doing?

Ports our deployment to use safe's singleton factory.

This has the massive advantage of making the addresses of our deployments predetermined for all networks and for any `tx.origin` used to do the deployment.

In making the switch, I've also fixed the deterministic generation of `VerificationGateway`'s `ProxyAdmin` using the new [`ProxyAdminGenerator`](https://github.com/web3well/bls-wallet/blob/bw-394-use-safe-singleton-factory/contracts/contracts/ProxyAdminGenerator.sol). This was critical to getting a predetermined address for `VerificationGateway` because `proxyAdmin` was a constructor parameter, which affects `VerificationGateway`'s address. This was a pre-existing issue: #332.

Removes:
- `create2Deployer`
- `DEPLOYER_*` variables
- contracts/scripts/0_deploy_precompile_cost_estimator.ts
- contracts/scripts/1_deploy_bls_wallet_contracts.ts
- contracts/scripts/deploy-deployer.ts
- `create2Fixture` (replaced by SafeSingletonFactory class)
- `lazyBlsWallets` (replaced by `fx.createBLSWallet()`) (was always a bit clunky, and was the cause of test failures)

There's also a workaround called `fx.processBundleWithExtraGas`. The automatic gasLimit isn't sufficient for `setPendingBLSKeyForWallet`, I think this has to do with the call depth involved... ie, the gasLimit is enough for the actual amount of gas used, but not enough gas is provided to one of the deeply nested calls since 63/64ths of the gas is provided to each next layer.

Finally, I've added `getCostEstimator()` to `BLS.sol` as a temporary workaround while we wait for Safe's singleton factory on 1337: https://github.com/safe-global/safe-singleton-factory/issues/97. This function just selects the right cost estimator based on the chain id, which means that testing using geth doesn't require editing this file.

- `yarn hardhat test`
- Check the scripts in `scripts/test`

Resolves #394, #332, #334.

## Checklist

## Guidelines


contracts/contracts/lib/hubble-contracts/contracts/libs/BLS.sol
```diff
@@ -39,7 +39,33 @@ library BLS {
...
+     * - 1824 L1 gas ($0.05) to the deployment of BLSOpen (114 bytes)
+     * - 71 L2 gas ($0.000012) to signature verification
+     */
+    function getCostEstimator() internal pure returns (address) {
+        uint256 id;
+        assembly {
+            id := chainid()
+        }
+
+        if (id == 1337) {
```
jacque006: Do we need `chainid` 31337 for hardhat support?

contracts/contracts/lib/hubble-contracts/contracts/libs/BLS.sol
```diff
@@ -39,7 +39,33 @@ library BLS {
...
+            return COST_ESTIMATOR_ADDRESS_1337;
+        }
+
+        return COST_ESTIMATOR_ADDRESS;
+    }
```
jacque006: Rather than modifying this file, I think it would be safer to just have a script/process that updates the `COST_ESTIMATOR_ADDRESS` before contract deploy.

Some risks with this approach:
- There is no guarantee that another EVM network would not use `chainid` 1337. See [this article](https://decrypt.co/109884/ethereum-fork-ethpow-gets-off-to-rocky-start) for an example of when this happened with an existing `chainid` with ETHPoW after The Merge
- Although I have never seen it happen, `chainid` is not guaranteed to be the same block to block. A chain could choose at some point to change its `chainid`. OpenZepplin accounts for this in their [EIP-712 implementation](`COST_ESTIMATOR_ADDRESS`)
- As an attacker, if I could get a chain/network to switch its `chainid` to 1337 for a short time & I deployed a malicious contract to `COST_ESTIMATOR_ADDRESS_1337` I could generate my own `precompileGasCost` and do Bad Things with BLS sig verification. This seems unlikely on most larger/established networks, but I could see it happening on a smaller EVM network/L2, especially if its security/consensus mechanisms are more centralized.

I also could see passing the cost estimator address as an arg to `verifySingle/multiple`, but that is a more significant change. IMO, asides from updating the `COST_ESTIMATOR_ADDRESS`, we should not update/change the BLS libs/contracts taken from Hubble.

Related: https://github.com/web3well/bls-wallet/issues/189

contracts/package.json
```diff
@@ -22,6 +22,9 @@
   "dependencies": {
     "@openzeppelin/contracts": "^4.7.3"
   },
+  "peerDependencies": {
+    "ethers": "^5.0.0"
```
jacque006, line 29: What was this needed for?

contracts/shared/SafeSingletonFactory.ts
```diff
@@ -0,0 +1,213 @@
...
+        to: deployment.signerAddress,
+        value: ethers.BigNumber.from(deployment.gasPrice).mul(
+          deployment.gasLimit,
+        ),
+      })
+    ).wait();
+
+    await (
+      await signer.provider.sendTransaction(deployment.transaction)
+    ).wait();
```
jacque006, line 111: :-1: Nested `await`s (others spots with this, will only comment here :) )

contracts/shared/SafeSingletonFactory.ts
```diff
@@ -0,0 +1,213 @@
...
+
+    const deployTx = {
+      to: this.address,
+      data: ethers.utils.solidityPack(["uint256", "bytes"], [salt, initCode]),
+    };
+
+    try {
+      await (await this.signer.sendTransaction(deployTx)).wait();
+    } catch (error) {
+      if ((error as any).code !== "INSUFFICIENT_FUNDS") {
```
jacque006:
```suggestion
      if ((error as { code: string }).code !== "INSUFFICIENT_FUNDS") {
```

contracts/shared/deploy.ts
```diff
@@ -0,0 +1,99 @@
...
+import SafeSingletonFactory from "./SafeSingletonFactory";
+
+export type Deployment = {
+  singletonFactory: SafeSingletonFactory;
+  precompileCostEstimator: BNPairingPrecompileCostEstimator;
+  blsLibrary: BLSOpen;
+  verificationGateway: VerificationGateway;
+  blsExpander: BLSExpander;
+  aggregatorUtilities: AggregatorUtilities;
+};
```
jacque006: Would be good to consolidate this type with [BlsWalletContracts](https://github.com/web3well/bls-wallet/blob/main/contracts/clients/src/BlsWalletContracts.ts#L23-L31) in clients

jacque006:
@jzaki should also review & determine final approval

contracts/contracts/lib/hubble-contracts/contracts/libs/BLS.sol
voltrevo: I hear you, but I think you are overlooking two things:
1. The 1337 address is still hardcoded to an address which can *only* contain the same logic, due to the deterministic deployment. Messing with the chainid does not allow an attacker to insert a malicious cost estimator.
2. This is not intended to ever deploy to a non-test network.

I wouldn't be happy with modifying the file when using 1337 deployments, because I think that leads to a lot of subtle confusion, as we experienced previously when the committed address here was correct for production but not development.

If you have any contacts at safe, it would be really nice if they [just provided the signature we need](https://github.com/safe-global/safe-singleton-factory/issues/97). It should really only take a few minutes of their time and avoid all this mess.

contracts/package.json
voltrevo: Fixes lint issue.

Related: https://github.com/web3well/bls-wallet/pull/511#discussion_r1102164985

contracts/contracts/lib/hubble-contracts/contracts/libs/BLS.sol
voltrevo: We do, but it's not needed here because safe has provided the signed transaction we need for 31337, so the cost estimator address is at the same location across networks.

contracts/shared/SafeSingletonFactory.ts
voltrevo: Conversation here: https://github.com/web3well/bls-wallet/pull/511#discussion_r1102173652

contracts/contracts/lib/hubble-contracts/contracts/libs/BLS.sol
voltrevo: @jacque006 Moot now, Safe came through and I deleted all this code 🎉.

contracts/contracts/VerificationGateway.sol
```diff
@@ -61,11 +65,11 @@ contract VerificationGateway
     constructor(
         IBLS bls,
         address blsWalletImpl,
-        address proxyAdmin
+        ProxyAdminGenerator proxyAdminGenerator
```
jzaki, line 68: Previously the PA and VG deployment were in the same function, and when needing to deploy our own deployer, this was deterministic. With the change to using the safe singleton factory, the smaller change (preserving audit as much as possible) would have been to change PA in place to be deployed with the safe singleton factory.
Also the design of a separate PA Generator although cleaner in some ways, implies use of PAs more generally in this application which I'd rather lean away from.

contracts/shared/helpers/Fixture.ts
```diff
@@ -159,6 +119,49 @@ export default class Fixture {
...
+
+    return receipt;
+  }
+
+  /**
+   * There seems to be a bug where the automatic gas limit is somtimes not
+   * enough in our testing environment. This method works around that by adding
+   * 50% to the gas estimate.
+   */
+  async processBundleWithExtraGas(bundle: Bundle, overrides: Overrides = {}) {
```
jzaki, line 153: The contract tests were working without this, so is this regarding aggregator tests?

jzaki: The smart contract tests have changed from a simple `call` line to multiple lines breaking things up to sign, adding many more lines to the test functions and making them much less clear.
Since this is seems to be a typescript preference, I'm willing to concede this for v1, but we can work together on more integrated use of client code in v2 contract tests.

contracts/contracts/VerificationGateway.sol
```diff
@@ -18,8 +19,11 @@ is the calling wallet's address.
  */
 contract VerificationGateway
 {
-    /** Domain chosen arbitrarily */
-    bytes32 BLS_DOMAIN = keccak256(abi.encodePacked(uint32(0xfeedbee5)));
+    /**
+     * Chosen arbitrarily
+     * =keccak256(abi.encodePacked(uint32(0xfeedbee5)))
+     */
+    bytes32 BLS_DOMAIN = 0x0054159611832e24cdd64c6a133e71d373c5f8553dde6c762e6bffe707ad83cc;
```
jzaki, line 26: ~~Is this still the arbitrarily generated hash? Arguably this takes more bytes than the opcodes, but VG size is not as important as wallet size. So just curious about the change.~~ Just noticed comment in #527

jzaki (approved):
Apart from the PA point, the rest I'm fine with going into v1. V2 will have a clearer plan from the beginning.

contracts/contracts/VerificationGateway.sol
jzaki: Noticed this was changed in the following commits (#536 ) :heavy_check_mark:

Issue #456, opened 2023-01-16
Use geth in client integration tests


voltrevo:
Follow-up of https://github.com/web3well/bls-wallet/pull/452
https://github.com/web3well/bls-wallet/pull/452#issuecomment-1381744558

jacque006:
Resolved in https://github.com/getwax/bls-wallet/pull/513

PR #399, opened 2022-11-15
BLS signer tests

blakecduncan:
## What is this PR doing?

- Adding tests for a normal signer that interacts with an erc20 and an erc721 contract.
- There is a future issue that will swap out the normal signer with a BlsSigner to make sure everything works. (@JohnGuilding  not sure if you have the issue number?)

## How can these changes be manually tested?
- Tests pass

## Does this PR resolve or contribute to any issues?
https://github.com/web3well/bls-wallet/issues/385

## Checklist
- [x] I have manually tested these changes
- [x] Post a link to the PR in the group chat

## Guidelines
- If your PR is not ready, mark it as a draft
- The `resolve conversation` button is for reviewers, not authors
  - (But add a 'done' comment or similar)

Issue #182, opened 2022-04-04
Add Contract deployment support


kautukkundan:
Aggregator throws error when trying to submit a contract deployment transaction

![Screenshot 2022-04-04 at 4 39 57 PM](https://user-images.githubusercontent.com/23727056/161536136-22cfdd19-19a6-4e97-8348-1be4f56d3d1b.png)

jacque006:
What was the contract deployment needed for in this case?

kautukkundan:
for the BLS account to be able to deploy any new contract. 

In case of contract deployment the tx body looks like 
```ts
{
  from: "<SENDER_ADDR>",
  to: "0x",
  value: "<ETH_VALUE>",
  data: "<DEPLOYMENT_DATA>"
}
```
So the aggregator crashes when the "to" field contains a zero/null value

voltrevo:
Contract changes needed for this because `_performOperation` only implements paid/non-paid actions (to existing addresses):

```solidity
a = op.actions[i];
if (a.ethValue > 0) {
    (success, result) = payable(a.contractAddress).call{value: a.ethValue}(a.encodedFunction);
}
else {
    (success, result) = address(a.contractAddress).call(a.encodedFunction);
}
require(success);
results[i] = result;
```
https://github.com/web3well/bls-wallet/blob/4dae24f/contracts/contracts/BLSWallet.sol#L179-L187

We need a new branch for contract creation, possibly using something like this:
```solidity
assembly {
    addr := create(0,add(code,0x20), mload(code))
    jumpi(invalidJumpLabel,iszero(extcodesize(addr)))
}
```
https://ethereum.stackexchange.com/a/11129

Parking this for now since we've decided not to implement it in V1:
https://discord.com/channels/943612659163602974/963404009266376774/1008748869514969200

jzaki:
Will need an extra flag in `ActionData` to be signed with txs (a `bool` in a struct is stored as a one byte).

PR #467, opened 2023-01-18
Bls provider and signer contracts interaction

### LINKED ISSUE #382: Test BlsProvider and BlsSigner can interact with contracts correctly
We should ensure that the BlsProvider & BlsSigner classes we create can interact with Contracts in the same way that an ethers provider & signer can. This issue is concerned with testing the following:

[Contract Interaction](https://docs.ethers.io/v5/api/contract/#contracts)

There are other ways of interacting with contracts, but these pieces of code involve a provider or signer so we should ensure they work as expected

```ts
// Creates a new instance of a Contract. In the context of ethers, a Contract is an abstraction 
// of a contract (EVM bytecode) that has been deployed to the Ethereum blockchain.
new ethers.Contract( address , abi , signerOrProvider )

// Returns a new instance of the Contract, but connected to providerOrSigner.
contract.connect( providerOrSigner ) ⇒ Contract

// If a provider was provided to the constructor, this is that provider. If a signer was provided 
// that had a Provider, this is that provider.
contract.provider ⇒ Provider

// If a signer was provided to the constructor, this is that signer.
contract.signer ⇒ Signer
```

[Contract Factory](https://docs.ethers.io/v5/api/contract/contract-factory/#ContractFactory)
```ts
// Creates a new instance of a ContractFactory for the contract described by the interface and bytecode initcode.
new ethers.ContractFactory( interface , bytecode [ , signer ] )

// Consumes the output of the Solidity compiler, extracting the ABI and bytecode from it, allowing for 
// the various formats the solc compiler has emitted over its life.
ContractFactory.fromSolidity( compilerOutput [ , signer ] ) ⇒ ContractFactory

// Returns a new instance of the ContractFactory with the same interface and bytecode, but with a different signer.
contractFactory.connect( signer ) ⇒ ContractFactory

// The Signer (if any) this ContractFactory will use to deploy instances of the Contract to the Blockchain.
contractFactory.signer ⇒ Signer

// Uses the signer to deploy the Contract with args passed into the constructor and returns a Contract 
// which is attached to the address where this contract will be deployed once the transaction is mined.
contractFactory.deploy( ...args [ , overrides ] ) ⇒ Promise< Contract >
```

JohnGuilding:
## What is this PR doing?
- Replaces regular signer with bls signer following on from #399. This included renaming `Signer.test.ts` and moving it to the `test-integration` directory.
- Completes #382 
    Adds tests for:
    - `new ethers.Contract( address , abi , signerOrProvider )`
    - `contract.connect( providerOrSigner ) ⇒ Contract`
    - `contract.provider ⇒ Provider`
    - `contract.signer ⇒ Signer`
    - `new ethers.ContractFactory( interface , bytecode [ , signer ] )`
   Could not test `contractFactory.deploy( ...args [ , overrides ] ) ⇒ Promise< Contract >`. This is a known issue slated for future development #182
- Runs integration tests using gethDev network instead of hardhat
- Moves non-integration tests to clients. These tests do not require node or agg instance to run successfully. See additions in `contracts/clients/test/*`
- Updates ethers to ^5.7.2 in `./contracts` to match `./clients`
- Updates provider receipt polling retries to overcome issue where workflow tests were timing out. See https://github.com/web3well/bls-wallet/actions/runs/3957500743/jobs/6777946250. The failing tests in this workflow take around 30 seconds locally. This will be improved in #456, we may also want to improve our polling mechanism in the future as it is quite simple.
- Adds README instructions on how to run integration tests.

Running integration tests

#382 

## Checklist

## Guidelines


kautukkundan (approved):
LGTM! 🔥

contracts/clients/package.json
```diff
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "@thehubbleproject/bls": "^0.5.1",
-    "ethers": "5.5.4",
+    "ethers": "^5.7.2",
```
blakecduncan, line 25: Maybe don't have to do this as part of this PR but when we update the bls-wallet-client version in the aggregator and the extension, we might want to update ethers there too.

contracts/test-integration/BlsProviderContractInteraction.test.ts
```diff
@@ -0,0 +1,70 @@
...
+        mockERC20.interface,
+        blsProvider,
+      );
+      const balance = await erc20.balanceOf(recipient);
+
+      expect(erc20.provider).to.equal(blsProvider);
+      expect(balance).to.equal(tokenSupply);
+    });
+  });
+});
```
blakecduncan, line 70: Just curious - I'm not sure if I'm missing this but do we not want to keep mockERC721 tests?

blakecduncan (commented):
Just saw this is already merged but had two small questions

contracts/test-integration/BlsProviderContractInteraction.test.ts
JohnGuilding: My thinking here was that we're testing the bulk of the contract interaction in `BlsSignerContractInteraction.test.ts`. Here, I just wanted to test that the **provider specifically** could be used for view calls e.g. `balanceOf`. So just included ERC20 mock here as that was easiest and didn't see much value for the ERC721 equivalents 

Open to adding the ERC721 mocks here if we think that's a good idea though. Don't have a strong opinion either way

blakecduncan: Ohh no that's perfect.  I just missed that file when scrolling through the PR so figured I'd ask.  Looks good to me 👍

contracts/clients/package.json
JohnGuilding: Yeah agreed, makes sense to use the same version everywhere. Do you think we should do that as a follow up to this change, or just when we update the bls-wallet-client versions in those places?

PR #471, opened 2023-01-19
Fix test integration

### LINKED ISSUE #470: Fix contracts/test-integration
Currently broken: #469.

I suspect this happened due to a change in the aggregator which didn't trigger these tests to run.

Part of this change should also expand the condition for running these tests. It's currently just `contracts/clients/**`, but the tests depend on the aggregator, contracts, and the test definitions themselves.

voltrevo:
## What is this PR doing?

Fixes test-integration by including the missing environment variable in `aggregator/.env.test`.

- Splits out `integration.yml` and makes these tests run on more files that affect these tests, to avoid this problem going forward
- Handles aggregator logs normally on CI instead of using artifact
- Fixes failing test caused by unhandled `nil` case when calling `lookupBundle`

Check CI passes.

Resolves #470.

## Checklist

## Guidelines


jacque006 (approved):
@voltrevo Will add back in agg log artifact in follow up as they are getting cutoff in GH actions log

Issue #435, opened 2023-01-04
Update values returned in bundle receipt to more closely match ethers transaction response


JohnGuilding:
We should update the bundle receipt to return additional information about the transaction. This means updating the `lookupReceipt()` method in the `BundleService`.

The bundle receipt currently looks like this:

```ts
export type BundleReceipt = {
  transactionIndex: string;
  transactionHash: string;
  bundleHash: string;
  blockHash: string;
  blockNumber: string;
};
```

As part of the provider work, we need to return an ethers transaction receipt that looks like this:

```ts
export interface TransactionReceipt {
  to: string;
  from: string;
  contractAddress: string,
  transactionIndex: number,
  root?: string,
  gasUsed: BigNumber,
  logsBloom: string,
  blockHash: string,
  transactionHash: string,
  logs: Array<Log>,
  blockNumber: number,
  confirmations: number,
  cumulativeGasUsed: BigNumber,
  effectiveGasPrice: BigNumber,
  byzantium: boolean,
  type: number;
  status?: number
};
```

The transaction receipt information is mostly derived from the bundle receipt, but there a several pieces of information missing. We can improve this by updating the `lookupReceipt()` method in the `BundleService`. 

As part of the BLS provider/signer work, we have overridden the ethers getTransactionReceipt() method which calls `_getTransactionReceipt()`. A developer would use this either by explicitly calling `provider.getTransactionReceipt()`, or more commonly, by calling it implicitly with `await tx.wait()` after sending a transaction.

We should update `_getTransactionReceipt()` so that it returns a more accurate transaction receipt in the `BlsProvider`.

Note: There are a few other TODOs that are related to this issue that should be completed as part of this work. In `bls-wallet`, search for "// TODO: bls-wallet #412" to find them

## Acceptance criteria

- The `BundleReceipt` type returned from `lookupReceipt()` in `./clients/src/Aggregator.ts` has been updated to include values from the ethers `TransactionReceipt`.
- The `lookupReceipt()` method at `./aggregator/src/app/BundleService.ts` returns the values needed for the `lookupReceipt()` function in `./clients/src/Aggregator.ts`.
```ts
// e.g.
return {
 bundleHash: hash,
 to: receipt.to,
 from: receipt.from,
...
```
- Add test for `lookupReceipt()` in `BundleServiceSubmitting.test.ts`
- The `BlsProvider` `_getTransactionReceipt()` method has been updated to return the updated bundle receipt
[...]

PR #480, opened 2023-01-23
Update bundle receipt to match ethers transaction response

JohnGuilding:
## What is this PR doing?
1. Updates the bundle receipt type returned from the aggregator to closer match an ethers response type.
2. Update test in `BundleServiceSubmitting.test.ts` to ensure `receiptFromBundle` works and returns the bundle hash.
3. Update types in the client module (`Aggregator.ts`)
4. Update the provider `_getTransactionReceipt` method to return the updated type.
5. Update provider tests.
6. Skip signer test that was causing issues.

1. Hit lookupReceipt endpoint and observe updated bundle receipt
2. Send a local transaction with Quill to ensure the changes have not broken how Quill retrieves transaction receipts.

#435 

## Checklist

## Guidelines


contracts/test-integration/BlsProvider.test.ts
```diff
@@ -201,36 +201,58 @@ describe("BlsProvider", () => {
...
       transactionResponse.hash,
       1,
-      10,
+      20,
     );
 
     // Assert
-    // TODO: bls-wallet #412 Update values returned in bundle receipt to more closely match ethers transaction response
+    const expectedBlockNumber = await blsProvider.getBlockNumber();
     expect(transactionReceipt).to.be.an("object").that.deep.includes({
```
JohnGuilding, line 216: These assertions are more verbose but feel like the tradeoff is worth it unless anyone has suggestions on making them more concise.

If someone edits the return value in the BundleService, either by removing a property or changing property types, the test will fail. Note the test does not fail if someone adds an invalid property to the bundle receipt [see comment](https://github.com/web3well/bls-wallet/pull/480/files#r1084232169)

contracts/clients/src/Aggregator.ts
```diff
@@ -38,12 +39,18 @@ export type BundleReceiptError = {
...
+export type BlsWallet = {
   bundleHash: string;
-  blockHash: string;
-  blockNumber: number;
+};
+
+/**
+ * The bundle receipt returned from a BLS Wallet Aggregator instance. It is a combination of an ethers {@link ContractReceipt} and a {@link BlsWallet} type.
+ */
+export type BundleReceipt = ContractReceipt & {
```
JohnGuilding: Been thinking about validation in this file. I tried adding a property to the return type for `receiptFromBundle` - `invalidProperty: "invalid property”,` and it was successfully propagated to the provider. When we’re using our own aggregator, we have confidence the bundleReceipt will have certain properties, and as mentioned in [this](https://github.com/web3well/bls-wallet/pull/480/files#r1084229320) comment, our integration tests will pick up certain changes to the returned bundleReceipt. But what if a third party developer is using the client module/provider and decides to use their own aggregator. In this case, invalid responses could reach client module code. 

My question is, should we validate incoming responses so that we don’t allow invalid data into the client module? Maybe a future issue?

contracts/clients/src/Aggregator.ts
```diff
@@ -38,12 +39,18 @@ export type BundleReceiptError = {
   submitError: string | undefined;
 };
 
-export type BundleReceipt = {
-  transactionIndex: number;
-  transactionHash: string;
+/**
+ * The BLS Wallet specific values in a {@link BundleReceipt}.
+ */
+export type BlsWallet = {
```
JohnGuilding: Can't decide if I like this name, any suggestions welcome. 

Few random ideas:
`BlsWalletReceipt`
`BlsTransactionReceipt`
`BlsBundleReceipt`
`BundleTransactionsReceipt`
`Bundles`
`BlsBundles`

contracts/test-integration/BlsSignerContractInteraction.test.ts
```diff
@@ -217,17 +217,17 @@ describe("Signer contract interaction tests", function () {
     });
 
     // TODO: Investigate why safeMint() fails with a BLS wallet address. Note - it passes in isolation
-    it("safeMint() call fails with BLS wallet address", async function () {
+    it.skip("safeMint() call fails with BLS wallet address", async function () {
```
blakecduncan, line 220: Do we still want this skipped?

contracts/clients/src/Aggregator.ts
blakecduncan: Will this updated type break anything in the extension?

blakecduncan (changes requested):
Nice, this looks good!

I think this is the process that is typically used for client updates.  But the only additional thing I would do is push up an experimental version of the client library.  That way we can update the client version in the aggregator and the extension and make sure everything is working together.  Then I think in a future PR if everything is working as expected we can create a new official release.

contracts/test-integration/BlsSignerContractInteraction.test.ts
JohnGuilding: It seems to be passing in isolation but then failing when run with other tests. Thought it would be better to skip it in case it causes issues like some other integration tests have, then implement a fix for it. I can add a follow up issue to look into it so it's tracked.

Let me know if you have any additional thoughts

contracts/clients/src/Aggregator.ts
JohnGuilding: Tested the extension locally and is able to get the transaction receipt successfully. The transaction hashes lined up with what was logged in the aggregator logs.

Thinking about it, I might check it works on testnet as well as locally - then I can check the transaction hash works on etherscan

contracts/test-integration/BlsSignerContractInteraction.test.ts
JohnGuilding: Follow up #484

blakecduncan: Nope, that sounds good!  I just wanted to make sure it wasn't an accidental add.

contracts/clients/src/Aggregator.ts
JohnGuilding: Just tested successfully via quill running a local aggregator pointing at arbitrum goerli.

The extension does not query the hash via `bundleReceipt.bundleHash`, so it still works (if it did query the hash this way, it would break).

It appears to [save the bundle hash here](https://github.com/web3well/bls-wallet/blob/b6fc2164a3a6b8785c88f18079e38b80fd7caeba/extension/source/background/AggregatorController.ts#L91-L94) so it doesn't have to query it.

It also uses the [returned bundle hash here](https://github.com/web3well/bls-wallet/blob/b6fc2164a3a6b8785c88f18079e38b80fd7caeba/extension/source/Home/Wallet/Wallets/SendDetail/SendDetail.tsx#L111) to get the transaction receipt.

Can't see anywhere else it could break to my knowledge.

contracts/clients/src/Aggregator.ts
blakecduncan: Yeah I agree, I think it would be good to have the word receipt in the name.  Maybe something like BlsBundleReceipt sounds good 👍

JohnGuilding: BlsBundleReceipt sounds good to me, will update

contracts/clients/src/Aggregator.ts
blakecduncan: I think that is probably a good future issue.  Just my opinion but I think validating the contract receipt right now might be a little scope creep.  This is probably something we can handle if people start using other aggregators that don't follow our pattern here.

blakecduncan (approved):
thanks for the updates 👍

Issue #442, opened 2023-01-10
Add function to generate a random BLS private key


jacque006:
*As a* web3 developer
*I would like to* be able to generate a random private key within the BN254 curve
*So that I can* be sure the private keys I generate have sufficient entropy and are compatible

### Notes
See @blakecduncan 's [implementation for the extension](https://github.com/web3well/bls-wallet/blob/main/extension/source/helpers/randFr.ts) as well as [randFr](https://github.com/thehubbleproject/hubble-bls/blob/master/src/mcl.ts#L186-L191) from the signer library we use.

Right now, if a private key is provided to a BLS Wallet that is outside of that curve, we hash it to fit inside. See https://github.com/thehubbleproject/hubble-bls/pull/28#issuecomment-1066123893 for some discussion around why this could be undesirable.

PR #486, opened 2023-01-25
Add client recovery methods

blakecduncan:
## What is this PR doing?
Adding the logic used recover an instant wallet to quill to the bls client module.  There's probably room to add some more recovery functionality.  This is just the logic that is currently used to recover from an Instant wallet to Quill.

Added tests to the contracts tests to make sure it works.  Also, you could try recovering an instant wallet.

https://github.com/web3well/bls-wallet/issues/220
https://github.com/web3well/bls-wallet/issues/442

## Checklist

## Guidelines


contracts/clients/src/signer/getRandomBlsPrivateKey.ts
```diff
@@ -0,0 +1,5 @@
+import { mcl } from "@thehubbleproject/bls";
+
+export default (): string => {
+  return `0x${mcl.randFr().serializeToHexStr()}`;
```
blakecduncan, line 5: I originally put this here because we know that `mcl` has been initialised inside the signer directory.  However, logically I'm thinking now that I should pull this out of the signer.  Curious to hear your thoughts.  Right now if you want to call this you have to create a wallet wrapper first like so.  You probably shouldn't have to create a wallet to get a new PK.

```
const wallet = await this.BlsWalletWrapper(pk);
const newPk = wallet.getRandomBlsPrivateKey();
```

contracts/clients/src/BlsWalletWrapper.ts
```diff
@@ -253,6 +253,76 @@ export default class BlsWalletWrapper {
     return this.blsWalletSigner.getPublicKeyStr(this.privateKey);
   }
 
+  getRandomBlsPrivateKey(): string {
+    return this.blsWalletSigner.getRandomBlsPrivateKey();
+  }
+
+  async getBundleSetRecoveryHash(
```
JohnGuilding: ~To check my understanding, this method sets the recovery hash and then returns that bundle?~

Was originally thrown off by the naming here but not convinced there's anything better. 

1. `getSetRecoveryHashBundle`/`getRecoverWalletBundle` reads more like a sentence but not a fan of "get" and "set" next to each other for the first method.
2. `createSetRecoveryHashBundle`/`createRecoverWalletBundle` reads better than "get" prefix to me, but prefer the "get" prefix semantically
3. `signSetRecoveryHashOperation`/`signRecoverWalletOperation` think this reads well but using bundle instead of operation makes more sense to me
4. `setRecoveryHash`/`recoverWallet` is misleading as the method is only creating the bundle and not executing it

Happy with to go with original (or one of 4 options if you think any make more sense) as I've spent far too long thinking about this lol

contracts/clients/src/BlsWalletWrapper.ts
```diff
@@ -316,4 +386,12 @@ export default class BlsWalletWrapper {
       );
     }
   }
+
+  private static saltToBytes32String(salt: string): string {
+    try {
+      return ethers.utils.formatBytes32String(salt);
+    } catch (e) {
+      throw new Error("Error convirting salt string to bytes 32 string.");
```
JohnGuilding: Making an assumption here, but is this error handling to ensure we have a more helpful error message when an invalid salt is passed in? 

This is more of a question than a suggestion, but if so, would it be helpful to also surface the ethers error too as part of the error message? 

Also nitpick but while I'm here, noticed a tiny typo in the error message :)

contracts/clients/src/signer/getRandomBlsPrivateKey.ts
JohnGuilding: Agree creating a new wallet to get a random private key is not optimal, unsure on the best place to put this. But could it be a static method somewhere? Similar to how we have static methods for an address on BlsWalletWrapper.

extension/source/helpers/randFr.ts
```diff
@@ -1,11 +0,0 @@
-import * as mcl from 'mcl-wasm';
-import { hexlify, randomBytes } from 'ethers/lib/utils';
-
-export default async function randFr(): Promise<mcl.Fr> {
```
JohnGuilding, line 4: Are there any differences between this function, and the `randFr` function used in `getRandomBlsPrivateKey`?

contracts/clients/src/BlsWalletWrapper.ts
```diff
@@ -253,6 +253,76 @@ export default class BlsWalletWrapper {
...
 
+  getRandomBlsPrivateKey(): string {
+    return this.blsWalletSigner.getRandomBlsPrivateKey();
+  }
+
+  async getBundleSetRecoveryHash(
+    salt: string,
+    recoverWalletAddress: string,
+  ): Promise<Bundle> {
+    const saltBytes32String = BlsWalletWrapper.saltToBytes32String(salt);
```
JohnGuilding: What was the reason for changing references of this from `saltHash`/`recoverySaltHash` to `saltBytes32String`?

Don't have a strong opinion, but wondering if `saltHash` would read better than having the type in the name?

contracts/test/recovery-test.ts
```diff
@@ -133,6 +137,55 @@ describe("Recovery", async function () {
...
+      fx.verificationGateway,
+    );
+    const recoveryBundleTxn = await fx.verificationGateway.processBundle(
+      recoveryBundle,
+    );
+    await recoveryBundleTxn.wait();
+
+    const newHash = wallet4.blsWalletSigner.getPublicKeyHash(newPrivateKey);
+    expect(await vg.hashFromWallet(wallet4.address)).to.eql(newHash);
+    expect(await vg.walletFromHash(newHash)).to.eql(wallet4.address);
```
JohnGuilding, line 186: General recovery question - my understanding of this scenario is that wallet 3 should not be able to send transactions after being recovered to wallet 4, is that correct?

If so, do you see value in a test that checks wallet 3 cannot send transactions after being recovered? That might be covered by existing assertions/not be desirable, curious on your thoughts here.

JohnGuilding (approved):
Looks good!

All comments are minor/non-blocking so approving

contracts/clients/src/BlsWalletWrapper.ts
blakecduncan: Yeah I spent some time thinking about this.  I ended up going with what I had cause I figured `getBundle{FUNCTION_NAME}` could be a good naming pattern. 

I think your first one might be the most readable though so I might switch to that.

contracts/clients/src/BlsWalletWrapper.ts
blakecduncan: No particular reason.  I think in a lot of my code I called it `saltBytes32String` to copy the function name and Kautuk called it saltHash in other places.  I might switch to saltHash

contracts/clients/src/signer/getRandomBlsPrivateKey.ts
blakecduncan: Yeah the more that I think about it, I'm gonna move it.

contracts/clients/src/BlsWalletWrapper.ts
blakecduncan: Yeah, this is a good shout.  After thinking about it, this private method is unnecessary and we can just go with the default error

extension/source/helpers/randFr.ts
blakecduncan: No they are meant to do the same thing.  That randFr function comes from the hubble library.  I added this file a while back by copying the logic from the hubble project

contracts/clients/src/BlsWalletWrapper.ts
```diff
@@ -120,6 +127,15 @@ export default class BlsWalletWrapper {
     return this.ExpectedAddress(verificationGateway, pubKeyHash);
   }
 
+  static async getRandomBlsPrivateKey(): Promise<string> {
+    await mcl.init(mcl.BN_SNARK1);
```
blakecduncan: Hey @jacque006, I was curious to get your thoughts on this.  I want the `getRandomBlsPrivateKey` function to be static but I need to make sure `.init` has been called on mcl.  Looks like the only way we can guarantee that right now is if I put this code in the signer (I originally did that and then pulled it out because I didn't think it makes sense).  Do you think it's an issue if I import the mcl library in this file?

extension/source/helpers/randFr.ts
JohnGuilding: gotcha cheers

contracts/clients/src/BlsWalletWrapper.ts
jacque006: I would try to keep any logic like this in the signer dir, and I would use `@thehubbleproject/bls` over a direct `mcl-wasm` import.

You should be able to do something like this:

`./contracts/clients/signers/getRandomPrivateKey.ts`
```typescript
import { mcl } from "@thehubbleproject/bls";

export default function getRandomPrivateKey(): Promise<string> {
    await mcl.init();
    return `0x${mcl.randFr().serializeToHexStr()}`;
}
```

Then add as an export to `./contracts/clients/signers/index.ts` and consume here.

jacque006: You might run into issues running `mcl.init` twice. If that happens will have to try another way.

blakecduncan: This worked for me.  Thanks, @jacque006 !

contracts/test/recovery-test.ts
blakecduncan: So in this situation, wallet 3 is actually the trusted wallet for wallet 4.  I've been using the term trusted wallet recently because I think that makes more sense then what I've been calling it.  In this situation, we can assume that it was actually wallet 4 that lost it's private key and needs to recover to a new one.  Wallet 3 is the trusted wallet that can call the recover wallet function to swap the PK for wallet 4.  

After the recovery is done, wallet 4's old private key should no longer be able to send txs.  But the new private key that now maps to wallet 4's public key should be able to send transactions for that hash.  That's why here I check that the public key hash for the new private key is the same as the hash that maps to wallet 4's address.  If that's the case, the new private key can send txs for that address.

With regard to the test.  We already have a decent amount of recovery tests in here which is why I just wanted to add a test to check that the bundle created from the new functions works.  I just looked though and it doesn't look like we have a test that explicitly checks that the old private key doesn't work anymore.  So I think that would be a valuable test.

I could add an extra line to this test to check that the old private key no longer maps to any public key hash.  Which would mean it can't send any txs.  When a recovery happens we just zero out the hash that the private key maps to in the mapping stored in the verification gateway. So that private key can't do anything anymore.

JohnGuilding: Yeah think I was confused about how this was working. I've got a good grasp now, thanks for clearing that up!

Trusted wallet as a term also makes sense. Wondering if renaming `blsWallet3` to something like `trustedBlsWallet` would improve readability?

Makes sense in regards to additional test case.

Issue #446, opened 2023-01-10
Add gas limit (audit fix 5)


jzaki:

jacque006:
Completed in https://github.com/web3well/bls-wallet/pull/487

PR #487, opened 2023-01-27
Bw 446 add gas limit (Audit Issue 5)

jzaki:
## What is this PR doing?
Adding a gas limit to operations, and estimation functions in client code.

#446 

## Checklist

## Guidelines

### DIFF
--- a/contracts/clients/src/BlsWalletWrapper.ts
+++ b/contracts/clients/src/BlsWalletWrapper.ts
@@ -225,16 +225,21 @@ export default class BlsWalletWrapper {
     return await walletContract.nonce();
   }
 
-  /** Returns the same operation but with a gas estimate. */
-  async opWithGasEstimate(operation: Operation): Promise<Operation> {
+  /** Sign an operation with an estimate of the gas required. */
+  async signWithGasEstimate(
+    operationOmittingGas: Omit<Operation, "gas">,
+  ): Promise<Bundle> {
+    const operation: Operation = {
+      ...operationOmittingGas,
+      gas: this.blockGasLimit,
+    };
     const bundle = this.sign(operation);
-    operation.gas = this.blockGasLimit;
     const gatewayAddress = await this.walletContract.trustedBLSGateway();
-    // const gatewaySigner = new VoidSigner(gatewayAddress);
+    // Estimate the gas required for the operation when called by the gateway
     operation.gas = await this.walletContract
       .connect(gatewayAddress)
       .estimateGas.performOperation(bundle.operations[0]);
-    return operation;
+    return this.sign(operation);
   }
 
   /** Sign an operation, producing a `Bundle` object suitable for use with an aggregator. */
--- a/contracts/clients/src/BlsWalletWrapper.ts
+++ b/contracts/clients/src/BlsWalletWrapper.ts
@@ -221,6 +221,20 @@ export default class BlsWalletWrapper {
     return await walletContract.nonce();
   }
 
+  /** Returns the same operation but with a gas estimate. */
+  async opWithGasEstimate(operation: Operation): Promise<Operation> {
+    const bundle = this.sign(operation);
+    operation.gas = (
+      await this.walletContract.provider.getBlock("latest")
+    ).gasLimit;
+    const gatewayAddress = await this.walletContract.trustedBLSGateway();
+    // const gatewaySigner = new VoidSigner(gatewayAddress);
+    operation.gas = await this.walletContract
+      .connect(gatewayAddress)
+      .estimateGas.performOperation(bundle.operations[0]);
+    return operation;
+  }
+
   /** Sign an operation, producing a `Bundle` object suitable for use with an aggregator. */
   sign(operation: Operation): Bundle {
     return this.blsWalletSigner.sign(
--- a/contracts/clients/src/BlsWalletWrapper.ts
+++ b/contracts/clients/src/BlsWalletWrapper.ts
@@ -27,6 +27,7 @@ type SignerOrProvider = ethers.Signer | ethers.providers.Provider;
  */
 export default class BlsWalletWrapper {
   public address: string;
+  public blockGasLimit: BigNumber = BigNumber.from(0);
   private constructor(
     public blsWalletSigner: BlsWalletSigner,
     public privateKey: string,
@@ -149,6 +150,9 @@ export default class BlsWalletWrapper {
       privateKey,
       await BlsWalletWrapper.BLSWallet(privateKey, verificationGateway),
     );
+    blsWalletWrapper.blockGasLimit = (
+      await provider.getBlock("latest")
+    ).gasLimit;
 
     return blsWalletWrapper;
   }
@@ -224,9 +228,7 @@ export default class BlsWalletWrapper {
   /** Returns the same operation but with a gas estimate. */
   async opWithGasEstimate(operation: Operation): Promise<Operation> {
     const bundle = this.sign(operation);
-    operation.gas = (
-      await this.walletContract.provider.getBlock("latest")
-    ).gasLimit;
+    operation.gas = this.blockGasLimit;
     const gatewayAddress = await this.walletContract.trustedBLSGateway();
     // const gatewaySigner = new VoidSigner(gatewayAddress);
     operation.gas = await this.walletContract
--- a/contracts/clients/src/BlsWalletWrapper.ts
+++ b/contracts/clients/src/BlsWalletWrapper.ts
@@ -32,6 +32,7 @@ export default class BlsWalletWrapper {
     public blsWalletSigner: BlsWalletSigner,
     public privateKey: string,
     public walletContract: BLSWallet,
+    public defaultGatewayAddress: string,
   ) {
     this.address = walletContract.address;
   }
@@ -149,6 +150,7 @@ export default class BlsWalletWrapper {
       blsWalletSigner,
       privateKey,
       await BlsWalletWrapper.BLSWallet(privateKey, verificationGateway),
+      verificationGateway.address,
     );
     blsWalletWrapper.blockGasLimit = (
       await provider.getBlock("latest")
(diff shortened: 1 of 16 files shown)


contracts/clients/src/BlsWalletWrapper.ts
```diff
@@ -148,7 +150,11 @@ export default class BlsWalletWrapper {
       blsWalletSigner,
       privateKey,
       await BlsWalletWrapper.BLSWallet(privateKey, verificationGateway),
+      verificationGateway.address,
     );
+    blsWalletWrapper.blockGasLimit = (
+      await provider.getBlock("latest")
+    ).gasLimit;
```
jacque006, line 157: With EIP-1559, gas limit can vary between blocks from 15 - 30 million gas. Not sure if this applies to current optimistic rollups, but if it can be variable we should also update this [every block](https://docs.ethers.org/v5/api/providers/provider/#Provider--events).

In constructor:
```typescript
// since this is async, we also need to drop any earlier block queries that are still processing
// and use the latest so older gas limit values don't overwrite new ones.
provider.on("block", async (blockNumber) => {
    const { gasLimit } = await provider.getBlock(blockNumber);
    this.blockGasLimit = gasLimit;
});
// We may need to add a cleanup/stop method to wrapper class to prevent event listener leaks.
```

contracts/clients/src/BlsWalletWrapper.ts
```diff
@@ -221,6 +227,41 @@ export default class BlsWalletWrapper {
     return await walletContract.nonce();
   }
 
+  /** Sign an operation with an estimate of the gas required. */
+  async signWithGasEstimate(
+    operation: Omit<Operation, "gas">,
+    overhead = 0,
+  ): Promise<Bundle> {
+    let gas = await this.estimateGas(operation);
+    gas = gas.add(gas.div(10000).mul(Math.ceil(overhead * 10000)));
```
jacque006, line 236: What are the `10000`'s here?

contracts/contracts/VerificationGateway.sol
```diff
@@ -296,12 +296,12 @@ contract VerificationGateway
             IWallet wallet = getOrCreateWallet(bundle.senderPublicKeys[i]);
 
             // check nonce then perform action
-            if (bundle.operations[i].nonce == wallet.nonce()) {
+            if (bundle.operations[i].nonce == wallet.nonce{gas:20000}()) {
```
jacque006, line 299: If someone upgrades to a different wallet implementation, its `nonce` function must be runable with 20k gas.
I would add a comment to `IWallet.nonce` to that effect.

contracts/contracts/VerificationGateway.sol
```diff
@@ -340,6 +340,30 @@ contract VerificationGateway
...
+     * This has two important differences over standard estimateGas methods:
+     *   1. It does not include gas for calldata, which has already been paid
+     *   2. It works for wallets which don't yet exist
+     */
+    function measureOperationGas(
+        uint256[BLS_KEY_LEN] memory publicKey,
+        IWallet.Operation calldata op
+    ) external returns (uint) {
+        // Don't allow this to actually be executed on chain. Static calls only.
+        require(msg.sender == address(0));
```
jacque006, line 356: Prefer `requires` with human readable error messages.
```suggestion
        require(msg.sender == address(0), "VG: read only");
```

contracts/contracts/VerificationGateway.sol
```diff
@@ -340,6 +340,30 @@ contract VerificationGateway
...
+
+        IWallet wallet = getOrCreateWallet(publicKey);
+
+        uint gasBefore = gasleft();
+        wallet.performOperation(op);
+        uint gasUsed = gasBefore - gasleft();
```
jacque006, line 362: The rest of our Solidity codebase specifies the precision of the uint, usually `uint256`. Is there a reason that is not done here?

contracts/shared/helpers/Fixture.ts
```diff
@@ -159,17 +160,27 @@ export default class Fixture {
     );
   }
 
+  async createBLSWallet(): Promise<BlsWalletWrapper> {
+    return BlsWalletWrapper.connect(
+      `0x${Math.floor(Math.random() * 0xffffffff).toString(16)}`,
```
jacque006, line 165: Prefer non-random/deterministically created private keys to aid in test reproducibility for triaging/debugging.

contracts/clients/src/BlsWalletWrapper.ts
```diff
@@ -221,6 +227,41 @@ export default class BlsWalletWrapper {
...
+
+    return this.sign({ ...operation, gas });
+  }
+
+  /** Estimate the gas needed for an operation. */
+  async estimateGas(operation: Omit<Operation, "gas">) {
```
jacque006, line 242:
```suggestion
  async estimateGas(operation: Omit<Operation, "gas">): Promise<BigNumberish> {
```

contracts/clients/src/BlsWalletWrapper.ts
```diff
@@ -221,6 +227,41 @@ export default class BlsWalletWrapper {
     return await walletContract.nonce();
   }
 
+  /** Sign an operation with an estimate of the gas required. */
+  async signWithGasEstimate(
+    operation: Omit<Operation, "gas">,
+    overhead = 0,
```
jacque006, line 233: It's not clear what this is for just based on the function sig. A docstring similar to other public functions in this class would be helpful.

jacque006 (approved):
Unless I missed it, would also be good to have a test case that expects a failure when an operation exceeds its specified gas limit.

Don't see anything that holds this up from merging to `contract-updates` :+1:

contracts/clients/src/BlsWalletWrapper.ts
voltrevo: @jacque006 I think that's wrong. It's not the block gas limit that varies in that kind of range but the block gas *used*. The limit is currently fixed at 30m. I wrote a script to look at some random samples to help confirm:

```js
(async () => {
    const latestBlock = await ethereum.request({ method: 'eth_getBlockByNumber', params: ['latest', false] });
    console.log(latestBlock);

    const rangeStart = 15537393; // First post-merge block
    const rangeEnd = 16530455; // Current latest

    for (let i = 0; i < 100; i++) {
        const sampleBlockNo = rangeStart + Math.floor((rangeEnd-rangeStart) * Math.random());
        const block = await ethereum.request({ method: 'eth_getBlockByNumber', params: ['0x'+sampleBlockNo.toString(16), false] });

        console.log(`Sample ${i} | Block ${sampleBlockNo} | Used: ${Number(block.gasUsed)} | Limit: ${Number(block.gasLimit)}`);
    }
})();
```

Screen Shot 2023-02-01 at 11 46 44 am

contracts/clients/src/BlsWalletWrapper.ts
voltrevo: @jacque006 This is approximating `gas + gas * overhead`. The reason to divide and multiply is to account for integer arithmetic - `overhead` is expected to be something like `0.1`, but we need to work with integers, so we multiply `overhead` by 10000 to get a whole number (after rounding) and compensate by dividing the other side of the product by 10000.

contracts/contracts/VerificationGateway.sol
voltrevo: Will try but we're right up against the 30m gas limit when deploying `VerificationGateway`. Strings are expensive 😔.

contracts/contracts/VerificationGateway.sol
voltrevo: Good point. My subjective preference leaking in here. Will fix.

contracts/contracts/VerificationGateway.sol
jacque006: @jzaki @voltrevo Upon further reflection, hardcoding gas values is an anti-pattern, and could potentially brick the VG (maybe linked wallets?) in the future.

It is not uncommon for gas costs of different ops in the EVM to change over time. Generally they are lowered as hardware gets more powerful and cheaper, but sometimes they are increased for various reasons. See [EIP-2929](https://eips.ethereum.org/EIPS/eip-2929) for an example.

Our BLS contracts from Hubble have to deal with this as well, and are able to [generate the gas costs for BLS sig ops as part of the initialization process](https://github.com/web3well/bls-wallet/blob/main/contracts/contracts/lib/hubble-contracts/contracts/libs/BNPairingPrecompileCostEstimator.sol#L28-L107) . (There also appear to be [some hardcoded values](https://github.com/web3well/bls-wallet/blob/main/contracts/contracts/lib/hubble-contracts/contracts/libs/BNPairingPrecompileCostEstimator.sol#L47-L51), not sure what impact that will have in the future)

Given that, one solution we could do instead would be to to generate & cache this value on init, and then allow someone (public? connected BLS Wallet?) to call a function to recalculate & cache this value if gas costs every change in a way that impacts the VG. There may be other options I'm unaware of as well.

contracts/clients/src/BlsWalletWrapper.ts
jacque006: Gotcha, that makes sense to me. Will this break/work sub-optimally with ~other~ certain overhead values ~on other L2 EVMs~?

voltrevo: @jacque006 I'm not sure what you mean. Could you give an example?

contracts/contracts/VerificationGateway.sol
voltrevo: @jacque006 Yikes. Yeah that sounds bad.

How about we just include it in the overall gas limit? Something like this:

```solidity=
IWallet wallet = getOrCreateWallet(bundle.senderPublicKeys[i]);

uint256 opGas = bundle.operations[i].gas;
uint256 gasBeforeNonce = gasleft();
uint256 nonce = wallet.nonce{gas:gasRemaining}();
uint256 nonceGas = gasBeforeNonce - gasleft();

// check nonce then perform action
if (bundle.operations[i].nonce == nonce) {
    // request wallet perform operation
    (
        bool success,
        bytes[] memory resultSet
    ) = wallet.performOperation{gas:opGas-nonceGas}(bundle.operations[i]);
    successes[i] = success;
    results[i] = resultSet;
    emit WalletOperationProcessed(
        address(wallet),
        bundle.operations[i].nonce,
        bundle.operations[i].actions,
        successes[i],
        results[i]
    );
}
```

@jzaki wdyt?

contracts/clients/src/BlsWalletWrapper.ts
jacque006: Sorry, explained/thought through this poorly. Because of the fixed `10000` value you are using and the lack of rounding on the entire gas value, you can still get decimal values out of this arithmetic. I used the gas limit from https://explorer.arbitrum.io/tx/0xa0cd48e1f221e9ff876c441979b28852da93c3cebc49562dce1d654e4c4fa89e to test with a small overhead value:

image

voltrevo: > the lack of rounding on the entire gas value

Nah, `gas.div(10000)` is integer division, since `gas` is a `BigNumber`:

```
➜  contracts git:(main) node
Welcome to Node.js v16.17.0.
Type ".help" for more information.
> ethers = require('ethers')
> ethers.BigNumber.from(15).div(7).toString()
'2'
```

Let's add a comment to clear this up.

Was that your only concern?

jacque006: Ah yes forgot about that, good is is then :+1: 

Comment would be good, test cases around a range of values would be better :)

contracts/contracts/VerificationGateway.sol
jzaki: Will change it to this :+1: This will cater for wallets that are more intensive with their nonce calculation. 
Related #505

PR #496, opened 2023-02-01
Followups for #487 (gas limit PR)

voltrevo:
## What is this PR doing?

Implements followups for #487.

Excludes adding test (this was the only substantial change, doesn't feel worth it right now).

`yarn test`

Nope.

## Checklist

## Guidelines

### DIFF
--- a/contracts/clients/src/BlsWalletWrapper.ts
+++ b/contracts/clients/src/BlsWalletWrapper.ts
@@ -239,7 +239,7 @@ export default class BlsWalletWrapper {
   }
 
   /** Estimate the gas needed for an operation. */
-  async estimateGas(operation: Omit<Operation, "gas">) {
+  async estimateGas(operation: Omit<Operation, "gas">): Promise<BigNumber> {
     const exists =
       (await this.walletContract.provider.getCode(this.address)) !== "0x";
 
--- a/contracts/clients/src/BlsWalletWrapper.ts
+++ b/contracts/clients/src/BlsWalletWrapper.ts
@@ -230,6 +230,11 @@ export default class BlsWalletWrapper {
   /** Sign an operation with an estimate of the gas required. */
   async signWithGasEstimate(
     operation: Omit<Operation, "gas">,
+
+    /**
+     * Optional: Multiply estimate by `(1+overhead)` to account for uncertainty.
+     * Reduces the chance of running out of gas.
+     */
     overhead = 0,
   ): Promise<Bundle> {
     let gas = await this.estimateGas(operation);
--- a/contracts/contracts/VerificationGateway.sol
+++ b/contracts/contracts/VerificationGateway.sol
@@ -91,7 +91,7 @@ contract VerificationGateway
             bytes32 keyHash = keccak256(abi.encodePacked(bundle.senderPublicKeys[i]));
             address walletAddress = address(walletFromHash[keyHash]);
             if (walletAddress == address(0)) {
-                walletAddress = address(uint160(uint(keccak256(abi.encodePacked(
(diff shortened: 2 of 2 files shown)


contracts/clients/src/BlsWalletWrapper.ts
```diff
@@ -230,6 +230,11 @@ export default class BlsWalletWrapper {
   /** Sign an operation with an estimate of the gas required. */
   async signWithGasEstimate(
     operation: Omit<Operation, "gas">,
+
+    /**
+     * Optional: Multiply estimate by `(1+overhead)` to account for uncertainty.
+     * Reduces the chance of running out of gas.
+     */
     overhead = 0,
   ): Promise<Bundle> {
```
jacque006, line 239:
```suggestion
  /**
   * Sign an operation with an estimate of the gas required.
   *
   * @param operation Operation to sign
   * @param overhead (optional) Multiplies estimate by `(1+overhead)` to account for uncertainty. Reduces the chance of running out of gas.
   * @returns Signed Bundle with gas estimate
   */
  async signWithGasEstimate(
    operation: Omit<Operation, "gas">,
    overhead = 0,
  ): Promise<Bundle> {
```

jacque006 (approved):

contracts/clients/src/BlsWalletWrapper.ts
voltrevo: I'm not a fan of this format because it encourages adding a docstring for all params and the return value.

The reason that is bad (imo) is because of exactly the things added in this example:

> @param operation Operation to sign

- This doesn't explain anything, and is therefore a distraction.

> @returns Signed Bundle with gas estimate

- This is already explained by the function name and the return type, so it doesn't add anything, and is therefore a distraction.

Granted, it sometimes can have marginal value (like if you're not thinking clearly, and then realise "oh, the *operation* is the thing getting signed... right"), but I don't see it as worth the distraction, and so isn't my jam.

A different point though is the potentially improved compatibility of putting the param docstring above the function instead of directly above the param. I also prefer next to the param because, well, that's the thing you're documenting. There might be tooling which works better with the above-function style. Is there any such tooling that's relevant to us?

jacque006: I agree that it is a bit too verbose, and is harder to read in source code. The reasons I have started doing it in my work in our client APIs are:
- If we ever want to generate a more formal API documentation page, tools like https://tsdoc.org/ (see https://tsdoc.org/play) can parse this to auto-generated docs. And since it is close enough to the code it is easy to remember to update while making changes without switching workflows too much.
- In VSCode/other IDEs, hovering over the function/prop/etc. gives a nice human readable popup #devex

image

voltrevo: Hmm! Yeah that's pretty compelling. I had hoped that tools like tsdoc would parse `/** comments */` next to parameters, but it looks like that's not the case.

In vs code, you do get those comments next to parameters, though granted it's only when editing that specific parameter:

Screen Shot 2023-02-03 at 12 30 04 pm

Anyway, in light of tsdoc, I think this is a good idea for public APIs. I'd still be reluctant to do this kind of thing (especially adding docstrings for every parameter) when documenting other code for internal use though. Wdyt?

jacque006: > I'd still be reluctant to do this kind of thing (especially adding docstrings for every parameter) when documenting other code for internal use though. Wdyt?

Ab-so-lutely. Normal/appropriate comments for private/encapsulated functionality is fine, TS/JSDoc is overkill.

Issue #413, opened 2022-11-30
Move references to private key outside of BlsSigner


JohnGuilding:
## Summary

Currently, the BlsSigner is based off of `JsonRpcSigner` and has a singleton BlsWalletWrapped instance attached. This means the class 'knows' it's own private key. 

In ethers, there are two main types of signer:

- [Wallet](https://docs.ethers.io/v5/api/signer/#Wallet), which is a class which knows its private key and can execute any operations with it.
- [JsonRpcSigner](https://docs.ethers.io/v5/api/providers/jsonrpc-provider/#JsonRpcSigner), which is connected to a [JsonRpcProvider](https://docs.ethers.io/v5/api/providers/jsonrpc-provider/) (or sub-class) and is acquired using [getSigner](https://docs.ethers.io/v5/api/providers/jsonrpc-provider/#JsonRpcProvider-getSigner)

If you do a ctrl f for "privateKey", you can find it implemented in the ethers `Wallet` class, but no reference in the `JsonRpcSigner`. This combined with the above information from the docs suggests that if a signer class knows it own private key, it is more similar to a Wallet class and may not accurately represent a JsonRpcSigner class.

## Tasks
- Move reference of private key from BlsSigner?
- Keep private key but re-model class based on `Wallet`?
- Do both of the above tasks and have two bls signers - something like `BlsWallet.ts` and `BlsJsonRpcSigner`?

### Further reading 1
"The Wallet singer is used when your ethers.js instance knows the private key directly."

"Since MetaMask doesn't share the key with other applications, ethers.js uses the JsonRpcSigner to be able to request the local MetaMask instance over its API to sign the transaction when needed, and then receive the signed transaction back, without ethers.js ever knowing the key."

Taken from [What is the difference between a Wallet and a JsonRpcSigner?](https://stackoverflow.com/questions/71755563/what-is-the-difference-between-a-wallet-and-a-jsonrpcsigner)

### Further reading 2
[...]

Issue #499, opened 2023-02-02
Create new private key and signing management class


JohnGuilding:
The private key gets passed around in several places within the `BlsWalletWrapper` class. By creating a new class that manages the private key and specific signing methods, we can reduce the number of places the private key is passed around. 

**Benefits this change will bring**
1. Pass the private key around less - improves [data abstraction](https://en.wikipedia.org/wiki/Abstraction_(computer_science)#Data_abstraction) and security
2. Single abstraction that unifies private key management and signing

The `BlsWalletWrapper` constructor currently takes a `BlsWalletSigner` instance and a private key. We could instead create a new class that would represent a single abstraction for both private key management and signing functionality.

This would represent a significant change to the client module, so we should carefully consider the implementation, tradeoffs and impacts it will have. Some methods on the existing signer such as `aggregate`, `getPublicKey` may not belong in this new class, so we would need to decide how to deal with these.

It's worth noting we've mentioned a "new class" for the private key and the signing here, it may turn out that moving the private key to the existing signer class (`BlsWalletSigner`), and moving non-relevant members out such as `aggregate`, out to a "new class" may work better. These specific implementation details are up for discussion, the main purpose of this issue is to unify the private key and signing into a single abstraction that we're happy with, and makes sense.

PR #506, opened 2023-02-06
Move private key references to signer wrapper class

JohnGuilding:
## What is this PR doing?
Adds a wrapper class that contains the private key associated with the account and a BlsWalletWrapper class instance.

#413 

## Checklist

## Guidelines

### DIFF
--- a/contracts/clients/src/BlsProvider.ts
+++ b/contracts/clients/src/BlsProvider.ts
@@ -9,7 +9,7 @@ import poll from "./helpers/poll";
 export default class BlsProvider extends ethers.providers.JsonRpcProvider {
   readonly aggregator: Aggregator;
   readonly verificationGatewayAddress: string;
-  signer!: BlsSigner;
+  signer: BlsSigner | undefined;
 
   constructor(
     aggregatorUrl: string,
@@ -74,7 +74,7 @@ export default class BlsProvider extends ethers.providers.JsonRpcProvider {
     return this.signer.constructTransactionResponse(
       actionData,
       result.hash,
-      this.signer.wallet.address,
+      this.signer.blsSignerWrapper!.wallet.address,
     );
   }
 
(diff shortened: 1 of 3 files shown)


contracts/clients/src/BlsSigner.ts
```diff
@@ -5,12 +5,25 @@ import BlsProvider from "./BlsProvider";
...
+/**
+ * Dedicated class to manage the private key and BlsWalletWrapper instance.
+ */
+export class BlsSignerWrapper {
+  wallet: BlsWalletWrapper;
+  privateKey: string;
+
+  constructor(wallet: BlsWalletWrapper, privateKey: string) {
+    this.wallet = wallet;
+    this.privateKey = privateKey;
```
JohnGuilding, line 17: Not actually using this property (`blsSignerWrapper.privateKey`) anywhere. `blsSignerWrapper.wallet.privateKey` is used instead (see `connectedUnchecked`). Wondering whether it is actually needed?

contracts/clients/src/BlsProvider.ts
```diff
@@ -9,7 +9,7 @@ import poll from "./helpers/poll";
 export default class BlsProvider extends ethers.providers.JsonRpcProvider {
```
JohnGuilding, line 9: This class has two todos that are tagged to be resolved as part of this issue/PR:
- https://github.com/web3well/bls-wallet/blob/8ef89face77a7b8cf7619f46d3cf6aafbab7551a/contracts/clients/src/BlsProvider.ts#L33
- https://github.com/web3well/bls-wallet/blob/8ef89face77a7b8cf7619f46d3cf6aafbab7551a/contracts/clients/src/BlsProvider.ts#L50

I could remove these checks and run `this.getSigner(privateKey)`, but that would require passing the private key to each of these methods. 

The original motivation to removing these checks, and subsequently the signer property on the BlsProvider class, was because the ethers JsonRpcProvider has no such reference to it's signer. Maybe such a difference from the ethers class is ok, but then again, we are trying to maintain a very similar interface. Open to suggestions on making this cleaner, or whether it's ok to be left as is.

JohnGuilding:
Abandoning this PR as it doesn't achieve the desired outcome, which is to obfuscate the private key. 

The BlsWalletWrapper class should be changed first. See #499

PR #510, opened 2023-02-08
Switch to sqlite

### LINKED ISSUE #507: Use sqlite instead of postgres
I was chatting with @jacque006 the other day and realized we are on the same page about switching to sqlite.

## Why?

Postgres is a heavier database solution that we solve by running in its own docker container. This is a significant dependency to deal with when running an aggregator. (Related: #498.)

Sqlite, by contrast, would run inside the same container by design. This means the user doesn't even need to be aware of it when getting started. A sqlite database is also simply a file and access to it is therefore controlled by the filesystem, which simplifies our code a bit because there isn't a socket involved and database names and passwords.

## Why not?

If you wanted to share data between multiple aggregators, then postgres might be more suitable. However, there are two big problems with this:
1. The aggregator isn't currently designed this way, it expects exclusive access.
2. Postgres would only help to scale aggregators for one centralized operator. For decentralized sharing with multiple operators (like the mempool), independent storage with sqlite is more appropriate.

## Isn't this a lot of work?

Not really. We're not using any postgres-specific features. We only have one table and it's a pretty simple one. All database logic is contained within `BundleTable`, so we just update that and the rest of the code remains unchanged because it only uses the database via that abstraction.

voltrevo:
## What is this PR doing?

Replaces postgres with sqlite in the aggregator.

Also implements `wait-for-rpc.sh` because the aggregator was starting up too fast in CI 😂. Used in `yarn start` too for a big improvement over `sleep 5`.

Run the premerge script / CI tests.

Resolves #507.

## Checklist

## Guidelines


aggregator/.env.example
```diff
@@ -11,13 +11,8 @@ PRIVATE_KEY_AGG=0x5de4111afa1a4b94908f83103eb1f1706367c2e68ca870fc3fb9a804cdab36
...
 TEST_BLS_WALLETS_SECRET=test-bls-wallets-secret
 
-PG_HOST=127.0.0.1
-PG_PORT=5432
-PG_USER=bls
-PG_PASSWORD=generate-a-strong-password
-PG_DB_NAME=bls_aggregator
+DB_PATH=aggregator.db
 
-BUNDLE_TABLE_NAME=bundles
```
voltrevo, line 20: The idea of setting the table name was two-fold:
- Different table names for different networks
- Use throwaway names when testing

However, with sqlite it makes more sense to just use a different file for different networks. In testing we can do even better and just bypass the filesystem completely since sqlite happily runs in-memory when you don't give it a file.

aggregator/programs/dropTables.ts
```diff
@@ -1,15 +0,0 @@
...
-import * as env from "../src/env.ts";
-import BundleTable from "../src/app/BundleTable.ts";
-
-const queryClient = createQueryClient(() => {});
-
-for (const tableName of [env.BUNDLE_TABLE_NAME]) {
-  const table = await BundleTable.create(queryClient, tableName);
-  await table.drop();
-  console.log(`dropped table ${tableName}`);
-}
```
voltrevo, line 15: Now you can just `rm aggregator.sqlite` 💪

.github/actions/local-aggregator-deploy/action.yml
```diff
@@ -4,14 +4,10 @@ description: Runs an aggregator instance
...
-      run: docker-compose up -d postgres
-
     - working-directory: ./aggregator
       shell: bash
       run: cp .env.test .env
 
     - working-directory: ./aggregator
       shell: bash
-      run: deno run --allow-read --allow-env --allow-net ./programs/aggregator.ts 2>&1 | tee -a aggregatorLogs.txt &
+      run: deno run --allow-read --allow-write --allow-env --allow-net ./programs/aggregator.ts 2>&1 | tee -a aggregatorLogs.txt &
```
jacque006, line 13: There are likely additional deno scripts in `./aggregator/manualTests` & `./aggregator/programs` that need `--allow-write` added to their bash run command (`#!/usr/bin/env -S deno...`) if they interact with the db.

aggregator/src/app/BundleTable.ts
```diff
@@ -109,109 +115,167 @@ function toInsertRawRow(row: InsertRow): InsertRawRow {
...
-    queryClient: QueryClient,
-    tableName: string,
-  ): Promise<BundleTable> {
-    const table = new BundleTable(queryClient, tableName);
-    await table.queryTable.create(tableOptions, CreateTableMode.IfNotExists);
-
-    return table;
+  dbQuery(sql: string, params?: sqlite.QueryParameterSet) {
+    this.onQuery(sql, params);
+    return this.db.query(sql, params);
```
jacque006, line 150: I'm surprised this is synchronous, most file i/o is not. Looks like under the covers it's making calls to a WASM binary. Most sync file i/o in NodeJS (i.e. `fs.readFileSync`) and other JS systems is discouraged as it blocks the event loop. Looks like this might be flushing writes in the background. Worth keeping an eye on if we see any stuttering/performance/persistence issues.

jacque006 (approved):

.github/actions/local-aggregator-deploy/action.yml
voltrevo: Nice catch.

Includes `aggregator.ts` 😅.

Noticed and fixed a few other things: https://github.com/web3well/bls-wallet/pull/516.

aggregator/src/app/BundleTable.ts
voltrevo: Yeah this is an interesting one. I've met this point before, and my understanding is that (surprisingly) it's normal and expected to do things synchronously when using sqlite. The idea is that sqlite queries need to be sequential anyway, and they're very fast, so async doesn't really fit. There are exceptions of course, but I don't think they apply to us. [More info](https://github.com/WiseLibs/better-sqlite3#why-should-i-use-this-instead-of-node-sqlite3). [Also](https://github.com/dyedgreen/deno-sqlite/issues/150).

PR #511, opened 2023-02-08
BLSExpanderDelegator and FallbackExpander

### LINKED ISSUE #509: Create new expander
Start working towards an expander like the one described here:
https://hackmd.io/0q7H3Ad0Su-I4RWWK8wQPA

The initial prototype should probably just be the top-level expander (which simply delegates to other expanders) and a fallback expander which will just copy the bytes over to allow us to encode any operation.

Register the fallback expander in `yarn start` and add a test demonstrating an eth transfer via the top-level expander.

voltrevo:
## What is this PR doing?

Implements the "top-level expander" [from this doc](https://hackmd.io/0q7H3Ad0Su-I4RWWK8wQPA) and includes a fallback expander which can be used with any operation.

Notably, this fallback expander is still able to do a lot of compression, due to the inherent verbosity of solidity's ABI which is optimized for the EVM runtime and not reducing calldata. [More info](https://github.com/web3well/bls-wallet/compare/contract-updates...bw-509-new-expander?expand=1#diff-429f3b394637354cf3ba9e8978dbce169c07cd04e21a1660ab06044373ebc10dR9-R44).

This is tested using a slightly modified version of the pre-existing `should send ETH with function` test. It just tests the same results are achieved but does so using compression.

By adding a bit of logging, we can see that the reduction in calldata is quite dramatic:

```
should send ETH with function call

fx.verificationGateway.interface.encodeFunctionData(
  "processBundle",
  [bundle],
)

708 bytes:

0xbca0ccc6
0000000000000000000000000000000000000000000000000000000000000020
0a3738223881f468acda63fae7e23b3bc9782d7de1455ef1a5711537c893c86e
09c49b13b259bf5e04a734762802078feef18734ad92e82a7c650fe138047d18
0000000000000000000000000000000000000000000000000000000000000080
0000000000000000000000000000000000000000000000000000000000000120
0000000000000000000000000000000000000000000000000000000000000001
305968ca37c5a6c65876dd6ab2f2099bc962f00f24ea1680a97d8c6894a64a4c
0c2ddc68876c406434cea5106569f2b33ecec31f1c0486eec8436fd31f51b028
2daff5162791d46f4faf6f674efabb5549d7e4f1f1119f1bda8769999868dc6b
07133be305a7c9496dc4863fa3f8048811a79b8018ddc2ec06834c2f5fe8f6f8
0000000000000000000000000000000000000000000000000000000000000001
0000000000000000000000000000000000000000000000000000000000000020
0000000000000000000000000000000000000000000000000000000000000001
0000000000000000000000000000000000000000000000000000000000008db0
0000000000000000000000000000000000000000000000000000000000000060
0000000000000000000000000000000000000000000000000000000000000001
0000000000000000000000000000000000000000000000000000000000000020
00000000000000000000000000000000000000000000000000038d7ea4c68000
0000000000000000000000002279b7a0a67db372996a5fab50d91eaa73d2ebe6
0000000000000000000000000000000000000000000000000000000000000060
0000000000000000000000000000000000000000000000000000000000000004
416429f100000000000000000000000000000000000000000000000000000000

(<--------------------- 32 bytes wide ------------------------>)

gasUsed: 278,064
```

```
should send ETH with function call via fallback expander

fx.blsExpanderDelegator.interface.encodeFunctionData(
  "run",
  [
    bundleCompressedOperations(
      [
        compressAsFallback(
          Fixture.expanderIndexes.fallback,
          bundle.senderPublicKeys[0],
          bundle.operations[0],
        ),
      ],
      bundle.signature,
    ),
  ],
)

324 bytes:

0x3a276523
0000000000000000000000000000000000000000000000000000000000000020
00000000000000000000000000000000000000000000000000000000000000e8
01001ea47e47adee4a93a6cfdbf68d07efd9b288722a4a6f4c93e186e82c526e
107224998ad3f3913a424d90980a613f7baac79ad8706dd3947b7df7d30cfd40
a3b709854854dae93565bc8743eb8528f617d450a2137fb1214d1234b7080051
9fe42433c29939ea6a9df4c4f4e176ed352f680debb56f5710d8b00deeb2fcde
a5bb01829b300181e3afeaa69a8000959922be3caee4b8cd9a407cc3ac1c251c
2007b104416429f10e5f95c1142d0c6316f5c6f83a8a34018d4748924cedcffa
6fa63f0e2c7eb24d15e1e41cd933428e96129efc5e00d1c439bdc699441f7be6
4d2a8c9416fd4abb000000000000000000000000000000000000000000000000

(<--------------------- 32 bytes wide ------------------------>)

gasUsed: 306210
(+28,146. Only local geth here, need to separate L1 & L2 gas for a more meaningful number.)
```

Still some solidity ABI overhead in there because of the use of the ABI at the top level to represent a `bytes` argument (88 bytes) and the function id (4 bytes). These can be fixed too 💪 (later).

Also note that this is looking at the bundle as a whole. If you look at the marginal cost of an operation, it's more like 640->194 bytes, which is even better ([same link as before](https://github.com/web3well/bls-wallet/compare/contract-updates...bw-509-new-expander?expand=1#diff-429f3b394637354cf3ba9e8978dbce169c07cd04e21a1660ab06044373ebc10dR9-R44)).

`yarn test`

Resolves #509.

## Checklist

## Guidelines


contracts/contracts/interfaces/IWallet.sol
```diff
@@ -5,6 +5,12 @@ pragma abicoder v2;
 /** Interface for a contract wallet that can perform Operations
  */
 interface IWallet {
+    /** Aggregated signature with corresponding senders + operations */
+    struct Bundle {
+        uint256[2] signature;
+        uint256[4][] senderPublicKeys;
+        IWallet.Operation[] operations;
+    }
```
jacque006, line 13: This would make more sense to me in `IBundleProcessor`, which could be extracted to its own interface.

contracts/package.json
```diff
@@ -20,7 +20,10 @@
   "author": "James Zaki",
   "license": "MIT",
   "dependencies": {
-    "@openzeppelin/contracts": "^4.7.3"
+    "@openzeppelin/contracts": "^4.7.3",
+    "dotenv": "^16.0.3",
+    "ethers": "^5.7.2",
+    "hardhat": "^2.12.1"
```
jacque006, line 26: Why were these moved? Should be dev deps as they are not used in the deployed contracts

contracts/contracts/BLSExpanderDelegator.sol
```diff
@@ -0,0 +1,93 @@
...
+        (vlqValue, stream) = VLQ.decode(stream);
+
+        bundle.senderPublicKeys = new uint256[BLS_KEY_LEN][](vlqValue);
+        bundle.operations = new IWallet.Operation[](vlqValue);
+
+        uint256 opsDecoded = 0;
+
+        while (stream.length > 0) {
+            // First figure out which expander to use.
+            (vlqValue, stream) = VLQ.decode(stream);
```
jacque006, line 56: I would expect the logic in this index of expanders to be non-VLQ specific. Is there another way we could determine which expander is selected?

contracts/contracts/FallbackExpander.sol
```diff
@@ -0,0 +1,86 @@
...
+ * 04 // 4 bytes for encodedFunction
+ * 1dfea6a0 // sendEthToTxOrigin
+ *
+ * The proposal doc for the new expander lists the same example ("Example of an
+ * Expanded User Operation" https://hackmd.io/0q7H3Ad0Su-I4RWWK8wQPA) using the
+ * solidity ABI, which uses 608 bytes. Here we've encoded the same thing (plus
+ * gas) in 194 bytes, which is (about) 70% smaller. (If you account for the
+ * zero-byte discount, the saving is still over 30%.)
+ */
+contract FallbackExpander is IExpander {
```
jacque006, line 45:
```suggestion
contract VLQExpander is IExpander {
```

contracts/test/walletAction-test.ts
```diff
@@ -176,6 +180,56 @@ describe("WalletActions", async function () {
...
+    });
+
+    const MockAuction = await ethers.getContractFactory("MockAuction");
+    const mockAuction = await MockAuction.deploy();
+    await mockAuction.deployed();
+
+    expect(await fx.provider.getBalance(sendWallet.address)).to.equal(
+      ethToTransfer,
+    );
+    expect(await fx.provider.getBalance(mockAuction.address)).to.equal(0);
```
jacque006, line 199: https://www.chaijs.com/plugins/chai-as-promised/

```suggestion
    await expect(fx.provider.getBalance(sendWallet.address)).to.eventually.equal(
      ethToTransfer,
    );
    await expect(fx.provider.getBalance(mockAuction.address)).to.eventually.equal(0);
```

contracts/test/walletAction-test.ts
```diff
@@ -176,6 +180,56 @@ describe("WalletActions", async function () {
...
+            compressAsFallback(
+              Fixture.expanderIndexes.fallback,
+              bundle.senderPublicKeys[0],
+              bundle.operations[0],
+            ),
+          ],
+          bundle.signature,
+        ),
+      )
+    ).wait();
```
jacque006, line 225: #nitpick I dislike nested awaits as it can be difficult to:
 - Understand what the multi async calls are just by looking at code
 - Can make debugging more difficult. On a throw, which await scope was it?

```suggestion
    const txn = await fx.blsExpanderDelegator.run(
        bundleCompressedOperations(
          [
            compressAsFallback(
              Fixture.expanderIndexes.fallback,
              bundle.senderPublicKeys[0],
              bundle.operations[0],
            ),
          ],
          bundle.signature,
        ),
      )
    );
    await txn.wait();
```

contracts/test/walletAction-test.ts
```diff
@@ -176,6 +180,56 @@ describe("WalletActions", async function () {
...
+
+    expect(await fx.provider.getBalance(sendWallet.address)).to.equal(0);
+    expect(await fx.provider.getBalance(mockAuction.address)).to.equal(
+      ethToTransfer,
+    );
+  });
```
jacque006, line 231:
```suggestion
    await expect(fx.provider.getBalance(sendWallet.address)).to.eventually.equal(0);
    await expect(fx.provider.getBalance(mockAuction.address)).to.eventually.equal(
      ethToTransfer,
    );
  });
```

jacque006 (approved):
With these additions, it might make sense to move existing & new expander contracts to a sub-dir in contracts i.e. `./contracts/contracts/expanders` to more explicitly point out that they are not core contracts.

Many, many :+1:'s for in code comments & test cases

contracts/package.json
voltrevo: Hmm. I did this to resolve the lint warnings.

It sounds like you are targeting the use-case where people list bls-wallet in their package.json in order to use the contracts directly.

It's a bit weird because you're not supposed to use `devDependencies` in your core/non-test code. This is why the linter is complaining. I suppose there are three levels of distinction in this project (instead of just 2):
1. Solidity code
2. Deployment code
3. Test code

Maybe we could tell the linter somehow that deployment code isn't 'core', like it should treat it more like test code.

Alternatively, it might be more technically correct to make an adjustment to only include the solidity dependencies if/when we decide to publish an npm module for importing our contracts.

Wdyt?

contracts/contracts/BLSExpanderDelegator.sol
voltrevo: I'm not sure what you mean. The idea is that the expander index is encoded as a VLQ. Alternatively, if we used a fixed width to encode which expander to use then we'd either have an uncomfortably finite number of slots for expanders or we'd be using more bytes than necessary.

From the hackmd proposal:

```solidity
contract BLSExpander {
    mapping(uint256 => address) public expanders;
    uint256 public expanderCount = 0;

    // Potential to use fallback() instead to save 4 bytes
    function run(bytes calldata input) external {
        // Loop until all bytes consumed:
        //   Read VLQ <-------------------------------------------- This line
        //   Delegate to specialized expander
        //   Accumulate operations
        // Call VerificationGateway
    }

    function registerExpander(address expander) external {
        expanders[expanderCount] = expander;
        expanderCount++;
    }
}
```

contracts/contracts/FallbackExpander.sol
voltrevo: I prefer `FallbackExpander` because 'fallback' is the primary idea here, not VLQ. In other words, the purpose of this expander is to be able to encode UserOperations without any limitations. Using VLQ heavily is just how it happens to be currently implemented.

Also, I'm cooking up some more ideas for the fallback expander to make it better, which would add some non-VLQ stuff in here.

contracts/test/walletAction-test.ts
voltrevo: Hmm, I have almost the opposite opinion - that nested awaits (when used sparingly) are helpful in keeping the code concise.

> Can make debugging more difficult. On a throw, which await scope was it?

I'm not sure I understand. On a throw, the stack trace should point to the particular call. The stack doesn't merely point to the offending statement, instead it points to the specific location inside the statement. (Also, this point would apply equally to all nested function calls, not just awaits, right?)

Having said all that, I think it's fair this statement is a bit bloated. Extracting the `.run` call leaves us with a statement that still mostly has this bloat though. I'd much rather extract the compressed bundle:

```ts
const compressedBundle = bundleCompressedOperations(
  [
    compressAsFallback(
      Fixture.expanderIndexes.fallback,
      bundle.senderPublicKeys[0],
      bundle.operations[0],
    ),
  ],
  bundle.signature,
);

await (
  await fx.blsExpanderDelegator.run(compressedOperation)
).wait();
```

Furthermore, I would say that the `await (await contract.fn(args)).wait()` is confusing for other reasons (two awaits and a .wait all next to each other... 😵‍💫). However, this is an established pattern we use everywhere, which:
1. Mitigates the problem, because it's something you get acclimatized to
2. If we are to fix it, I'd like to fix it everywhere, and I'd prefer something like `await receiptOf(contract.fn(args))`.

contracts/package.json
jacque006: I looked into those lint warnings awhile back, and did not find a satisfactory answer that didn't involve turning off that lint rule or manually listing packages to ignore.

In the short/medium term I don't think it will matter much. But I think
> Maybe we could tell the linter somehow that deployment code isn't 'core', like it should treat it more like test code.

would be a better way to go.

> Alternatively, it might be more technically correct to make an adjustment to only include the solidity dependencies if/when we decide to publish an npm module for importing our contracts.

If we did this it would also matter more (package bloat), but you can directly import contracts/libs via url so I think there is less of a need for that unless we are also including deployment scripts in that package. Then this would stay as is per your changes.

contracts/contracts/BLSExpanderDelegator.sol
jacque006: So for example, let's say I write a gzip expander. I would need to encode the expander index VLQ, but then the remaining operation/bundle data would be gzipped. I could see that being confusing/more complicated than using a non-VLQ format.

> ...have an uncomfortably finite number of slots for expanders or we'd be using more bytes than necessary

If an expander ever ran out of slots with a smaller index, you just deploy another expander delegator or nest one by having another delegator implement `IExpander`.

voltrevo: I mean, it's VLQ, but for a particular expander it's always a fixed byte sequence. I don't really see the point of confusion there... you'll need to prepend a fixed byte sequence when using gzip/etc anyway. (And even so, VLQ is an incredible simple format which can be quickly understood conceptually and implemented in a few lines. Why would this be a pain point?)

> If an expander ever ran out of slots with a smaller index, you just deploy another expander delegator or nest one by having another delegator implement IExpander.

Seems a lot more of a headache than putting a VLQ there. Maybe worth a chat about this, seems like there's some subtle thing that we're not clicking on.

contracts/test/walletAction-test.ts
jacque006: > I'm not sure I understand. On a throw, the stack trace should point to the particular call. The stack doesn't merely point to the offending statement, instead it points to the specific location inside the statement.

Yes, but I would argue that it is easier to find the statement that failed when not nested. Example:

`nestedAwait.js`
```javascript
const txnFactory = async (fail) => ({
  wait: async () => {
    if (fail) {
      throw new Error("txn failed");
    } 
  },
});

(async() => {
  try {
    await (await txnFactory(true)).wait();
  } catch (err) {
    console.error(err);
  }

  console.log("---");

  try {
    const txn = await txnFactory(true);
    await txn.wait();
  } catch (err) {
    console.error(err);
  }

})();
```
```sh
node nestedAwait.js 
Error: txn failed
    at Object.wait (/home/jacque006/workspaces/bls-wallet/nestedAwait.js:4:13)
    at /home/jacque006/workspaces/bls-wallet/nestedAwait.js:11:36
---
Error: txn failed
    at Object.wait (/home/jacque006/workspaces/bls-wallet/nestedAwait.js:4:13)
    at /home/jacque006/workspaces/bls-wallet/nestedAwait.js:20:15
```
For the nested await, I need to find the *character column* to find the statement that threw.
When not nested, I know the await on line 20 failed. IMO much easier to find. It also gets worse if the code your debugging is transpiled.

> (Also, this point would apply equally to all nested function calls, not just awaits, right?)

Yes, before NodeJS went more promisy it was referred to as Callback Hell :)

contracts/test/walletAction-test.ts
voltrevo: Love it. Let's do it everywhere: https://github.com/web3well/bls-wallet/pull/518.

contracts/test/walletAction-test.ts
voltrevo: > Yes, before NodeJS went more promisy it was referred to as Callback Hell :)

Ah, the good old days /s

(Side comment, will follow-up)

contracts/test/walletAction-test.ts
voltrevo: Love it. Let's do it everywhere: https://github.com/web3well/bls-wallet/pull/518.

contracts/test/walletAction-test.ts
voltrevo: @jacque006 Point taken about the improved stack trace. Appreciate the isolated demo code 🚀.

I do feel it is not enough though since the same applies to all nested function calls, and the verbosity of flattening all function calls would not be worth it.

How do you feel about `await receiptOf(contract.fn(args))`?

PR #513, opened 2023-02-08
Use geth in client integration tests

JohnGuilding:
## What is this PR doing?
Runs integrations tests against geth node instead of hardhat node.

Integration tests now run locally in 185 -190 seconds (3 minutes and 5/10 seconds). They previously took about 250-255 seconds (4 minutes and 10/15 seconds). This is around a 24-27% decrease.

Run integration tests using `yarn start` first instead of `yarn start-hardhat`

#456 

## Checklist

## Guidelines


.github/workflows/contracts.yml
```diff
@@ -40,5 +40,6 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - uses: ./.github/actions/setup-contracts-clients
-    - uses: ./.github/actions/local-contract-deploy
+    - uses: ./.github/actions/local-contract-deploy-hardhat
```
JohnGuilding, line 43: I decided to separate the contract deploy scripts out into a hardhat version and a geth version, as using just the geth version caused issues with `test-gas-measurements`. See [this failing workflow](https://github.com/web3well/bls-wallet/actions/runs/4142899084/jobs/7164203145). This was resolved by running these measurements against hardhat. The issue with using geth is line 45 in this file.

As we seem to be moving towards running things on geth instead of hardhat, imagine this will need to be changed at some point, but thought that was out of scope for this PR

aggregator/.env.test
```diff
@@ -23,7 +23,7 @@ MAX_UNCONFIRMED_AGGREGATIONS=3
 LOG_QUERIES=true
 TEST_LOGGING=true
 
-REQUIRE_FEES=true
+REQUIRE_FEES=false
```
JohnGuilding, line 26: The provider does not yet support paying the aggregator. I'll be making those changes next and have some working code stashed locally, so will be setting this back to true once that is done. Thought making that change would be out of scope for this PR.

contracts/scripts/wait-for-contract-deploy.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
JohnGuilding, line 1: Was getting a few errors as the aggregator was not waiting for the contracts to deploy. See [example errors in this workflow](https://github.com/web3well/bls-wallet/actions/runs/4139071494/jobs/7156223194). 

Andrew suggested creating a similar script to `wait-for-rpc.sh` to overcome this, and this was the solution.

Following code was courtesy of chatGPT.

.github/workflows/contracts.yml
JohnGuilding: You can observe the error locally by running `yarn start` and then `yarn hardhat run ./scripts/measure_gas/run.ts --network gethDev`

voltrevo (approved):

blakecduncan (approved):
25% faster is awesome 💯

PR #518, opened 2023-02-10
Use .eventually in contracts

voltrevo:
## What is this PR doing?

```ts
// bad
expect(await foo()).to.eq('bar');

// good
await expect(foo()).to.eventually.eq('bar');
```

This allows better errors to be surfaced - when the promise rejects, instead of emitting a naked exception, chai can capture it and give it more context.

`yarn test`

Nope.

## Checklist

## Guidelines


JohnGuilding (approved):
LGTM

Just the linting error that might need fixing. Think you can get round it with `Promise<typeof Experimental.BlsSigner[]>`

PR #527, opened 2023-02-13
Fix deterministic creation of VG using ProxyAdminGenerator

### LINKED ISSUE #332: Bug: ProxyAdmin's separate creation prevents consistent create2 addresses

voltrevo:
## What is this PR doing?

Extracted from https://github.com/web3well/bls-wallet/pull/464:

> In making the switch, I've also fixed the deterministic generation of VerificationGateway's ProxyAdmin using the new [ProxyAdminGenerator](https://github.com/web3well/bls-wallet/blob/bw-394-use-safe-singleton-factory/contracts/contracts/ProxyAdminGenerator.sol). This was critical to getting a predetermined address for VerificationGateway because proxyAdmin was a constructor parameter, which affects VerificationGateway's address. This was a pre-existing issue: https://github.com/web3well/bls-wallet/issues/332.

I had to inline `BLS_DOMAIN` and shorten some revert messages to stay inside the 30m deployment limit. These changes weren't needed in the original PR, I expect because deploying SafeSingletonFactory contract is more efficient. We can probably undo these changes later if we want to (or just benefit from the extra headroom).

`yarn test`

Resolves #332, #334.

## Checklist

## Guidelines

### DIFF
--- a/contracts/contracts/ProxyAdminGenerator.sol
+++ b/contracts/contracts/ProxyAdminGenerator.sol
@@ -1,0 +1,25 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity >=0.8.4 <0.9.0;
+
+import "@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol";
+
+/**
+ * Generates a ProxyAdmin and transfers its ownership to msg.sender.
+ * 
+ * This is useful to avoid including ProxyAdmin in your contract bytecode.
+ * Instead ProxyAdmin is included in this contract, and this contract's address
+ * can be provided to your contract's constructor.
+ */
+contract ProxyAdminGenerator {
+  function generate(bytes32 salt) external returns (ProxyAdmin) {
+    // This salting technique ensures two things:
+    // 1. Your ProxyAdmin has a predetermined address
+    // 2. No one else can generate or prevent access to your ProxyAdmin
+    bytes32 fullSalt = keccak256(abi.encode(msg.sender, salt));
+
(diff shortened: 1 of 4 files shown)


jzaki:
Commits reviewed in #464

PR #535, opened 2023-02-14
Provider pays aggregator fees

### LINKED ISSUE #534: Bls Provider pays aggregator

JohnGuilding:
## What is this PR doing?
Makes the Bls provider pay fees to the aggregator as it previously wasn't.
- Sets require fees to true in test env
- Update provider and signer to send fees to tx.origin
- Adds aggregatorUtilitiesAddress to provider constructor. This address is used when sending fees to the aggregator

Running integration tests

#534 

## Checklist

## Guidelines


contracts/clients/src/BlsProvider.ts
```diff
@@ -37,11 +41,32 @@ export default class BlsProvider extends ethers.providers.JsonRpcProvider {
...
+    const nonce = await BlsWalletWrapper.Nonce(
+      this.signer.wallet.PublicKey(),
+      this.verificationGatewayAddress,
+      this,
+    );
+
+    const estimateFeeBundle = this.signer.wallet.sign({
+      nonce,
+      actions: [actionWithoutEthValue],
+    });
```
voltrevo: I'm not sure what's going on here, why would you use the user's action without an eth value for estimation?

The strategy used [elsewhere](https://github.com/web3well/bls-wallet/blob/55ad789/aggregator/manualTests/mint1ViaAggregator.ts#L55-L65) is to append an action that pays tx.origin but only pays 1 wei.

contracts/clients/src/BlsSigner.ts
```diff
@@ -67,20 +71,40 @@ export default class BlsSigner extends Signer {
...
+      actions: [
+        action,
+        {
+          ethValue: feeEstimate,
+          contractAddress: this.aggregatorUtilitiesAddress,
+          encodedFunction:
+            aggregatorUtilitiesContract.interface.encodeFunctionData(
+              "sendEthToTxOrigin",
+            ),
+        },
```
voltrevo, line 105: `feeEstimate` didn't account for this extra action

contracts/clients/src/BlsProvider.ts
voltrevo: Note also that it's possible the user's transaction will revert if the `ethValue` is not included:

```
require(msg.value >= someThreshold, "Not enough funds provided");
```

In this case the aggregator wouldn't be able to provide a fee estimate.

JohnGuilding: Ahh yeah, this appears to be the solution to the issues I was having. Amended the code to follow the strategy that appends an action to pay tx.origin with 1 wei.

I've also updated the client docs to include more accurate examples of paying fees.

Fixed now, but for context, the issue I was getting was an intermittent error in the contract interaction tests with the following message: `TypeError: Cannot read properties of undefined (reading 'map'). The changes are to ensure the provider pays the aggregator fees by default`. It was occurring across different tests randomly. According to the call stack the error occurs here https://github.com/ethers-io/ethers.js/blob/ec1b9583039a14a0e0fa15d0a2a6082a2f41cf5b/packages/contracts/src.ts/index.ts#L338. 

Example of failing run https://github.com/web3well/bls-wallet/actions/runs/4176116862/jobs/7231972753

JohnGuilding: The source where I originally got the idea of a payment of zero was [this sentence](https://github.com/web3well/bls-wallet/blob/db968c34fd7ce30e01e0a3048aa9075370d367a0/aggregator/README.md?plain=1#L178-L180)  in the agg README, that mentions to include the payment action with a payment of zero. 

For my understanding, what context is the payment of zero mentioned in the readme relevant?

contracts/clients/src/BlsSigner.ts
JohnGuilding: Fee estimation updated

contracts/clients/README.md
```diff
@@ -157,34 +155,65 @@ const wallet = await BlsWalletWrapper.connect(
...
+    },
+  ],
 });
 ```
 
+```ts
+
+```
+
 ### Paying aggregator fees with custom currency (ERC20)
```
JohnGuilding, line 216: @voltrevo one piece of logic that confused me when looking around the code was [this](https://github.com/web3well/bls-wallet/blob/55ad789c3a86831d4bbadff92a402ae1d017f96c/aggregator/test/AggregationStrategy.test.ts#L95). Why is the fee included as the second argument to the mint function? It was my understanding that the second param for mint was the [amount to mint](https://github.com/web3well/bls-wallet/blob/db968c34fd7ce30e01e0a3048aa9075370d367a0/contracts/contracts/mock/MockERC20.sol#L15), not the fee?

I think I've updated the transfer example below to reflect this, but interested to understand the mechanics of how the fee is used with contract interactions

contracts/clients/README.md
```diff
@@ -157,34 +155,65 @@ const wallet = await BlsWalletWrapper.connect(
   provider,
 );
 const aggregator = new Aggregator("https://arbitrum-goerli.blswallet.org");
-const estimateFee = await aggregator.estimateFee(bundle); // Remember to include no payment with this bundle
+
+// Create a fee estimate bundle
```
blakecduncan, line 159: Nice this is a good example

contracts/clients/README.md
```diff
@@ -157,34 +155,65 @@ const wallet = await BlsWalletWrapper.connect(
...
+    {
+      ethValue: safeFee, // fee amount
+      contractAddress: aggregatorUtilitiesContract.address,
+      encodedFunction:
+        aggregatorUtilitiesContract.interface.encodeFunctionData(
+          "sendEthToTxOrigin",
+        ),
+    },
+
+    // You can also just add an action that transfers the fee amount to the aggregator address, instead of
```
blakecduncan: Sorry just for clarification, I'm not sure I'm understanding, what's the difference between these two actions?

contracts/clients/README.md
```diff
@@ -194,52 +223,75 @@ const wallet = await BlsWalletWrapper.connect(
...
+  ],
+});
+
+const feeEstimate = await aggregator.estimateFee(estimateFeeBundle);
+
+// Add a safety premium to the fee to account for fluctuations in gas estimation
+const safetyDivisor = 5;
+const feeRequired = BigNumber.from(feeEstimate.feeRequired);
+const safetyPremium = feeRequired.div(safetyDivisor);
+const safeFee = feeRequired.add(safetyPremium);
```
blakecduncan, line 260: We aren't using the feeEstimate/safeFee anywhere below.  Does this have anything to do with the bundle example created below?

contracts/clients/README.md
JohnGuilding: Because the aggregator detects it gets paid by observing the effect of a bundle on its own balance, we have different options for paying the aggregator. The first action in this code block is one way of paying the aggregator, and this second action is another way of paying the aggregator, by transferring it funds directly instead of calling `sendEthToTxOrigin`. So the example is supposed to show both ways of doing things.

I've split this into a different code block and updated the description with slightly more context. Let me know if that doesn't make sense still.

contracts/clients/README.md
JohnGuilding: Good catch, updated the example to include `safeFee`

blakecduncan (approved):
lgtm

contracts/clients/src/BlsProvider.ts
voltrevo: Ohh I think I see the confusion.

That sentence is about an additional action to pay 0 wei to tx.origin, which maybe also works. It doesn't mean changing the action to zero wei, but adding one that has 0 wei.

It's not completely settled whether appending a 1 wei payment to tx.origin is better than appending a 0 wei payment to tx.origin:

0 wei has the advantage that if the aggregator doesn't charge fees, then the wallet doesn't need any funds to simulate the transaction, and the user won't need any funds for the transaction anyway.

On the other hand, we should be de-emphasizing this possibility that 'the aggregator doesn't charge fees', because that's not realistic. And a 1 wei payment is maybe a better proxy for the gas that will be needed for the actual payment, since the payment might involve changing a slot from zero to non-zero, which can have its own cost, and a 0 wei payment doesn't capture that.

contracts/clients/README.md
voltrevo: Ah yeah, in that test, `mint` is being used to pay the fee. The test doesn't include an action for the user because it's not really relevant to the test - you can think of it like the user is paying for a no-op, which could actually happen if you just wanted to increment your wallet's nonce.

The aggregator can be configured to accept payments in an ERC20 token, which is what's configured for that test ([here](https://github.com/web3well/bls-wallet/blob/55ad789c3a86831d4bbadff92a402ae1d017f96c/aggregator/test/AggregationStrategy.test.ts#L42-L43)). Of course, you'd usually use a transfer to do that, but since we have the authority to mint tokens (because it's a test token), we can simply mint tokens into the aggregator's account.

contracts/clients/src/BlsProvider.ts
JohnGuilding: Right I see. I don't think I fully grasped that you could provide a 1 wei payment, **or** a 0 wei payment to tx.origin, and there were different tradeoffs to each.

~I've got two follow up questions:~

> ~0 wei has the advantage that if the aggregator doesn't charge fees...~

~1) To clarify on the above statement, using a 0 wei value doesn't work if the aggregator charges fees? (which they realistically would be)~

~2) So I like the idea of being able to estimate fees without having to add 1 wei. As it seems better UX to not require any funds to simulate txs. But if I'm following correctly and 0 wei only has an advantage when not charging fees, then using and documenting the use of 1 wei seems like the best default option as it's more realistic. So my question would be, **what should we recommend in our documentation?** Should we recommended and provide examples for 1 wei payments? Or should we recommend and provide examples for 1 wei and 0 wei payments, while explaining the tradeoffs for both? Or something else?~

contracts/clients/README.md
JohnGuilding: Ahhh gotcha, makes total sense now, thanks for the explanation

contracts/clients/src/BlsProvider.ts
JohnGuilding: Hmmm just tried running the integration tests locally with payment action values set to 0 wei, and they passed. This was with require fees set to true on the aggregator. So think my understanding that using a 0 wei payment only worked when not charging fees was wrong. Let me know if I'm missing anything still lol

In which case, I do quite like the idea that a user can estimate the gas cost for a transaction without requiring funds. In regards to emphasising that fees must be paid, and a 1 wei payment maybe being a better proxy for fees, my thoughts are:
- It seems more intuitive coming from the "ethers world" that you don't have to pay for anything when estimating transaction fees e.g. with [just a provider](https://docs.ethers.org/v5/api/providers/provider/#Provider-estimateGas). Don't want to fall into the trap of copying ethers without thinking about it, but in this example, paying 0 wei for fee estimation may be a more familiar experience for a lot of devs. As a counter argument, maybe this is me being biased coming from a background of mainly using ethers.
- Could we provide helper methods to add the fee payment action for estimating fees, and also the fee payment for sending the real bundle? This could help abstract this thinking about 0 wei vs 1 wei away from developers, and would result in less boilerplate code (Tbh this is basically what the bls provider does, with the caveat of being less granular & configurable, but that could be changed)

voltrevo: Yeah the devX concern should really be solved with helpful APIs, and shouldn't affect what strategy is used. The strategy should really only be determined by what's best for the end user.

I do think that 1 wei is a better proxy when it comes down to it. Probably the best of both worlds is to do this:
- If the user has funds, use 1 wei
- If the user doesn't have funds, use 0 wei, and then if the aggregator says a payment is needed then throw an error about insufficient funds (or at least a warning, and just make do with the possibly inaccurate estimate)

JohnGuilding: Yeah I like this approach. 

If creating the action is handled by the underlying logic, and the user just sees a helpful API, then I don't feel strongly about either option (0 wei vs 1 wei) :)

PR #536, opened 2023-02-15
Use BLSWallet and ProxyAdmin directly inside VerificationGateway

voltrevo:
# *Dependent PR*

This PR builds on top of #464. To reduce confusion, please either merge that PR first or retarget this branch to `contract-updates` to make it one PR that contains everything.

## What is this PR doing?

With the revelation that the 30m gas limit issue is a hardhat testing bug (which doesn't affect this branch) and that we therefore have plenty of headroom, this PR uses `BLSWallet` and `ProxyAdmin` directly inside `VerificationGateway`.

This reduces complexity, removes the `ProxyAdminGenerator` workaround entirely.

`VerificationGateway` now deploys with 4,798,984, so we still have 25m gas of headroom.

We could also set `runs: 1000` in the optimization settings, but it looks like the gas used by an operation actually goes *up* by doing that, so I've left it alone.

`yarn test`

Contributes to #394.

## Checklist

## Guidelines


jzaki (approved):

voltrevo:
Not sure why the aggregator and integration tests ran previously. This PR shouldn't be triggering them / those tests failing is *expected*.

Bumping seems to have fixed it 🤷‍♂️.

PR #537, opened 2023-02-15
Use BLS as library instead of contract

voltrevo:
# *Dependent PR*

This PR builds on top of #536. To reduce confusion, please either merge that PR first or retarget this branch to `contract-updates` to make it one PR that contains everything.

## What is this PR doing?

Uses BLS as a library instead of a contract. This means it doesn't need to be separately deployed and passed in to `VerificationGateway`, reducing complexity.

I have left the cost estimator as a separate contract as it looks like the idea is that there will be a precompile for this that could be substituted in future.

To make this work, I needed to upgrade the compiler version in the BLS contracts. Notably, this means that checked arithmetic is applied, but it hasn't broken and surprisingly the gas difference is rather small (maybe 1% - went through and wrapped everyting in `unchecked`).

`yarn test`

Contributes to #394 (kinda).

## Checklist

## Guidelines


contracts/contracts/lib/hubble-contracts/contracts/libs/BLS.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.6.12;
+pragma solidity >=0.8.4 <0.9.0;
```
jzaki, line 2: Was explicitly avoiding this since the hubble contracts were audited at that specific version. This is why there was an interface decoupling the lib's older solidity version from the VG's newer version.

jzaki (changes requested):

Issue #411, opened 2022-11-23
Add health endpoint to aggregator


jacque006:
**As an** aggregator operator,
**I would like** to know if an aggregator instance is running,
**So I can** take action when it goes down/offline/crashes

### Recommendation
- Add a `/health` endpoint to the aggregator that just responds with `HTTP 200`. If the request times out then the agg is down.

### Considerations
- As part of the health endpoint being called, we may also want to run 'pings' against our postgres & rpc connections to make sure they are alive as well.

voltrevo:
> just responds with HTTP 200

Also worth responding with `{}` in the body. This way clients can parse the response body as json without worrying about the empty special case. (It's common to respond to `/health` with some useful data.)

nooma-42:
I sent a draft PR to this issue, do I have to add a test for healthendpoint?

jacque006:
@NOOMA-42 Yes, please add a simple test to make sure the endpoint responds as expected. See existing aggregator tests for examples.

nooma-42:
I've just added a test

PR #539, opened 2023-02-19
Healthendpoint

nooma-42:
## What is this PR doing?
this PR adds a healthendpoints to check multiple service running or not

enter `http://localhost:YOURPORT/health`. `http://localhost:3000/health` in my case. Ideally, it shows `{"status":"HEALTHY","dependencies":[{"name":"DB","status":"HEALTHY"},{"name":"RPC","status":"HEALTHY"}]}`

issue #411 

## Checklist
- [X] I have manually tested these changes
- [X] Post a link to the PR in the group chat

## Guidelines

### DIFF
--- a/aggregator/deps.ts
+++ b/aggregator/deps.ts
@@ -7,6 +7,8 @@ export { oakCors } from "https://deno.land/x/cors@v1.2.0/mod.ts";
 import { config as dotEnvConfig } from "https://deno.land/x/dotenv@v2.0.0/mod.ts";
 export { dotEnvConfig };
 
+export { axiod } from "https://deno.land/x/axiod/mod.ts";
+
 // Oak framework dependencies
 export {
   Application,
--- a/aggregator/deps.ts
+++ b/aggregator/deps.ts
@@ -7,8 +7,6 @@ export { oakCors } from "https://deno.land/x/cors@v1.2.0/mod.ts";
 import { config as dotEnvConfig } from "https://deno.land/x/dotenv@v2.0.0/mod.ts";
 export { dotEnvConfig };
 
-export { axiod } from "https://deno.land/x/axiod/mod.ts";
-
 // Oak framework dependencies
 export {
   Application,
--- a/aggregator/src/app/AppEvent.ts
+++ b/aggregator/src/app/AppEvent.ts
@@ -100,7 +100,7 @@ type AppEvent =
       name: string;
       status: string;
       detail: string;
-    }
+    };
   }
   ;
 
--- a/aggregator/src/app/AppEvent.ts
+++ b/aggregator/src/app/AppEvent.ts
@@ -87,21 +87,6 @@ type AppEvent =
       duration: number;
     };
   }
-  | {
-    type: "service-healthy";
-    data: {
-      name: string;
-      status: string;
-    };
-  } 
-  | {
-    type: "service-unhealthy";
-    data: {
-      name: string;
-      status: string;
-      detail: string;
-    };
-  }
   ;
 
 
--- a/aggregator/src/app/AppEvent.ts
+++ b/aggregator/src/app/AppEvent.ts
@@ -86,8 +86,7 @@ type AppEvent =
       status: number;
       duration: number;
     };
-  }
-  ;
+  };
 
 
 export default AppEvent;
--- a/aggregator/src/app/AppEvent.ts
+++ b/aggregator/src/app/AppEvent.ts
@@ -86,6 +86,23 @@ type AppEvent =
       status: number;
       duration: number;
     };
-  };
+  }
+  | {
+    type: "service-healthy";
+    data: {
+      name: string;
+      status: string;
+    };
+  } 
+  | {
+    type: "service-unhealthy";
+    data: {
+      name: string;
+      status: string;
+      detail: string;
+    }
+  }
+  ;
+
 
 export default AppEvent;
--- a/aggregator/src/app/BundleRouter.ts
+++ b/aggregator/src/app/BundleRouter.ts
@@ -2,7 +2,6 @@ import { Router } from "../../deps.ts";
 import failRequest from "./helpers/failRequest.ts";
 import BundleHandler from "./helpers/BundleHandler.ts";
 import nil from "../helpers/nil.ts";
-
 import BundleService from "./BundleService.ts";
 
 export default function BundleRouter(bundleService: BundleService) {
@@ -39,6 +38,11 @@ export default function BundleRouter(bundleService: BundleService) {
       ctx.response.body = bundleService.receiptFromBundle(bundleRow);
     },
   );
+
+  router.get("bundle/health", (ctx) => {
+    ctx.response.status = 200;
+    ctx.response.body = { status: "OK" };
+  });
 
   return router;
 }
--- a/aggregator/src/app/BundleRouter.ts
+++ b/aggregator/src/app/BundleRouter.ts
@@ -39,10 +39,5 @@ export default function BundleRouter(bundleService: BundleService) {
     },
   );
 
-  router.get("bundle/health", (ctx) => {
-    ctx.response.status = 200;
-    ctx.response.body = { status: "OK" };
-  });
-
   return router;
 }
--- a/aggregator/src/app/HealthRouter.ts
+++ b/aggregator/src/app/HealthRouter.ts
@@ -1,0 +1,31 @@
+import { Router } from "../../deps.ts";
+import {HealthService} from "./HealthService.ts";
+
+export default function healthRouter(healthService: HealthService) {
+  const router = new Router({ prefix: "/" });
+
+  router.get("health", async (ctx) => {
+    const healthResults = await healthService.getHealth();
+    console.log(`Status: ${healthResults.ResourceHealth}\n`);
+    ctx.response.headers.set("Content-Type", "application/json");
+    ctx.response.body = healthResults.ResourceHealth;
+    // ctx.response.status = healthResults.status === ResourceHealth.Healthy ? 200 : 503
+    /* 
+        res.status(healthResults.status === ResourceHealth.Healthy ? 200 : 503)
+      .send({
+        status: healthResults.status, dependencies: healthResults.results
+      });
+    */
+  });
+  return router;
+}
+
+/* 
+
+  const healthService = new HealthService(
+    [
+      new SomeServiceCheck(),
+      // Add more checks here...
+    ]
+  );
+*/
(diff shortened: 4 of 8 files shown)


jacque006:
Thanks for the contribution @NOOMA-42 ! I'm at an event this week, but will try and take look. Worst case will review early next week.

aggregator/deps.ts
```diff
@@ -7,6 +7,8 @@ export { oakCors } from "https://deno.land/x/cors@v1.2.0/mod.ts";
 import { config as dotEnvConfig } from "https://deno.land/x/dotenv@v2.0.0/mod.ts";
 export { dotEnvConfig };
 
+export { axiod } from "https://deno.land/x/axiod/mod.ts";
```
voltrevo: I think we should just use the native `fetch` API. What's the reason for introducing `axiod`?

aggregator/src/app/HealthRouter.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/* 
+
+  const healthService = new HealthService(
+    [
+      new SomeServiceCheck(),
+      // Add more checks here...
+    ]
+  );
+*/
```
voltrevo: Please don't include commented code

aggregator/src/app/HealthService.ts
```diff
@@ -0,0 +1,192 @@
...
+}
+
+export class AggregatorServiceHealthCheck extends HealthIndicator {
+  // Starts out in the Unhealthy state by default until it can be verified as Healthy
+  name = 'Aggregator';
+  status: ResourceHealth = ResourceHealth.Unhealthy;
+  
+  async checkHealth(): Promise<void> {
+    try {
+      const response = await axiod.get(new URL(env.ORIGIN, "/Bundle/health").toString());
```
voltrevo: I don't like that we're doing a request that loops back to the aggregator here. The ability to make a request to the aggregator is already covered by the fact that the health check itself is provided as a response to a request.

If the bundle service has something to say about the health of the aggregator then we should just read that internally. And currently it doesn't have anything to say - it just says "OK".

Overall, my read on this is that `AggregatorServiceHealthCheck` isn't adding anything, and should be removed. Unless I'm missing something?

aggregator/src/app/HealthService.ts
```diff
@@ -0,0 +1,192 @@
...
+  abstract name: string;
+  status: ResourceHealth = ResourceHealth.Unhealthy;
+  details: string | undefined;
+  abstract checkHealth(): Promise<void> | void;
+}
+
+export enum ResourceHealth {
+  Healthy = 'HEALTHY',
+  Unhealthy = 'UNHEALTHY'
+}
```
voltrevo: We don't really use enums... I suggest this should be replaced with something like:

```ts
export type ResourceHealth = "healthy" | "unhealthy";
```

Open to discussing this further, especially if you have a concrete reason for preferring the enum version for these changes.

aggregator/src/app/HealthService.ts
```diff
@@ -0,0 +1,192 @@
...
+
+abstract class HealthIndicator {
+  constructor(
+    public emit: (evt: AppEvent) => void,
+  ){}
+  abstract name: string;
+  status: ResourceHealth = ResourceHealth.Unhealthy;
+  details: string | undefined;
+  abstract checkHealth(): Promise<void> | void;
+}
```
voltrevo: We don't use any class inheritance yet and should continue to avoid it unless there's a compelling reason to introduce it.

I think this can just be a type:

```ts
type HealthIndicator = {
  name: string;
  status: ResourceHealth;
  details?: string;
  checkHealth(): Promise<void> | void;
};
```

Then you can write:

```ts
class ExampleHealthIndicator implements HealthIndicator { ... }
```

Instead of:

```ts
class ExampleHealthIndicator extends HealthIndicator { ... }
```

Related: https://www.youtube.com/watch?v=hxGOiiR9ZKg

aggregator/src/app/HealthService.ts
```diff
@@ -0,0 +1,192 @@
...
+
+export class RPCServiceHealthCheck extends HealthIndicator  {
+  // Starts out in the Unhealthy state by default until it can be verified as Healthy
+  name = 'RPC';
+  status: ResourceHealth = ResourceHealth.Unhealthy;
+
+  async checkHealth(): Promise<void> {
+    try {
+      const response = await axiod.get(env.RPC_URL);
```
voltrevo: Good idea checking the RPC. We should probably send them a concrete request though, I'm not sure whether all providers will respond nicely to a blank request.

You can get chainId by using this for the body:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "eth_chainId",
  "params": []
}
```

aggregator/src/app/HealthService.ts
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  // Starts out in the Unhealthy state by default until it can be verified as Healthy
+  name = 'DB';
+  status: ResourceHealth = ResourceHealth.Unhealthy;
+
+  checkHealth(): void {
+    // wrap synchronous code in a promise so we can use async/await
+      try {      
+        const [[now]] = [...this.bundleTable.dbQuery("SELECT STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW')")];
```
voltrevo: Unnecessary spread
```suggestion
        const [[now]] = this.bundleTable.dbQuery("SELECT STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW')");
```

voltrevo (changes requested):
Hi @NOOMA-42, it's great to have you here!

I hope my feedback isn't too discouraging, we'd absolutely like to get some version of this merged in.

Overall, I'd also like to see this design simplified a bit:
- Health checks should just return their results rather than storing them
- It would be simpler to handle health check exceptions in the health service rather than have each health check add a similar try-catch wrapper

Having said that, I think it would also be productive to first do a much simpler change: create a health service exposing `/health` that just unconditionally responds `{status:"healthy"}`. The idea being that we have some indication of health just by being able to get that response.

That way we can get the ball rolling with some code from you merged in as soon as possible, and we can follow-up with an injected array of health checks based on the ideas you've developed here. Wdyt?

nooma-42:
> That way we can get the ball rolling with some code from you merged in as soon as possible, and we can follow-up with an injected array of health checks based on the ideas you've developed here. Wdyt?

Thanks for your review! I'll work on this during weekend. I totally agree with you that a simpler change first. I'll modified the code based on your code review and the generally simpler design. 
Just to make sure that by the injected array, are you saying this kind of design logic?
```
services.AddHealthChecks()
    .AddCheck<DatabaseHealthCheck>("database")
    .AddCheck<ApiHealthCheck>("api")
    .AddCheck<ConfigHealthCheck>("config");
```

voltrevo:
@NOOMA-42 Awesome, look forward to it.

For the injection, I prefer the way you're currently doing it:

```ts
  const healthService = new HealthService(
    emit,
    [
      dbServiceHealthCheck,
      rpcServiceHealthCheck,
    ]
  )
```

In other words, injection means just passing them in as a parameter.

voltrevo:
@NOOMA-42 Thanks for the update. It looks like this still includes a lot of changes. Let's only do the minimal version first as mentioned previously:

> Having said that, I think it would also be productive to first do a much simpler change: create a health service exposing /health that just unconditionally responds {status:"healthy"}. The idea being that we have some indication of health just by being able to get that response.
>
> That way we can get the ball rolling with some code from you merged in as soon as possible, and we can follow-up with an injected array of health checks based on the ideas you've developed here. Wdyt?

Please either remove the other code or open a new PR with only this change.

nooma-42:
Is this modification small enough? or does it need to be simpler? 
By the way, I rebase this PR based on most recent commit, but I made many unnecessary commits. Do I need to squash the commits? (Never do this before, but find a relevant [article](https://stackoverflow.com/questions/5667884/how-to-squash-commits-in-git-after-they-have-been-pushed))

aggregator/test/HealthService.test.ts
```diff
@@ -0,0 +1,10 @@
+import { assertEquals } from "./deps.ts";
+
+import Fixture from "./helpers/Fixture.ts";
+
+Fixture.test("HealthService returns healthy", async (fx) => {
+  const healthCheckService = fx.createHealthCheckService()
+  const healthStatus = await healthCheckService.getHealth();
+  const expected = {"status":"HEALTHY"};
```
voltrevo: Should be lowercase. Test failure here: https://github.com/web3well/bls-wallet/actions/runs/4593302809/jobs/8111858924?pr=539#step:9:686.

aggregator/src/app/HealthService.ts
```diff
@@ -0,0 +1,17 @@
...
+export default class HealthService {
+  public overallHealth: ResourceHealth = 'healthy';
+
+  getHealth(): Promise<HealthCheckResult> {
+    this.overallHealth = 'healthy';
+          
+    return Promise.resolve({
+      status: this.overallHealth,
+    });
+  }
```
voltrevo: Not sure why we should store `overallHealth`. Let's compute this every time unless we decide to optimize it in future.

```suggestion
  getHealth(): Promise<HealthCheckResult> {
    return Promise.resolve({ status: 'healthy' });
  }
```

aggregator/src/app/BundleRouter.ts
```diff
@@ -40,5 +39,10 @@ export default function BundleRouter(bundleService: BundleService) {
     },
   );
 
+  router.get("bundle/health", (ctx) => {
+    ctx.response.status = 200;
+    ctx.response.body = { status: "OK" };
+  });
+
```
voltrevo: Did you mean to include this?

```suggestion
```

aggregator/src/app/AppEvent.ts
```diff
@@ -86,6 +86,8 @@ type AppEvent =
       status: number;
       duration: number;
     };
-  };
+  }
+  ;
+
```
voltrevo: Style.

```suggestion
  };

```

voltrevo (changes requested):

voltrevo:
> Is this modification small enough? or does it need to be simpler?

Small enough 👍. Have a few things for you to fix up though. See above.

>By the way, I rebase this PR based on most recent commit, but I made many unnecessary commits. Do I need to squash the commits? (Never do this before, but find a relevant [article](https://stackoverflow.com/questions/5667884/how-to-squash-commits-in-git-after-they-have-been-pushed))

You could rewrite your history to include only commits with relevant changes, but if you're happy to squash it into a single commit we can just do that for you using the GitHub UI 🙌.

nooma-42:
I just fix them up. 
It'll be great if you can squash it for me!

voltrevo (approved):

voltrevo:
Merged! 🎉

Thanks for your patience @NOOMA-42, it's great to have your contribution included. Looking forward to your followup with the injected checks :).

PR #540, opened 2023-02-21
Add multi-action send transaction method to bls signer

### LINKED ISSUE #375: Add multi-action transactions to BlsProvider & BlsSigner
As part of the work to create a BlsProvider and BlsSigner, we should create a way for a user to send multi-action transactions. This work should be undertaken after we have added functionality to send BLS transactions using the BlsProvider and BlsSigner.

There are methods to send transactions using the provider and signer, so both classes will require changes.

Note: not including sending regular BLS transactions, the remaining major BLS Wallet features such as recovery and gasless transactions are deemed out of scope for the Bls provider/signer MVP.

JohnGuilding:
## What is this PR doing?
Adds multi-action transactions to bls provider and bls signer.

New methods were added as to not break the ethers provider base-class function signatures:
```ts
async sendTransactionBatch(transactionBatch: TransactionBatch): Promise<TransactionBatchResponse> {}

async constructTransactionBatchResponse(actions: Array<ActionData>, hash: string, from: string, nonce?: BigNumber): Promise<TransactionBatchResponse> {}

async signTransactionBatch(transactionBatch: TransactionBatch): Promise<string> {}

async sendTransactionBatch(signedTransactionBatch: string): Promise<TransactionBatchResponse> {}
```

### multi-action standards

TLDR ended up going for the [wallet_batchTransactions standard](https://gist.github.com/mikhailxyz/773ba5196b5a7046423592204e5bc5c4). It was the simplest option and we can easily build upon it as future requirements come up.
~**TLDR**: Gut feeling is EIP 5792 is the option to go for. But after reading the discussion on eth magicians, it feels like there is some pushback which makes me think it won't get adopted in it's current state. e.g. disagreements about standardisation of bundle identifiers and disagreement about level of 'atomicity'.~

New types:

```ts
export type TransactionBatch = {
  transactions: Array<Transaction>;
  batchOptions?: BatchOptions;
};
```

These were based on this [draft batch transaction standard](https://gist.github.com/mikhailxyz/773ba5196b5a7046423592204e5bc5c4) - `wallet_batchTransactions` JSON-RPC method. The idea was to ensure our code abides by a spec if one exists. There is no formal standard and it was a choice between the `wallet_batchTransactions` draft and [EIP 5792](https://eips.ethereum.org/EIPS/eip-5792) (draft). While the draft EIP is further along in it's journey towards becoming a standard (by this, I mean it's an actual draft EIP with discussion, as opposed to an informal draft), it's had some pushback in it's eth magicians thread. I've gone for the `wallet_batchTransactions` option for simplicity.

I've also created a new response type, this isn't based on any of the above standards, but it was easy to implement.
```ts
interface TransactionBatchResponse {
  transactions: Array<ethers.providers.TransactionResponse>;
  awaitBatchReceipt: (
    confirmations?: number,
  ) => Promise<ethers.providers.TransactionReceipt>;
}
```
___

## Further notes and other changes

1. Preliminary decision with `signer.sendTransactionBatch` was to only add transactions to a single operation. This means that all transactions in the batch will be atomic. Another option may be to give the developer a choice on whether to add transactions to different operations, and thus permit non-atomic transaction flows. Non-atomic tx flows are out of scope for this work.

2. provider.sendTransactionBatch operates on an already signed bundle, so could theoretically receive multiple operations. I've added some code to support this in case of that scenario:

```ts
    const actionData: Array<ActionData> = bundle.operations
      .map((operation) => operation.actions)
      .flat();
```

3. Abstracted common fee estimate code into helper functions:
    - Provider. _addFeePaymentActionForFeeEstimation
    - Provider. _addFeePaymentActionWithSafeFee
    - addSafetyDivisorToFee

- run integration tests in workflow or locally

Resolves https://github.com/web3well/bls-wallet/issues/375

## Checklist

## Guidelines

### DIFF
--- a/contracts/clients/src/BlsSigner.ts
+++ b/contracts/clients/src/BlsSigner.ts
@@ -416,6 +416,83 @@ export default class BlsSigner extends Signer {
     return JSON.stringify(bundleToDto(bundle));
   }
 
+  async signTransactionBatch(
+    transactionBatch: TransactionBatch,
+  ): Promise<string> {
+    await this.initPromise;
+
+    const actions: Array<ActionData> = transactionBatch.transactions.map(
+      (transaction) => {
+        if (!transaction.to) {
+          throw new TypeError(
+            "Transaction.to should be defined for all transactions",
+          );
+        }
+
+        return {
+          ethValue: transaction.value?.toString() ?? "0",
+          contractAddress: transaction.to!.toString(),
+          encodedFunction: transaction.data?.toString() ?? "0x",
+        };
+      },
+    );
+
+    const nonce = await BlsWalletWrapper.Nonce(
+      this.wallet.PublicKey(),
+      this.verificationGatewayAddress,
+      this.provider,
+    );
+
+    const aggregatorUtilitiesContract = AggregatorUtilities__factory.connect(
+      this.aggregatorUtilitiesAddress,
+      this.provider,
+    );
+
+    const feeEstimate = await this.provider.aggregator.estimateFee(
+      this.wallet.sign({
+        nonce,
+        actions: [
+          ...actions,
+          {
+            ethValue: 1,
+            // Provide 1 wei with this action so that the fee transfer to
+            // tx.origin can be included in the gas estimate.
+            contractAddress: this.aggregatorUtilitiesAddress,
+            encodedFunction:
+              aggregatorUtilitiesContract.interface.encodeFunctionData(
+                "sendEthToTxOrigin",
+              ),
+          },
+        ],
+      }),
+    );
+
+    // Due to small fluctuations is gas estimation, we add a little safety premium
+    // to the fee to increase the chance that it actually gets accepted during
+    // aggregation.
+    const safetyDivisor = 5;
+    const feeRequired = BigNumber.from(feeEstimate.feeRequired);
+    const safetyPremium = feeRequired.div(safetyDivisor);
+    const safeFee = feeRequired.add(safetyPremium);
+
+    const bundle = this.wallet.sign({
+      nonce,
+      actions: [
+        ...actions,
+        {
+          ethValue: safeFee,
+          contractAddress: this.aggregatorUtilitiesAddress,
+          encodedFunction:
+            aggregatorUtilitiesContract.interface.encodeFunctionData(
+              "sendEthToTxOrigin",
+            ),
+        },
+      ],
+    });
+
+    return JSON.stringify(bundleToDto(bundle));
+  }
+
   /** Sign a message */
   // TODO: Come back to this once we support EIP-1271
   override async signMessage(message: Bytes | string): Promise<string> {
--- a/contracts/clients/src/BlsSigner.ts
+++ b/contracts/clients/src/BlsSigner.ts
@@ -1,6 +1,13 @@
 /* eslint-disable camelcase */
-import { ethers, BigNumber, Signer, Bytes } from "ethers";
-import { Deferrable, hexlify, isBytes, RLP } from "ethers/lib/utils";
+import { ethers, BigNumber, Signer, Bytes, BigNumberish } from "ethers";
+import {
+  AccessListish,
+  BytesLike,
+  Deferrable,
+  hexlify,
+  isBytes,
+  RLP,
+} from "ethers/lib/utils";
 import { AggregatorUtilities__factory } from "../typechain-types";
 
 import BlsProvider from "./BlsProvider";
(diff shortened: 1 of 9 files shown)


contracts/clients/src/BlsSigner.ts
```diff
@@ -1,14 +1,77 @@
...
+  accessList?: Array<AccessListish>;
+};
+
+/**
+ * @param transactions - an array of transaction objects
+ * @param batchOptions - optional batch options taken into account by SC wallets
+ */
+export type TransactionBatch = {
+  transactions: Array<Transaction>;
+  batchOptions?: BatchOptions;
```
JohnGuilding, line 41: Not really utilising this optional property. Tbh slightly confused by what it's purpose is from the [draft standard](https://gist.github.com/mikhailxyz/773ba5196b5a7046423592204e5bc5c4). My interpretation is that these values help smart contract wallets deal with figuring out batch transaction options from multiple transactions, but not sure if that's correct

contracts/clients/src/BlsSigner.ts
```diff
@@ -1,14 +1,77 @@
...
+  to?: string;
+  value?: BigNumberish;
+  gas?: BigNumberish;
+  maxPriorityFeePerGas?: BigNumberish;
+  maxFeePerGas?: BigNumberish;
+  data?: BytesLike;
+  nonce?: BigNumberish;
+  chainId?: number;
+  accessList?: Array<AccessListish>;
+};
```
jacque006: You should be able to import this from ethers.js

contracts/clients/src/BlsSigner.ts
jacque006: I think using this spec we need to accept them, but some of them we will be unable to pass forward to the agg. `gas` can be used with latest contract changes in `contract-updates`.

contracts/clients/src/BlsSigner.ts
```diff
@@ -119,6 +170,70 @@ export default class BlsSigner extends Signer {
...
+  async sendTransactionBatch(
+    transactionBatch: TransactionBatch,
+  ): Promise<TransactionBatchResponse> {
+    await this.initPromise;
+
+    const actions: Array<ActionData> = transactionBatch.transactions.map(
+      (transaction) => {
+        if (!transaction.to) {
+          throw new TypeError(
+            "Transaction.to should be defined for all transactions",
```
```suggestion
    const actions: Array<ActionData> = transactionBatch.transactions.map(
      (transaction, i) => {
        if (!transaction.to) {
          throw new TypeError(
            `Transaction.to missing on txn {i}`,
```

contracts/clients/src/BlsSigner.ts
```diff
@@ -202,24 +360,68 @@ export default class BlsSigner extends Signer {
...
+  async signTransactionBatch(
+    transactionBatch: TransactionBatch,
+  ): Promise<string> {
+    await this.initPromise;
+
+    const actions: Array<ActionData> = transactionBatch.transactions.map(
+      (transaction) => {
+        if (!transaction.to) {
+          throw new TypeError(
+            "Transaction.to should be defined for all transactions",
```
jacque006: Same here.

jacque006 (commented):
Overall, think this is a great start on batch txn handling for provider (nice tests btw). Also sets us up well to add additional batch specs. Let's sync offline on how to handle some of the batch options.

To start, I think we can assume that all actions (batched txns) are atomic. We can add more logic later to split actions across multiple operations later.

contracts/test-integration/BlsSigner.test.ts
```diff
@@ -160,6 +160,313 @@ describe("BlsSigner", () => {
...
+      accessList: [],
+    };
+
+    // Act
+    const result = await blsSigner._validateBatchOptions(batchOptions);
+
+    // Assert
+    expect(result).to.deep.equal(batchOptions);
+    expect(result.nonce).to.equal(BigNumber.from("39"));
+    expect(result.nonce).to.have.property("add");
```
JohnGuilding, line 443: `expect(result.nonce).to.equal(BigNumber.from("39"));` can result in false positives, so these 2 extra assertions test "BigNumber.from" worked.
(e.g. before these extra assertions, the test will still pass if "BigNumber.from" is removed in the method which is incorrect behaviour)

contracts/test-integration/BlsProviderContractInteraction.test.ts
```diff
@@ -161,6 +161,9 @@ describe("Provider tests", function () {
         .transfer(recipient, amountToTransfer);
       await tx.wait();
 
+      // wait 1 second to ensure listener count updates
+      await new Promise((resolve) => setTimeout(resolve, 1000));
```
JohnGuilding, line 165: When running the whole test suite locally, the `expect(blsProvider.listenerCount(filter)).to.equal(0);` assertion would periodically fail as the event listener was still subscribed. Seems a bit hacky for testing a function that is supposed to be synchronous, but the timer appears to fix that issue.

contracts/clients/src/BlsSigner.ts
JohnGuilding: Done. Added a helper function (_validateBatchOptions) that validates the batch options if they are passed as an argument. It checks the chain id is correct and ensures the nonce is the correct type.

jacque006 (approved):
:+1: We we decide to promote this for general usage, we should make sure to document the batch function & spec so people are aware they can use that in addition to the normal provider/signer APIs

PR #541, opened 2023-02-28
Improve private key management in BlsWalletWrapper

### LINKED ISSUE #499: Create new private key and signing management class

JohnGuilding:
## What is this PR doing?
This PR modifies private key management in the `BlsWalletWrapper` class to:
- improve data abstraction
- improve security
- create a more unified abstraction.

Places where private key is removed:
- Directly in the `BlsWalletWrapper ` constructor
- `getPublicKeyHash()`
- `sign()`
- `signMessage()`
- `PublicKey()`
- `PublicKeyStr()`
- `validateWalletNotRecovered`

Places where privateKey is introduced:
- `initBlsWalletSigner` the private that would have been part of the BlsWalletWrapper class, is now passed to the BlsWalletSigner. So init method now takes in a private key. 
- `#BlsWalletSigner` returns a `BlsWalletSigner`, which takes a private key, so this private method also does that now. The only place it is used is in the static `Address` method, which already takes a private key, so it doesn't actually change how an end user would call the method.

Also added a `setBlsWalletSigner` helper method. This was to ensure the `should recover blswallet via blswallet to new bls key using bls client module` recovery test still passed as a new public key needed generating.

1. Running all test suites that use the client module (contracts, clients, aggregator, integration)
2. Sending a transaction with the browser wallet
3. Recovering an account from the browser wallet to quill

#499 

## Checklist

## Guidelines

### DIFF
--- a/contracts/clients/src/BlsWalletWrapper.ts
+++ b/contracts/clients/src/BlsWalletWrapper.ts
@@ -64,16 +64,14 @@ export default class BlsWalletWrapper {
    * @param privateKey private key associated with the wallet
    * @param verificationGatewayAddress address of the VerficationGateway contract
    * @param signerOrProvider ethers.js Signer or Provider
-   * @param blsWalletSigner (optional) a BLS Wallet signer
    * @returns The wallet's address
    */
   static async Address(
     privateKey: string,
     verificationGatewayAddress: string,
     signerOrProvider: SignerOrProvider,
-    blsWalletSigner?: BlsWalletSigner,
   ): Promise<string> {
-    blsWalletSigner ??= await this.#BlsWalletSigner(
+    const blsWalletSigner = await this.#BlsWalletSigner(
       signerOrProvider,
       privateKey,
     );
--- a/contracts/clients/src/BlsWalletWrapper.ts
+++ b/contracts/clients/src/BlsWalletWrapper.ts
@@ -330,6 +330,11 @@ export default class BlsWalletWrapper {
     return await initBlsWalletSigner({ chainId, privateKey });
   }
 
+  /**
+   * Binds the BlsWalletSigner instance to a new private key and chainId
+   *
+   * @returns The updated BlsWalletSigner object
+   */
   async setBlsWalletSigner(
     signerOrProvider: SignerOrProvider,
     privateKey: string,
--- a/contracts/clients/src/BlsWalletWrapper.ts
+++ b/contracts/clients/src/BlsWalletWrapper.ts
@@ -157,12 +157,9 @@ export default class BlsWalletWrapper {
     return blsWalletWrapper;
   }
 
-  async syncWallet(
-    verificationGateway: VerificationGateway,
-    privateKey: string,
-  ) {
+  async syncWallet(verificationGateway: VerificationGateway) {
     this.address = await BlsWalletWrapper.Address(
-      privateKey,
+      this.blsWalletSigner.privateKey,
       verificationGateway.address,
       verificationGateway.provider,
     );
--- a/contracts/clients/src/BlsWalletWrapper.ts
+++ b/contracts/clients/src/BlsWalletWrapper.ts
@@ -30,7 +30,6 @@ export default class BlsWalletWrapper {
   public address: string;
   private constructor(
     public blsWalletSigner: BlsWalletSigner,
-    public privateKey: string,
     public walletContract: BLSWallet,
   ) {
     this.address = walletContract.address;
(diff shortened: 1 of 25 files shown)


contracts/test/recovery-test.ts
```diff
@@ -327,10 +330,13 @@ describe("Recovery", async function () {
       recoveredWalletNonce++, // await wallet2.Nonce(),
     );
 
+    console.log("here1");
     expect(await vg.walletFromHash(hash1)).to.not.equal(blsWallet.address);
+    console.log("here2");
     expect(await vg.walletFromHash(hashAttacker)).to.not.equal(
```
JohnGuilding: This is the line this test fails on @blakecduncan

JohnGuilding: Fixed by using the attacker's `blsWalletSigner` to create the `attackerHash`, instead of wallet 2s signer.

contracts/clients/src/BlsWalletWrapper.ts
```diff
@@ -65,22 +64,23 @@ export default class BlsWalletWrapper {
...
    * @param verificationGatewayAddress address of the VerficationGateway contract
    * @param signerOrProvider ethers.js Signer or Provider
-   * @param blsWalletSigner (optional) a BLS Wallet signer
    * @returns The wallet's address
    */
   static async Address(
     privateKey: string,
     verificationGatewayAddress: string,
     signerOrProvider: SignerOrProvider,
-    blsWalletSigner?: BlsWalletSigner,
```
JohnGuilding, line 75: Previously, the `privateKey` argument was used to derive the `pubKeyHash` and run `validateWalletNotRecovered`. We no longer need to pass the `blsWalletSigner` in, as we can achieve the same by just passing the `privateKey` and `signerOrProvider` into a new instance of a `BlsWalletSigner`.

Another reason to remove it would be, to avoid scenarios where the `blsWalletSigner` passed into this method is bound to a different `privateKey` than the dedicated `privateKey` argument. Which, may be confusing as you'd essentially be passing in 2 private keys to the method, and the address would be dervied from the `blsWalletSigner`'s private key, not the dedicated `privateKey` arg.

___

I was hesitant about doing this first, as it seemed a more significant change to the API, but will raise for discussion:

Option 2 would be to keep in line with the data abstraction theme of this PR and remove the `privateKey` arg and do something like this:

```ts
static async Address(
    blsWalletSigner: BlsWalletSigner,
    verificationGatewayAddress: string,
    signerOrProvider: SignerOrProvider,
  ): Promise<string> {
    const verificationGateway = VerificationGateway__factory.connect(
      verificationGatewayAddress,
      signerOrProvider,
    );
    const pubKeyHash = blsWalletSigner.getPublicKeyHash();

    const existingAddress = await verificationGateway.walletFromHash(
      pubKeyHash,
    );
    const hasExistingAddress = !BigNumber.from(existingAddress).isZero();
    if (hasExistingAddress) {
      return existingAddress;
    }

    const expectedAddress = await this.ExpectedAddress(
      verificationGateway,
      pubKeyHash,
    );
    this.validateWalletNotRecovered(
      blsWalletSigner,
      verificationGateway,
      expectedAddress,
    );

    return expectedAddress;
}
```

voltrevo:
Approved ✅

(Forgot to formally do this before using the merge button... (I thought there was supposed to be a block/warning about that 🤔))

PR #548, opened 2023-03-08
413 improve private key management in bls provider and signer

### LINKED ISSUE #427: Using new provider w/ unchecked signer fails to find txn receipt
Ref: https://github.com/web3well/bls-wallet/pull/422/files#r1048836078

Update provider/signer code and test case such that a new provider w/ an unchecked signer successfully finds a txn receipt.

JohnGuilding:
## What is this PR doing?
This PR removes all references to the signer in the BlsProvider. This reduces the tight coupling between the two classes, better represents a JsonRpcProvider, and we can remove code like:

```ts
if (!this.signer) {
      throw new Error("Call provider.getSigner first");
}
```

Other changes:
1. Moved the constructTransactionResponse and constructTransactionBatchResponse methods to the provider. This consolidates the logic in one place. 
2. Added _validateTransaction and _validateTransactionBatch methods in the signer to split validation logic. This reduces the complexity of the send transaction methods and makes testing easier.
3. Moved some tests from integration tests to ./clients as they don't require a running network to pass.
4. Removed some regular `JsonRpcProvider` and `JsonRpcSigner` tests at the bottom of the BlsProvider.test.ts and BlsSigner.test.ts files.
5. Cleans up provider and signer code and test code

run integration tests and clients tests

#413 
Also fixes #427. Turns out I wasn't funding the new signer in the test. 

## Checklist

## Guidelines


contracts/clients/src/BlsProvider.ts
```diff
@@ -39,34 +44,39 @@ export default class BlsProvider extends ethers.providers.JsonRpcProvider {
...
-      this.verificationGatewayAddress,
-      this,
+    const nonce = await this.getTransactionCount(
+      resolvedTransaction.from.toString(),
     );
 
     const actionWithFeePaymentAction =
       this._addFeePaymentActionForFeeEstimation([action]);
 
+    const throwawayPrivateKey = await BlsWalletWrapper.getRandomBlsPrivateKey();
```
JohnGuilding, line 75: Open to suggestions on being able to estimate fees in the provider **without having reference to a signer**. This was my hacky solution. It does result in slightly different gas estimates each time, but still close enough to send the transaction successfully. 

See [this comment ](https://github.com/web3well/bls-wallet/pull/548/files#r1139122179)for examples of differences in tests without mocking a consistent private key value.

Not a fan of creating a throwaway BlsWalletWrapper, but feels slightly better than having a signer tightly coupled to a provider.

contracts/test-integration/BlsSigner.test.ts
```diff
@@ -567,6 +510,14 @@ describe("BlsSigner", () => {
       blsSigner,
     );
 
+    // BlsWalletWrapper.getRandomBlsPrivateKey from "estimateGas" method results in slightly different
+    // fee estimates. Which leads to a different signature. This fake avoids signature mismatch by stubbing a constant value.
+    sinon.replace(
```
JohnGuilding, line 515: Example of differences observed without sinon.fake:


Test | Expected Fee Estimate | Fee Estimate
-- | -- | --
1 | 0.000079145258744232 | 0.000077749364772554
2 | 0.00006806648707833 | 0.000069299213638008
3 | 0.000060669559559918 | 0.000060666067046826

<br class="Apple-interchange-newline">

contracts/clients/src/BlsProvider.ts
```diff
@@ -295,4 +286,104 @@ export default class BlsProvider extends ethers.providers.JsonRpcProvider {
...
+  ): Promise<TransactionBatchResponse> {
+    const chainId = await this.send("eth_chainId", []);
+    const verificationGateway = VerificationGateway__factory.connect(
+      this.verificationGatewayAddress,
+      this,
+    );
+
+    const transactions: Array<ethers.providers.TransactionResponse> = [];
+
+    for (const publicKeyLinkedToActions of publicKeysLinkedToActions) {
```
JohnGuilding, line 349: It is possible in the provider to get a bundle with multiple public keys and operations, this logic ensures all of the transactions in `transactions: Array<ethers.providers.TransactionResponse> = []` are linked to the correct `nonce` and `from` address.

e.g. if I just mapped across all of the actions like I was before, then there would be no way of telling which public key was linked to which group of actions.

contracts/clients/src/BlsSigner.ts
```diff
@@ -464,6 +390,54 @@ export default class BlsSigner extends Signer {
...
+    const resolvedTransaction = await ethers.utils.resolveProperties(
+      transaction,
+    );
+
+    if (!resolvedTransaction.to) {
+      throw new TypeError("Transaction.to should be defined");
+    }
+
+    if (!resolvedTransaction.from) {
+      resolvedTransaction.from = await this.getAddress();
```
JohnGuilding, line 405: This is a subtle, but important change, it ensures we don't require the following logic in many places and have to enforce adding `from` to transaction requests:

```ts
if (!resolvedTransaction.from) {
   throw new TypeError("Transaction.from should be defined");
}
```

This is because the provider `estimateGas` method now requires `from` to calculate the nonce, and the transaction that this method is validating will get pass to that method later in the send transaction methods. See changes to estimateGas and [this comment ](https://github.com/web3well/bls-wallet/pull/548/files#r1139154257)for more context

contracts/test-integration/BlsProvider.test.ts
```diff
@@ -85,11 +85,11 @@ describe("BlsProvider", () => {
...
     // Arrange
-    const recipient = ethers.Wallet.createRandom().address;
-    const transactionAmount = parseEther("1");
+    const getAddressPromise = blsSigner.getAddress();
     const transactionRequest = {
-      to: recipient,
-      value: transactionAmount,
+      to: ethers.Wallet.createRandom().address,
+      value: parseEther("1"),
+      from: getAddressPromise,
```
JohnGuilding, line 92: `from` needed to be added manually here as calling estimateGas directly. The _validateTransaction method on the signer does this work for us in cases where the transaction comes from the signer.

contracts/clients/src/BlsProvider.ts
jacque006: I think for now this is fine. I would document why your are doing this in code.

We should have an estimation endpoint/method for v2 that does not require a signed bundle.

jacque006 (approved):

contracts/clients/src/BlsProvider.ts
JohnGuilding: Added a comment and tracked via a follow up issue for V2 https://github.com/web3well/bls-wallet/issues/560

Issue #523, opened 2023-02-13
Use registries in fallback expander


voltrevo:
Requires:
- https://github.com/web3well/bls-wallet/issues/521
- https://github.com/web3well/bls-wallet/issues/522
- https://github.com/web3well/bls-wallet/issues/524

After [pseudo-floats](https://github.com/web3well/bls-wallet/pull/519), an ETH transfer is expected to be encoded as 184 bytes using the fallback expander:

| Field               | Format      | Bytes Used |
|---------------------|-------------|------------|
| expanderIndex       | [VLQ](https://github.com/web3well/bls-wallet/blob/a4aaa42/contracts/contracts/lib/VLQ.sol#L6-L24)         | 1          |
| BLS Public Key      | Raw         | 128        |
| Nonce               | [VLQ](https://github.com/web3well/bls-wallet/blob/a4aaa42/contracts/contracts/lib/VLQ.sol#L6-L24)         | 2 (ish)    |
| Gas                 | [PseudoFloat](https://github.com/web3well/bls-wallet/blob/74d31d5/contracts/contracts/lib/PseudoFloat.sol#L8-L70) | 2          |
| Action Count        | [VLQ](https://github.com/web3well/bls-wallet/blob/a4aaa42/contracts/contracts/lib/VLQ.sol#L6-L24)         | 1          |
| [0] ethValue        | [PseudoFloat](https://github.com/web3well/bls-wallet/blob/74d31d5/contracts/contracts/lib/PseudoFloat.sol#L8-L70) | 2 (ish)    |
| [0] contractAddress | Raw         | 20         |
| [0] dataLen         | [VLQ](https://github.com/web3well/bls-wallet/blob/a4aaa42/contracts/contracts/lib/VLQ.sol#L6-L24)         | 1          |
| [0] data            | Raw         | 0          |
| [1] ethValue        | [PseudoFloat](https://github.com/web3well/bls-wallet/blob/74d31d5/contracts/contracts/lib/PseudoFloat.sol#L8-L70) | 2          |
| [1] contractAddress | Raw         | 20         |
| [1] dataLen         | [VLQ](https://github.com/web3well/bls-wallet/blob/a4aaa42/contracts/contracts/lib/VLQ.sol#L6-L24)         | 1          |
| [1] data            | Raw         | 4          |
|                     |             |            |
| Total               |             | 184 (ish)  |

(Note that this includes an action to compensate tx.origin.)

[...]

PR #553, opened 2023-03-14
Registry wrapper classes

voltrevo:
# *Dependent PR*

This PR depends on:
- #546

The diff will include the changes from that PR until it is merged. Merging that PR first is recommended.

## What is this PR doing?

- Moves `SafeSingletonFactory` into `clients`
- Adds `BlsPublicKeyRegistryWrapper` and `AddressRegistryWrapper`

The wrapper classes have these methods for deployment:
- `deployNew` - deploy a new instance of the contract the old fashioned way
- `connectOrDeploy` - connect to *the* instance of the contract using `SafeSingletonFactory`, deploying if needed
- `connectIfDeployed` - connect to *the* instance of the contract using `SafeSingletonFactory`, if it has been deployed

Additionally, registries expose these methods:
- `lookup` - given an id, if it has associated data, return the associated data
- `reverseLookup` - given associated data, if it has an associated id, return the most recent associated id
- `register` - register some data, returning the allocated id
- `registerIfNeeded` - do a `reverseLookup` on some data, registering it if it's not already registered

`yarn hardhat test`
(See `addressRegistry-test.ts`, `blsPublicKeyRegistry-test.ts`)

Contributes to #523.

## Checklist

## Guidelines

### DIFF
--- a/contracts/clients/src/SafeSingletonFactory.ts
+++ b/contracts/clients/src/SafeSingletonFactory.ts
@@ -1,5 +1,6 @@
 import assert from "assert";
-import { ethers } from "ethers";
+import { ethers, Signer } from "ethers";
+import SignerOrProvider from "./helpers/SignerOrProvider";
 
 /**
  * Filters out the optional elements of an array type because an optional
@@ -63,11 +64,24 @@ export default class SafeSingletonFactory {
     },
   };
 
-  constructor(
+  provider: ethers.providers.Provider;
+
+  // eslint-disable-next-line no-use-before-define
+  viewer: SafeSingletonFactoryViewer;
+
+  private constructor(
     public signer: ethers.Signer,
-    public provider: ethers.providers.Provider,
+    public chainId: number,
     public address: string,
-  ) {}
+  ) {
+    if (signer.provider === undefined) {
+      throw new Error("Expected signer with provider");
+    }
+
+    this.provider = signer.provider;
+
+    this.viewer = new SafeSingletonFactoryViewer(signer, chainId);
+  }
 
   static async init(signer: ethers.Signer): Promise<SafeSingletonFactory> {
     assert(signer.provider !== undefined, "Expected signer with provider");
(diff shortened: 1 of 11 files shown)


contracts/clients/src/SafeSingletonFactory.ts
```diff
@@ -63,11 +64,24 @@ export default class SafeSingletonFactory {
...
   };
 
-  constructor(
+  provider: ethers.providers.Provider;
+
+  // eslint-disable-next-line no-use-before-define
+  viewer: SafeSingletonFactoryViewer;
+
+  private constructor(
     public signer: ethers.Signer,
```
jacque006, line 73: Why is a signer now required for this? Someone may want to use this in a read-only mode for safety, and it looks like `SafeSingletonFactoryViewer` can work with just a provider.

contracts/clients/src/index.ts
```diff
@@ -36,6 +36,14 @@ import { BlsWalletContracts, connectToContracts } from "./BlsWalletContracts";
 
 export * from "./signer";
 
+export {
+  default as SafeSingletonFactory,
+  SafeSingletonFactoryViewer,
+} from "./SafeSingletonFactory";
+
+export { default as AddressRegistryWrapper } from "./AddressRegistryWrapper";
+export { default as BlsPublicKeyRegistryWrapper } from "./BlsPublicKeyRegistryWrapper";
```
jacque006, line 45: Would be good to have tsdocs on these classes & their public functions & props

jacque006 (approved):

contracts/clients/src/SafeSingletonFactory.ts
voltrevo: I think the diff is just a bit confusing here - it was requiring a signer previously.

This PR actually introduces `SafeSingletonFactoryViewer` for this purpose, which allows you to create it with `signerOrProvider`.

contracts/clients/src/index.ts
voltrevo: https://github.com/web3well/bls-wallet/pull/556

PR #555, opened 2023-03-15
Add expander for registration

### LINKED ISSUE #525: BLSRegistration contract (including expander)
Requires:
- https://github.com/web3well/bls-wallet/issues/522

At some point, the full 128 bytes of the user's BLS public key need to be registered.

However, before registration, users need to send those 128 bytes to identify their account. Without a dedicated expander, they would then need to provide those 128 bytes *again* to encode the act of registering their BLS public key.

We should make it standard for users to register their address as well, so that other users can send them transfers efficiently. This doesn't cost any calldata because we can simply use `msg.sender`.

Here's an expected outline of the contract:

```solidity
contract BLSRegistration is IExpander {
  BLSPublicKeyRegistry blsPublicKeyRegistry;
  AddressRegistry addressRegistry;

  constructor() {
    // set registry addresses
  }

  function expand(bytes calldata stream) {
    // extract 128 bytes from stream as senderPublicKey
    // return senderPublicKey and an operation encoding a call to register with senderPublicKey
  }
  
  function register(uint256[4] blsPublicKey) {
    blsPublicKeyRegistry.register(blsPublicKey);
    addressRegistry.register(msg.sender);
  }
}
```

voltrevo:
## What is this PR doing?

- Adds `BLSRegistration.sol`
  - Both an expander and supplies the .register method so it can use `msg.sender` for the address
- Adds compressor `BlsRegistrationCompressor`
- Adds tests with detailed explanations of actual compressed data sent to `BLSExpanderDelegator.run`

`yarn hardhat test`

(Search 'registration' in `expander-test.ts` for relevant new tests.)

Resolves #525

## Checklist

## Guidelines


contracts/contracts/AddressRegistry.sol
```diff
@@ -1,4 +1,4 @@
-//SPDX-License-Identifier: MIT
```
jacque006, line 1: Why were these changed? Our repo and the code within is on the [MIT License](https://github.com/web3well/bls-wallet/blob/main/LICENSE) (not including our deps)

jacque006 (changes requested):
Only blocking issue I see is license change.

I personally prefer using Unlicense, but most of our groups projects use MIT I imagine for some specific legal reason.

contracts/contracts/AddressRegistry.sol
jacque006: Core contracts are Unlicense, repo is MIT. Will discuss with James. Fine as is for now.

jacque006 (approved):
License matches existing contracts, fine as is https://github.com/web3well/bls-wallet/pull/555#discussion_r1139469421

PR #561, opened 2023-03-24
201 implement erc 1271 signature validation

### LINKED ISSUE #201: Implement EIP-1271 (Signature Validation)
### Problem

Currently, it is difficult for a dApp to verify BLS signed messages done by our client library/extension. They must implement custom logic.

### Possible Solution

[EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) provides a common interface that allows a contract to validate a signature. We could implement it by:
- Add new validate signature function to `VerificationGateway.sol` that can verify a BLS signed message.
- Implement `ERC1271` spec on `BLSWallet.sol`, using the new signature validation function above.

This creates tight coupling between the wallet and gateway, which could be undesirable.

JohnGuilding:
## What is this PR doing?
This PR does two things:
1. Adds a new mapping to the verification gateway:
`mapping(bytes32 => uint256[BLS_KEY_LEN]) public BLSPublicKeyFromHash;`

2. Adds support for [ERC-1271](https://eips.ethereum.org/EIPS/eip-1271) signature validation

### Why do we need a new mapping?
After speaking with James, we decided to go with the solution Jake outlined in #201 of adding the 1271 spec (`isValidSignature` function) to the `BLSWallet` and have that call the `VerificationGateway` to run the actual validation. This creates tighter coupling between `BLSWallet` and the `VerificationGateway`, but this solution was deemed the most appropriate for the purpose of ERC-1271.

In order for that solution to work, the `VerificationGateway` needs to have reference to a wallets `PublicKey`, in order to run `blsLib.verifySingle` to verify the signature. The new mapping was added in order to meet this requirement. The lowest level function call that updates the hash mappings is `updateWalletHashMappings`, so that function was chosen to update the new `BLSPublicKeyFromHash` mapping. This additional change increases the complexity of the PR beyond the initial scope.

`yarn hardhat test`
- ERC-1271 tests have been added to `walletAction-test.ts`
- Existing tests in `recovery-test.ts` & `upgrade-test.ts` were updated to test the new mapping

resolves #201 

## Checklist

## Guidelines

### DIFF
--- a/contracts/contracts/BLSWallet.sol
+++ b/contracts/contracts/BLSWallet.sol
@@ -7,6 +7,9 @@ pragma abicoder v2;
 import "@openzeppelin/contracts/proxy/utils/Initializable.sol";
 import "./interfaces/IWallet.sol";
 
+interface IVerificationGateway {
+    function validateSignature(bytes32 _hash, bytes memory _signature) external view returns (bool);
+}
 
 /** Minimal upgradable smart contract wallet.
     Generic calls can only be requested by its trusted gateway.
@@ -191,6 +194,21 @@ contract BLSWallet is Initializable, IWallet
             }
 
             results[i] = result;
+        }
+    }
+
+    /**
+    * @dev ERC-1271 signature validation
+    */
+    function isValidSignature(
+        bytes32 _hash,
+        bytes memory _signature
+    ) public view returns (bytes4 magicValue) {
+
+        bool verified = IVerificationGateway(trustedBLSGateway).validateSignature(_hash, _signature);
+
+        if (verified) {
+            magicValue = 0x1626ba7e;
         }
     }
 
--- a/contracts/contracts/BLSWallet.sol
+++ b/contracts/contracts/BLSWallet.sol
@@ -208,6 +208,8 @@ contract BLSWallet is Initializable, IWallet
 
         if (verified) {
             magicValue = 0x1626ba7e;
+        } else {
+            magicValue = 0xffffffff;
         }
     }
 
--- a/contracts/contracts/BLSWallet.sol
+++ b/contracts/contracts/BLSWallet.sol
@@ -201,11 +201,11 @@ contract BLSWallet is Initializable, IWallet
     * @dev ERC-1271 signature validation
     */
     function isValidSignature(
-        bytes32 _hash,
-        bytes memory _signature
+        bytes32 hash,
+        bytes memory signature
     ) public view returns (bytes4 magicValue) {
 
-        bool verified = IVerificationGateway(trustedBLSGateway).validateSignature(_hash, _signature);
+        bool verified = IVerificationGateway(trustedBLSGateway).validateSignature(hash, signature);
 
         if (verified) {
             magicValue = 0x1626ba7e;
--- a/contracts/contracts/BLSWallet.sol
+++ b/contracts/contracts/BLSWallet.sol
@@ -8,7 +8,7 @@ import "@openzeppelin/contracts/proxy/utils/Initializable.sol";
 import "./interfaces/IWallet.sol";
 
 interface IVerificationGateway {
-    function validateSignature(bytes32 _hash, bytes memory _signature) external view returns (bool);
+    function isValidSignature(bytes32 _hash, bytes memory _signature) external view returns (bool);
 }
 
 /** Minimal upgradable smart contract wallet.
@@ -204,8 +204,7 @@ contract BLSWallet is Initializable, IWallet
         bytes32 hash,
         bytes memory signature
     ) public view returns (bytes4 magicValue) {
-
-        bool verified = IVerificationGateway(trustedBLSGateway).validateSignature(hash, signature);
+        bool verified = IVerificationGateway(trustedBLSGateway).isValidSignature(hash, signature);
 
         if (verified) {
             magicValue = 0x1626ba7e;
--- a/contracts/contracts/VerificationGateway.sol
+++ b/contracts/contracts/VerificationGateway.sol
@@ -31,8 +31,9 @@ contract VerificationGateway
     BLSWallet public immutable blsWalletLogic = new BLSWallet();
     mapping(bytes32 => IWallet) public walletFromHash;
     mapping(IWallet => bytes32) public hashFromWallet;
+    mapping(bytes32 => uint256[BLS_KEY_LEN]) public BLSPublicKeyFromHash;
 
-    //mapping from an existing wallet's bls key hash to pending variables when setting a new BLS key
+    // mapping from an existing wallet's bls key hash to pending variables when setting a new BLS key
     mapping(bytes32 => uint256[BLS_KEY_LEN]) public pendingBLSPublicKeyFromHash;
     mapping(bytes32 => uint256[2]) public pendingMessageSenderSignatureFromHash;
     mapping(bytes32 => uint256) public pendingBLSPublicKeyTimeFromHash;
@@ -323,7 +324,7 @@ contract VerificationGateway
                 address(walletProxyAdmin),
                 getInitializeData()
             )));
-            updateWalletHashMappings(publicKeyHash, blsWallet);
+            updateWalletHashMappings(publicKeyHash, blsWallet, publicKey);
             emit WalletCreated(
                 address(blsWallet),
                 publicKey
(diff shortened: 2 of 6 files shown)


contracts/contracts/BLSWallet.sol
```diff
@@ -7,6 +7,9 @@ pragma abicoder v2;
 import "@openzeppelin/contracts/proxy/utils/Initializable.sol";
 import "./interfaces/IWallet.sol";
 
+interface IVerificationGateway {
+    function validateSignature(bytes32 _hash, bytes memory _signature) external view returns (bool);
+}
```
JohnGuilding, line 12: Chose to add an interface over a low level call for compile-time safety. Happy to change if this is undesirable

contracts/test/walletAction-test.ts
```diff
@@ -288,6 +288,123 @@ describe("WalletActions", async function () {
...
+      ethers.utils.hexlify(signature[0]) +
+      ethers.utils.hexlify(signature[1]).substring(2);
+    const signatureBytes = ethers.utils.arrayify(concatenatedSignature);
+
+    const expectedBytes4MagicValue = ethers.utils
+      .solidityKeccak256(["string"], ["isValidSignature(bytes32,bytes)"])
+      .substring(0, 10);
+
+    // initial call needed to ensure isValidSignature is successful
+    await fx.call(wallet, blsWallet, "nonce", [], 0, 30_000_000);
```
JohnGuilding, line 310: Had to run this call in order to get the tests working, otherwise an exception was thrown on `blsWallet.isValidSignature`. Wasn't entirely sure on the underlying reason it works. Wallet initialisation?

contracts/contracts/VerificationGateway.sol
```diff
@@ -112,6 +113,33 @@ contract VerificationGateway
         require(verified, "VG: Sig not verified");
     }
 
+    function validateSignature(
+        bytes32 hash,
+        bytes memory signature
+    ) public view returns (bool verified) {
+        IWallet wallet = IWallet(msg.sender);
+        bytes32 existingHash = hashFromWallet[wallet];
+        require(
```
jzaki: Can use `onlyWallet` modifier instead.

contracts/contracts/VerificationGateway.sol
```diff
@@ -112,6 +113,33 @@ contract VerificationGateway
...
+            (IWallet(msg.sender) == walletFromHash[existingHash]),
+            "VG: not called from wallet"
+        );
+
+        uint256[BLS_KEY_LEN] memory publicKey = BLSPublicKeyFromHash[existingHash];
+
+        bytes memory concatenatedHash = bytes.concat(hash);
```
jzaki: Using `concat` to convert to bytes? Consider `abi.encode`?

contracts/contracts/VerificationGateway.sol
```diff
@@ -112,6 +113,33 @@ contract VerificationGateway
         require(verified, "VG: Sig not verified");
     }
 
+    function validateSignature(
```
jzaki: A point around naming. The result (_is valid_ or not) makes sense as the state. The process used with the signature is _verification_ (`verifySingle`).
Whereas "validate" is to make valid (eg validate a parking ticket at a store). Can this also be "isValidSignature"?

contracts/contracts/VerificationGateway.sol
```diff
@@ -112,6 +113,33 @@ contract VerificationGateway
...
+        uint256[2] memory message = blsLib.hashToPoint(
+            BLS_DOMAIN,
+            concatenatedHash
+        );
+
+        require(signature.length == 64, "VG: Sig bytes length must be 64");
+        uint256[2] memory decodedSignature = abi.decode(signature, (uint256[2]));
+
+        verified = blsLib.verifySingle(decodedSignature, publicKey, message);
+        require(verified, "VG: Sig not verified");
```
jzaki: The 1271 reference implementation returns `0xffffffff`, which I guess could be any other value. Is throwing preferred?

jzaki (approved):

contracts/contracts/VerificationGateway.sol
JohnGuilding: I did consider `abi.encode`, but it seems like the code only needs to convert the `bytes32` to `bytes`. In which case I believe `bytes.concat` was added specifically for this use-case:

> introduce concat which takes a variable number of bytes memory as an input and returns a single bytes memory

from https://github.com/ethereum/solidity/issues/10903

So would `bytes.concat` make more sense in this case? Or would `abi.encode` make more sense?

jzaki: To me, encoding a single `bytes32` to `bytes` makes more sense with [`abi.encode`](https://docs.soliditylang.org/en/latest/cheatsheet.html#global-variables), rather than [concatenating](https://docs.soliditylang.org/en/latest/types.html#the-functions-bytes-concat-and-string-concat) itself with nothing else to get the `bytes` type.

JohnGuilding: Yeah I think reading deeper into it, I agree with you

Looking at the [release notes](https://blog.soliditylang.org/2021/04/21/solidity-0.8.4-release-announcement/), looks like it was mainly designed to concatenate values instead of convert them. Will swap to use abi.encode

contracts/contracts/VerificationGateway.sol
JohnGuilding: Yeah like that and that makes sense, will change

contracts/contracts/VerificationGateway.sol
JohnGuilding: Yeah noticed that. My original thought was to use a similar require statement to the verify function in the same file. So my original intent was consistency. 

From different implementations:
[0xsequence - returns only happy path magic value (0x1626ba7e), but doesn't throw if signature is incorrect](https://github.com/0xsequence/wallet-contracts/blob/50204fd69535207cfbf5894c9ce4c100a3f82b91/src/contracts/modules/commons/ModuleAuth.sol#L147)
[ambire - uses 0xffffffff](https://github.com/AmbireTech/wallet/blob/f6c952c8b7ad0c0a5d48d7ba22ab265d8a5429d0/contracts/QuickAccManager.sol#L124)
[etherspot - uses 0xffffffff](https://github.com/etherspot/etherspot-contracts/blob/3d07a849a00832d4cad7ac6d56c9ab9d31b2e772/src/common/account/AccountImplementationV1.sol#L20)
[biconomy - uses 0xffffffff](https://github.com/bcnmy/scw-contracts/blob/ac19b1569dca05f76eec233f7b686776141f231f/contracts/smart-contract-wallet/SmartAccount.sol#L798)

So maybe throwing is not preferred and returning 0xffffffff would be more consistent with other implementations?

JohnGuilding: changed to `0xffffffff`, to match erc reference implementation, and implementations from other projects

contracts/contracts/VerificationGateway.sol
```diff
@@ -112,6 +113,28 @@ contract VerificationGateway
...
+    function isValidSignature(
+        bytes32 hash,
+        bytes memory signature
+    ) public view onlyWallet(hashFromWallet[IWallet(msg.sender)]) returns (bool verified) {
+        IWallet wallet = IWallet(msg.sender);
+        bytes32 existingHash = hashFromWallet[wallet];
+
+        uint256[BLS_KEY_LEN] memory publicKey = BLSPublicKeyFromHash[existingHash];
+
+        bytes memory concatenatedHash = abi.encode(hash);
```
jzaki: Minor point, name can probably be updated now. eg hashBytes

contracts/contracts/VerificationGateway.sol
```diff
@@ -112,6 +113,28 @@ contract VerificationGateway
         require(verified, "VG: Sig not verified");
     }
 
+    function isValidSignature(
+        bytes32 hash,
+        bytes memory signature
+    ) public view onlyWallet(hashFromWallet[IWallet(msg.sender)]) returns (bool verified) {
```
jzaki: See the logic inside "onlyWallet", it verifies the sender matches the given hash. So you want to just pass in the `hash` parameter.

jzaki (approved):

contracts/contracts/VerificationGateway.sol
jzaki: Err, I mistook the hash here.

contracts/contracts/VerificationGateway.sol
JohnGuilding: As discussed async, arrived at the conclusion that the following code achieves the desired result in the best way:
```solidity
        IWallet wallet = IWallet(msg.sender);
        bytes32 existingHash = hashFromWallet[wallet];
        require(existingHash != 0, "VG: not called from wallet");
```

contracts/contracts/VerificationGateway.sol
JohnGuilding: Good point, changing

PR #562, opened 2023-03-27
549 add documentation for bls provider and bls signer

### LINKED ISSUE #549: Add documentation for BlsProvider & BlsSigner
Add documentation for BlsProvider and BlsSigner. This includes adding to READMEs, and some js doc comments on custom methods

JohnGuilding:
## What is this PR doing?
- Adds documentation for BlsProvider and BlsSigner

- Read documentation

#549 

## Checklist

## Guidelines

### DIFF
--- a/aggregator/src/app/BundleTable.ts
+++ b/aggregator/src/app/BundleTable.ts
@@ -60,14 +60,14 @@ export type BundleRow = Row;
 function fromRawRow(rawRow: RawRow | sqlite.Row): Row {
   if (Array.isArray(rawRow)) {
     rawRow = {
-      id: rawRow[0],
-      status: rawRow[1],
-      hash: rawRow[2],
-      bundle: rawRow[3],
-      eligibleAfter: rawRow[4],
-      nextEligibilityDelay: rawRow[5],
-      submitError: rawRow[6],
-      receipt: rawRow[7],
+      id: rawRow[0] as number,
+      status: rawRow[1] as string,
+      hash: rawRow[2] as string,
+      bundle: rawRow[3] as string,
+      eligibleAfter: rawRow[4] as string,
+      nextEligibilityDelay: rawRow[5] as string,
+      submitError: rawRow[6] as string | null,
+      receipt: rawRow[7] as string | null,
     };
   }
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -3,7 +3,7 @@
 - [See an overview of BLS Wallet & how the components work together](./system_overview.md)
 - [Use BLS Wallet in a browser/NodeJS/Deno app](./use_bls_wallet_clients.md)
 - [Use BLS Wallet in your L2 dApp for cheaper, multi action transactions](./use_bls_wallet_dapp.md)
-- [Use BLS Wallet components and features with an Ethers-like provider](./use_bls_provider.md)
+- [Use BLS Wallet components and features with an ethers.js provider and signer](./use_bls_provider.md)
 - Setup the BLS Wallet components for:
   - [Local development](./local_development.md)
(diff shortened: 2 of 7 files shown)


docs/README.md
```diff
@@ -3,6 +3,7 @@
 - [See an overview of BLS Wallet & how the components work together](./system_overview.md)
 - [Use BLS Wallet in a browser/NodeJS/Deno app](./use_bls_wallet_clients.md)
 - [Use BLS Wallet in your L2 dApp for cheaper, multi action transactions](./use_bls_wallet_dapp.md)
+- [Use BLS Wallet components and features with an Ethers-like provider](./use_bls_provider.md)
```
jacque006:
```suggestion
- [Use BLS Wallet components and features with an ethers.js provider & signer](./use_bls_provider.md)
```

jacque006: Would also be good to add this entry to top level `README`

docs/use_bls_provider.md
```diff
@@ -0,0 +1,226 @@
...
+```
+
+### BlsSigner
+
+**Important:** Ensure that the BLS wallet you are linking the `BlsSigner` to via the private key is funded. Alternatively, if a wallet doesn't yet exist, it will be lazily created on the first transaction. In this scenario, you can create a random BLS private key with the following helper method and fund that account. It will need to be funded in order to send its first transaction.
+
+```ts
+import { BlsWalletWrapper } from "bls-wallet-clients";
+
+const privateKey = await BlsWalletWrapper.getRandomBlsPrivateKey();
```
jacque006, line 45: We should expose this as a static method on `BlsProvider/Signer` so in some use cases users can avoid the wrapper import. Future follow up, out of scope.

docs/use_bls_provider.md
```diff
@@ -0,0 +1,226 @@
...
+
+// Send funds to this address if the wallet does not exist
+const address = await BlsWalletWrapper.Address(
+    privateKey,
+    verificationGatewayAddress,
+    provider
+);
```
jacque006:
```suggestion
const signer = provider.getSigner(privateKey);
const address = await signer.getAddress();
```

jacque006 (approved):
In general, when documenting/using the provider and signer we should avoid the need for `BlsWalletWrapper` & other classes/functions to reduce complexity of examples. People can use existing guides if they need to go lower level.

docs/README.md
JohnGuilding: Good shout, missed that

docs/use_bls_provider.md
JohnGuilding: Makes sense, will create a follow up issue

JohnGuilding: https://github.com/web3well/bls-wallet/issues/565

aggregator/src/app/BundleTable.ts
```diff
@@ -60,14 +60,14 @@ export type BundleRow = Row;
...
-      submitError: rawRow[6],
-      receipt: rawRow[7],
+      id: rawRow[0] as number,
+      status: rawRow[1] as string,
+      hash: rawRow[2] as string,
+      bundle: rawRow[3] as string,
+      eligibleAfter: rawRow[4] as string,
+      nextEligibilityDelay: rawRow[5] as string,
+      submitError: rawRow[6] as string | null,
+      receipt: rawRow[7] as string | null,
```
JohnGuilding, line 70: Was getting some typescript errors when running ./programs/checkTs.ts and the aggregator tests. See [workflow attempts](https://github.com/web3well/bls-wallet/actions/runs/4541834483).

Was able to resolve by making this change

Issue #459, opened 2023-01-16
Audit fix 07


jzaki:
PR - https://github.com/web3well/bls-wallet/pull/571

PR #566, opened 2023-03-28
Audit issue 7

blakecduncan:
## What is this PR doing?

Issues:
- `BLS_DOMAIN` is just an arbitrary string so it doesn't protect against replay across contracts or chains.
- Ex: The message in `safeSetWallet()` is only signed with the wallet address.  
- Ex: In the `verify()` function, the `chainId` is appended to messages but not the verification gateway address.

Solution:
- Include the chainId and the verification gateway address in the domain.
- Create a seperate domain for `safeSetWallet` and `verify` so that it's not possible for overlap between a bundle message and a proof of possession message.

- Running all tests and checking Quil/example dapps still work

https://github.com/web3well/bls-wallet/issues/459

## Checklist

## Guidelines

### DIFF
--- a/contracts/clients/src/signer/encodeMessageForSigning.ts
+++ b/contracts/clients/src/signer/encodeMessageForSigning.ts
@@ -1,7 +1,7 @@
 import { keccak256, solidityPack } from "ethers/lib/utils";
 import { Operation } from "./types";
 
-export default (chainId: number) =>
+export default () =>
   (operation: Operation, walletAddress: string): string => {
     let encodedActionData = "0x";
 
@@ -18,7 +18,7 @@ export default (chainId: number) =>
     }
 
     return solidityPack(
-      ["uint256", "address", "uint256", "bytes32"],
-      [chainId, walletAddress, operation.nonce, keccak256(encodedActionData)],
+      ["address", "uint256", "bytes32"],
+      [walletAddress, operation.nonce, keccak256(encodedActionData)],
     );
   };
--- a/contracts/clients/src/signer/index.ts
+++ b/contracts/clients/src/signer/index.ts
@@ -18,9 +18,9 @@ export type BlsWalletSigner = AsyncReturnType<typeof initBlsWalletSigner>;
 export async function initBlsWalletSigner({
   chainId,
   privateKey,
-  verificationGateway,
+  verificationGatewayAddress,
 }: {
-  verificationGateway: string;
+  verificationGatewayAddress: string;
   chainId: number;
   privateKey: string;
 }) {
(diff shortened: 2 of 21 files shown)


contracts/clients/src/signer/encodeMessageForSigning.ts
```diff
@@ -18,7 +18,7 @@ export default (chainId: number) =>
     }
 
     return solidityPack(
-      ["uint256", "address", "uint256", "bytes32"],
-      [chainId, walletAddress, operation.nonce, keccak256(encodedActionData)],
+      ["address", "uint256", "bytes32"],
+      [walletAddress, operation.nonce, keccak256(encodedActionData)],
```
blakecduncan, line 22: Removing chainId because that is now part of the domain

contracts/clients/src/signer/index.ts
```diff
@@ -32,14 +32,17 @@ export async function initBlsWalletSigner({
...
-    getPublicKeyHash: getPublicKeyHash(signerFactory, domain, privateKey),
-    getPublicKeyStr: getPublicKeyStr(signerFactory, domain, privateKey),
-    sign: sign(signerFactory, domain, chainId, privateKey),
-    signMessage: signMessage(signerFactory, domain, privateKey),
-    verify: verify(domain, chainId),
+    getPublicKey: getPublicKey(signerFactory, bundleDomain, privateKey),
+    getPublicKeyHash: getPublicKeyHash(signerFactory, bundleDomain, privateKey),
+    getPublicKeyStr: getPublicKeyStr(signerFactory, bundleDomain, privateKey),
+    sign: sign(signerFactory, bundleDomain, chainId, privateKey),
+    signMessage: signMessage(signerFactory, walletDomain, privateKey),
```
blakecduncan, line 44: I think this is the only function that needs the wallet domain (proof of possession).  If I'm missing one definitely let me know though.

contracts/contracts/VerificationGateway.sol
```diff
@@ -73,6 +75,16 @@ contract VerificationGateway
         blsLib = bls;
         blsWalletLogic = blsWalletImpl;
         walletProxyAdmin = ProxyAdmin(proxyAdmin);
+        WALLET_DOMAIN = keccak256(abi.encodePacked(
+            block.chainid,
+            address(this),
+            "Wallet"
```
blakecduncan, line 81: Differentiating between the domains by adding a string (wallet/bundle)

contracts/test/upgrade-test.ts
```diff
@@ -121,14 +121,14 @@ describe("Upgrade", async function () {
...
-    );
     // Sign simple address message
     const addressMessage = solidityPack(["address"], [walletAddress]);
-    const addressSignature = wallet.signMessage(addressMessage);
+    const walletNewVg = await BlsWalletWrapper.connect(
+      blsSecret,
+      vg2.address,
+      vg2.provider,
+    );
+    const addressSignature = walletNewVg.signMessage(addressMessage);
```
blakecduncan, line 131: Now that the message needs to be signed with the VG address, I needed to sign this message with a wallet for the new VG rather than the original wallet

PR #570, opened 2023-03-29
567 bls provider and bls signer release

### LINKED ISSUE #567: Formally release BlsProvider and BlsSigner
This issue is concerned with formally releasing the BlsProvider and BlsSigner in the client module

**Steps**:
1. Remove from experimental namespace
2. Experimental release
3. Formal release

JohnGuilding:
## What is this PR doing?
- Removes BlsProvider & BlsSigner from Experimental namespace
- Publishes new client module release - v0.8.3

I've manually tested the new release against testnet and a local aggregator. I did this by adding the provider to the browser wallet and sending a tx to and from the wallet.

- test provider on testnet with browser wallet (I've already done this)

Resolves #567 

## Checklist

## Guidelines

### DIFF
--- a/aggregator-proxy/package.json
+++ b/aggregator-proxy/package.json
@@ -21,7 +21,7 @@
     "@types/koa__cors": "^3.3.0",
     "@types/koa__router": "^8.0.11",
     "@types/node-fetch": "^2.6.1",
-    "bls-wallet-clients": "0.8.2-1452ef5",
+    "bls-wallet-clients": "0.8.2-de12b3c",
     "fp-ts": "^2.12.1",
     "io-ts": "^2.2.16",
     "io-ts-reporters": "^2.0.1",
--- a/aggregator-proxy/package.json
+++ b/aggregator-proxy/package.json
@@ -21,7 +21,7 @@
     "@types/koa__cors": "^3.3.0",
     "@types/koa__router": "^8.0.11",
     "@types/node-fetch": "^2.6.1",
-    "bls-wallet-clients": "0.8.2-de12b3c",
+    "bls-wallet-clients": "0.8.3",
     "fp-ts": "^2.12.1",
     "io-ts": "^2.2.16",
     "io-ts-reporters": "^2.0.1",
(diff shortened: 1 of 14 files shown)


JohnGuilding:
### Note on testing:
I was not able to get transactions to work out of the box with the deployed aggregator OR a local aggregator. This was the case with just using the client module as normal, or the provider, so it wasn't an issue specific to the provider changes. Testing the client module by itself required new logic in the browser wallet to pay the aggregator (I haven't committed these changes but we should make this change at some point)

I can't say with confidence what the issue was with the deployed aggregator, as I'm not sure how recently that was updated. But the issue with the local aggregator was that the code failed in the `AggregationStrategy.ts` file. [This if statement](https://github.com/web3well/bls-wallet/blob/5529e078d9e190a52a636f8fef593905e017a086/aggregator/src/app/AggregationStrategy.ts#L417) evaluated to true, and thus the bundle row never got added on [this line](https://github.com/web3well/bls-wallet/blob/5529e078d9e190a52a636f8fef593905e017a086/aggregator/src/app/AggregationStrategy.ts#L424).

I was able to get around the issue by manually updating this env var like so:
```bash
# old value - this value is provided in .env.example
MAX_GAS_PER_BUNDLE=2000000

# new value
MAX_GAS_PER_BUNDLE=8000000
```

Made the decision to publish these changes regardless as appears to be an issue with aggregator logic/aggregator config, as opposed to being related to an issue with the client module. If it is worth another issue to look into, then would deem it out of scope. Feel free to disagree!

**Edit:**

Some example values I was getting when checking why the bundle rows weren't getting added. These values correspond to values observed for [`newAggregatedGas`](https://github.com/web3well/bls-wallet/blob/5529e078d9e190a52a636f8fef593905e017a086/aggregator/src/app/AggregationStrategy.ts#L413-L415):
- 15838902 (notice the extra digit, not sure why this was so high. Makes me think I copied it wrong lol)
- 4999229
- 5088000

Whereas this value should be under (according to the default config):
- 2000000

blakecduncan (approved):
LGTM!

That's a good shout, we should update the browser wallet to pay the aggregator.

I'm not sure about the `MAX_GAS_PER_BUNDLE` part though.  I'll defer to @voltrevo maybe for that one?  But I agree that shouldn't block this PR.

voltrevo:
@JohnGuilding Hmm. `MAX_GAS_PER_BUNDLE` should be higher anyway, so all good with 8m.

*However*, that's A LOT of gas... could you provide some more detail on how you tested?

> test provider on testnet with browser wallet (I've already done this)

I would have tested with an ETH transfer using Quill, but I can't tell from your description whether this is what you did.

Could you also confirm which network you used? testnet doesn't mean very much, because we've only ever used testnets so far. ie, do you mean geth-dev? hardhat? arbitrum-goerli?

JohnGuilding:
@voltrevo yeah 8m does seem quite high. 

> could you provide some more detail on how you tested?

- **Testing method:** Sending transactions using the [browser wallet](https://github.com/web3well/bls-browser-wallet) pointing at a local aggregator (didn't test using quill)
- **Testnet**: arbitrum-goerli

PR #577, opened 2023-04-03
ERC20 expander

### LINKED ISSUE #576: Add expander for ERC20 operations

voltrevo:
## What is this PR doing?

(Also includes fixes for bls-wallet-clients... guessing there was a bad merge involved previously.)

Adds ERC20 expander that encodes transfers in about 20 bytes plus:
- 1 byte to indicate the choice of expander
- shared signature bytes

Supports multiple actions and the other ERC20 methods (approve, transferFrom, mint).

Example:

```
0f      - 0x0f = 0b1111 bit stream:
          - 1: Use registry for BLS key
          - 1: Include a tx.origin payment
          - 1: Use registry for ERC20 address
          - 1: Use registry for recipient address
000000  - Registry index for sendWallet's public key
00      - nonce: 0
0bda28  - gas: 92,483
02      - two actions
000000  - Registry index for ERC20 address
00      - transfer
000002  - Registry index for recipient address
9100    - 0.1 MCK (amount/value)
6d00    - Pay 0.000005 ETH to tx.origin
```

Improvements that could be made for super-optimized transfers:
- Optimize specifically for transfer
- Require registries, removing the need for a bit stream
- Use a fixed gas value
- Use a dedicated ERC20 registry, allowing a single byte for popular currencies (and 2 bytes for all but the most obscure currencies)
- Use a fixed tx.origin reward (or call contract that computes an appropriate reward)

This would get us down to about 11 bytes.

`yarn test`

Resolves #576

## Checklist

## Guidelines

### DIFF
--- a/contracts/clients/src/Erc20Compressor.ts
+++ b/contracts/clients/src/Erc20Compressor.ts
@@ -221,8 +221,6 @@ export default class Erc20Compressor implements IOperationCompressor {
         result.push(encodeRegIndex(addressId));
       }
 
-      // TODO: Be different here
-
       const success = await this.compressFunctionCall(
         action.encodedFunction,
         result,
--- a/contracts/clients/src/Erc20Compressor.ts
+++ b/contracts/clients/src/Erc20Compressor.ts
@@ -1,0 +1,392 @@
+import { BigNumber, ethers, Signer } from "ethers";
+import {
+  AddressRegistry__factory as AddressRegistryFactory,
+  BLSPublicKeyRegistry__factory as BLSPublicKeyRegistryFactory,
+  ERC20Expander,
+  ERC20Expander__factory as ERC20ExpanderFactory,
+  AggregatorUtilities,
+  AggregatorUtilities__factory as AggregatorUtilitiesFactory,
+} from "../typechain-types";
+import AddressRegistryWrapper from "./AddressRegistryWrapper";
+import BlsPublicKeyRegistryWrapper from "./BlsPublicKeyRegistryWrapper";
+import {
+  encodeBitStream,
+  encodePseudoFloat,
+  encodeRegIndex,
+  encodeVLQ,
+  hexJoin,
+} from "./encodeUtils";
+import SignerOrProvider from "./helpers/SignerOrProvider";
+import IOperationCompressor from "./IOperationCompressor";
+import SafeSingletonFactory, {
+  SafeSingletonFactoryViewer,
+} from "./SafeSingletonFactory";
+import { ActionData, Operation, PublicKey } from "./signer/types";
+
+export default class Erc20Compressor implements IOperationCompressor {
+  private constructor(
+    public erc20Expander: ERC20Expander,
+    public blsPublicKeyRegistry: BlsPublicKeyRegistryWrapper,
+    public addressRegistry: AddressRegistryWrapper,
+    public aggregatorUtilities: AggregatorUtilities,
+  ) {}
+
+  static async wrap(erc20Expander: ERC20Expander): Promise<Erc20Compressor> {
+    const [
+      blsPublicKeyRegistryAddress,
+      addressRegistryAddress,
+      aggregatorUtilitiesAddress,
+    ] = await Promise.all([
+      erc20Expander.blsPublicKeyRegistry(),
+      erc20Expander.addressRegistry(),
+      erc20Expander.aggregatorUtilities(),
+    ]);
+
+    return new Erc20Compressor(
+      erc20Expander,
+      new BlsPublicKeyRegistryWrapper(
+        BLSPublicKeyRegistryFactory.connect(
+          blsPublicKeyRegistryAddress,
+          erc20Expander.signer,
+        ),
+      ),
+      new AddressRegistryWrapper(
+        AddressRegistryFactory.connect(
+          addressRegistryAddress,
+          erc20Expander.signer,
+        ),
+      ),
+      AggregatorUtilitiesFactory.connect(
+        aggregatorUtilitiesAddress,
+        erc20Expander.signer,
+      ),
+    );
+  }
+
+  static async deployNew(signer: Signer): Promise<Erc20Compressor> {
+    const blsPublicKeyRegistryFactory = new BLSPublicKeyRegistryFactory(signer);
+    const addressRegistryFactory = new AddressRegistryFactory(signer);
+    const aggregatorUtilitiesFactory = new AggregatorUtilitiesFactory(signer);
+
+    const [
+      blsPublicKeyRegistryContract,
+      addressRegistryContract,
+      aggregatorUtilitiesContract,
+    ] = await Promise.all([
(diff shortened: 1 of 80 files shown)


contracts/clients/src/Erc20Compressor.ts
```diff
@@ -0,0 +1,390 @@
...
+      return undefined;
+    }
+
+    return await Erc20Compressor.wrap(erc20Expander);
+  }
+
+  async compress(blsPublicKey: PublicKey, operation: Operation) {
+    const result: string[] = [];
+
+    const resultIndexForRegUsageBitStream = result.length;
```
blakecduncan, line 166: Is there any situation where this isn't `0`?  Just curious if there is a reason to initialize this with `result.length` when you just created an empty array.

contracts/clients/src/Erc20Compressor.ts
```diff
@@ -0,0 +1,390 @@
...
+    );
+
+    if (!erc20Expander) {
+      return undefined;
+    }
+
+    return await Erc20Compressor.wrap(erc20Expander);
+  }
+
+  async compress(blsPublicKey: PublicKey, operation: Operation) {
```
blakecduncan, line 163: Should we add a return type here?

contracts/test/expanders-test.ts
```diff
@@ -507,4 +508,168 @@ describe("Expanders", async function () {
...
+
+      01      - One operation
+      02      - Use expander 2 (ERC20 expander)
+      0f      - 0x0f = 0b1111 bit stream:
+                - 1: Use registry for BLS key
+                - 1: Include a tx.origin payment
+                - 1: Use registry for ERC20 address
+                - 1: Use registry for recipient address
+      000000  - Registry index for sendWallet's public key
+      00      - nonce: 0
```
blakecduncan, line 575: If the nonce is one byte, does that mean the highest nonce we will support for compression is `255`?

contracts/test/expanders-test.ts
```diff
@@ -507,4 +508,168 @@ describe("Expanders", async function () {
...
+      0bda28  - gas: 92,483
```
blakecduncan, line 576: Sorry I might be misunderstanding this but is this a hex to base 10 conversion? I don't understand how we are getting 92,483 here.  Since  `0bda28 base16` = `704341 base 10`

contracts/test/expanders-test.ts
```diff
@@ -507,4 +508,168 @@ describe("Expanders", async function () {
...
+                - 1: Use registry for ERC20 address
+                - 1: Use registry for spender address
+                (The read order is 'backwards' because the lowest bit is read
+                first)
+      000000  - Registry index for sendWallet's public key
+      00      - nonce: 0
+      0bb917  - gas: 58,555
+      01      - one action
+      000000  - Registry index for ERC20 address
+      03      - approve (max)
```
blakecduncan, line 655: Should we add a test for `approve`, `transferFrom` and `mint`?

contracts/contracts/ERC20Expander.sol
```diff
@@ -0,0 +1,316 @@
...
+    ) {
+        blsPublicKeyRegistry = blsPublicKeyRegistryParam;
+        addressRegistry = addressRegistryParam;
+        aggregatorUtilities = aggregatorUtilitiesParam;
+    }
+
+    function expand(bytes calldata stream) external view returns (
+        uint256[4] memory senderPublicKey,
+        IWallet.Operation memory operation,
+        uint256 bytesRead
```
blakecduncan, line 63: We have this return value in the function signature but it doesn't look like we return anything below.  Is that intentional?

blakecduncan (approved):
Looks good to me!

Just had a couple small questions for my clarity.  Also, a nice to have might be to add test coverage for the other 3 erc20 functions in the expander/compressor.

contracts/clients/src/Erc20Compressor.ts
voltrevo: Ah true, it is definitely zero. However, the idea is that this version is more compatible with future changes - if there was a different field in there then using `.length` will still get the right index. It also means that you can copy this pattern elsewhere and it'll work with less changes.

Having said that, there's an argument to be made that putting zero here is still simpler + preferable. I'm actually unsure. If you still feel `0` is better after reading this then let's go ahead and do that :).

contracts/clients/src/Erc20Compressor.ts
voltrevo: My opinion is definitely not.

In general, I think it is much better to mostly only use type definitions where they are needed. If the compiler infers the type correctly, then you should generally leave it alone.

This makes writing and modifying code much easier as the code is more fluid. Types often change but the usage remains consistent, and the compiler just checks that it flows through correctly. In the alternative where extra type annotations are added, you need to update those annotations too, which is extra friction.

For the readability side, you can get the type from hovering - I don't think there is a significant benefit to the reader. It can even be better for the reader since that type only shows up if the reader wants to see it, otherwise they don't need to be distracted by it.

Here's some more information on this point of view: https://www.youtube.com/watch?v=RmGHnYUqQ4k.

Definitely aware there are other perspectives here and happy to talk about it further.

contracts/contracts/ERC20Expander.sol
voltrevo: > it doesn't look like we return anything below

Ah but it does return stuff :).

This is a solidity feature that doesn't exist in many other languages - when you give names to those return types, the function returns those variables implicitly with no need for a `return` statement.

contracts/test/expanders-test.ts
voltrevo: Nah, we're using VLQ for the nonce, which stands for Variable Length Quantity. This means the encoding expands as needed to accommodate large values - this will work for the full `uint256` range (and theoretically beyond!).

Examples:

```
 * 0x00: 0
 * 0x01: 1
 * 0x7f: 127
 * 0x8100: 128
 * 0x8105: 133 = 1 * 128 + 5
 * 0x838005: 49,157 = 3 * 128^2 + 0 * 128^1 + 5 * 128^0
```

More information: https://github.com/web3well/bls-wallet/blob/8c91d2f/contracts/contracts/lib/VLQ.sol

contracts/test/expanders-test.ts
voltrevo: This one is encoded using `PseudoFloat`, which I created for us :). It's hard to read it directly because it does packing at the bit level. On the plus side though, it encodes anything in the uint256 range and generally delivers the value from just 2 bytes as long as you stick to 3 digits of precision.

Here's where we do that: https://github.com/web3well/bls-wallet/blob/8c91d2f/contracts/contracts/lib/PseudoFloat.sol

contracts/test/expanders-test.ts
voltrevo: Probably :)

voltrevo: https://github.com/web3well/bls-wallet/issues/593

PR #579, opened 2023-04-12
461 audit fix 9

### LINKED ISSUE #461: Audit fix 09

JohnGuilding:
## What is this PR doing?
Implements audit fix 9

- run contract tests

contributes to #461 

## Checklist

## Guidelines

### DIFF
--- a/contracts/test/recovery-test.ts
+++ b/contracts/test/recovery-test.ts
@@ -653,6 +653,9 @@ describe("Recovery", async function () {
       wallet2.blsWalletSigner.privateKey,
     );
 
+    // this call will succeed even with invalidSignatureExpiryTimestamp, as the
+    // wallet already has a bls key registered with the verification gateway.
+    // This means setSafeWallet is not called, and we don't make the signature expiry check
     await fx.call(
       wallet1,
       vg,
@@ -662,6 +665,7 @@ describe("Recovery", async function () {
       30_000_000,
     );
 
+    // advance time by safetyDelaySeconds + 1 to ensure the signature has expired
     await fx.advanceTimeBy(safetyDelaySeconds + 1);
 
     const bundleWithExpiredTimestamp = wallet1.sign({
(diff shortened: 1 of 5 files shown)


contracts/test/recovery-test.ts
```diff
@@ -25,6 +29,7 @@ const signWalletAddress = async (
 
 describe("Recovery", async function () {
   const safetyDelaySeconds = 7 * 24 * 60 * 60;
+  const signatureExpiryWindow = 1200;
```
JohnGuilding: Chose this arbitrarily. Is equal to 100 12 second blocks.

Was a bit torn on whether this was a good variable name. Suggestions welcome :)

JohnGuilding:
CI seems to be failing due to some typescript issues in the client module. Doesn't seem related to this PR - got the same failures running clients tests in latest from `contract-updates`. Can look into this if needed but leaving for now.

contracts/test/recovery-test.ts
jzaki: It looks to be used as an offset, so `signatureExpiryOffsetSeconds`? (I like putting units in where meaningful)

contracts/test/recovery-test.ts
```diff
@@ -581,4 +627,116 @@ describe("Recovery", async function () {
...
+    await fx.call(
+      wallet1,
+      vg,
+      "setBLSKeyForWallet",
+      [addressSignature, wallet2.PublicKey(), invalidSignatureExpiryTimestamp],
+      1,
+      30_000_000,
+    );
+
+    await fx.advanceTimeBy(safetyDelaySeconds + 1);
```
jzaki, line 669: Minor: a couple of comments around the expiry timestamp, and expectations that the first `set..` will succeed.

jzaki (approved):

contracts/test/recovery-test.ts
JohnGuilding: Like that - agree the units make sense here. Will update to `signatureExpiryOffsetSeconds`

contracts/test/recovery-test.ts
JohnGuilding: Added

PR #580, opened 2023-04-19
551 run quill against geth node and pay aggregator fees

### LINKED ISSUE #551: Ensure Quill runs against Geth node and pays aggregator fees
Quill has not been updated to work with some recent changes to `bls-wallet`. This issue is concerned with updating Quill to ensure it runs against a geth node (as opposed to a hardhat node), and that it compensates the aggregator for transaction fees.

## Geth node
- Test running against geth node and document errors here if any exist
- Ensure Quill runs against geth node

## Pay aggregator fees
Quill does not yet pay the aggregator fees, we should enforce paying fees by default. Working examples of paying fees can be found in the aggregator tests and the bls provider integration tests.

JohnGuilding:
## What is this PR doing?
1. Changes the chainId of local Quill network config to `0x539` (1337), from `0x7a69` (31337).
2. Adds logic so that Quill pays the aggregator fees

- Send a transaction locally with Quill, using `yarn start`, instead of `yarn start-hardhat`
- You may need to change you MetaMask localhost chainId to 1337.

resolves #551 

## Checklist

## Guidelines

### DIFF
--- a/contracts/hardhat.config.ts
+++ b/contracts/hardhat.config.ts
@@ -131,6 +131,7 @@ const config: HardhatUserConfig = {
   },
   networks: {
     hardhat: {
+      chainId: 1337,
       initialBaseFeePerGas: 0, // workaround from https://github.com/sc-forks/solidity-coverage/issues/652#issuecomment-896330136 . Remove when that issue is closed.
       accounts,
       blockGasLimit: 30_000_000,
--- a/docs/local_development.md
+++ b/docs/local_development.md
@@ -109,6 +109,7 @@ yarn run dev:chrome # or dev:firefox, dev:opera
 - In general, the bundle or submission issues we've encountered have been us misconfiguring the data in the bundle or not configuring the aggregator properly.
 - Be careful using Hardhat accounts 0 and 1 in your code when running a local aggregator. This is because the local aggregator config uses the same key pairs as Hardhat accounts 0 and 1 by default. You can get around this by not using accounts 0 and 1 elsewhere, or changing the default accounts that the aggregator uses locally.
 - When packages are updated in the aggregator, you'll need to reload the deno cache as the setup script won't do this for you. You can do this with `deno cache -r deps.ts` in the `./aggregator` directory.
+- If running Quill against a local hardhat node with (`yarn start-hardhat`) instead of the local geth node (`yarn start`), you'll need to update the local network in `./extension/config.json` to use chainId `31337`. This is because the geth node uses a chainId of `1337` instead. If you're using MetaMask to fund Quill, make sure the localhost network uses chainId `31337` too.
 
 ### Tests
 
--- a/docs/local_development.md
+++ b/docs/local_development.md
@@ -109,7 +109,8 @@ yarn run dev:chrome # or dev:firefox, dev:opera
 - In general, the bundle or submission issues we've encountered have been us misconfiguring the data in the bundle or not configuring the aggregator properly.
 - Be careful using Hardhat accounts 0 and 1 in your code when running a local aggregator. This is because the local aggregator config uses the same key pairs as Hardhat accounts 0 and 1 by default. You can get around this by not using accounts 0 and 1 elsewhere, or changing the default accounts that the aggregator uses locally.
 - When packages are updated in the aggregator, you'll need to reload the deno cache as the setup script won't do this for you. You can do this with `deno cache -r deps.ts` in the `./aggregator` directory.
-- If running Quill against a local hardhat node with (`yarn start-hardhat`) instead of the local geth node (`yarn start`), you'll need to update the local network in `./extension/config.json` to use chainId `31337`. This is because the geth node uses a chainId of `1337` instead. If you're using MetaMask to fund Quill, make sure the localhost network uses chainId `31337` too.
+- If running Quill against a local node, and if you're using MetaMask to fund Quill, make sure the MetaMask 
+localhost network  uses chainId `1337`.
 
 ### Tests
 
--- a/docs/local_development.md
+++ b/docs/local_development.md
@@ -100,6 +100,7 @@ yarn run dev:chrome # or dev:firefox, dev:opera
 
 - pull latest from `main` and run the setup script from the root directory `./setup.ts`.
 - Restart the node and redeploy contracts
+- Delete `aggregator.sqlite` in `./aggregator`. This is the local DB which will get regenerated when the aggregator is started.
 - Restart the aggregator and add the "-r" flag to the command e.g `./programs/aggregator.ts -r`.
 - Reset the Quill extension in your browser if you're developing with Quill. You can do this by removing the extension and then re-adding via "Load unpacked" again. Or run `debug.reset();` twice in the background page console.
 
@@ -107,6 +108,7 @@ yarn run dev:chrome # or dev:firefox, dev:opera
 
 - In general, the bundle or submission issues we've encountered have been us misconfiguring the data in the bundle or not configuring the aggregator properly.
 - Be careful using Hardhat accounts 0 and 1 in your code when running a local aggregator. This is because the local aggregator config uses the same key pairs as Hardhat accounts 0 and 1 by default. You can get around this by not using accounts 0 and 1 elsewhere, or changing the default accounts that the aggregator uses locally.
+- When packages are updated in the aggregator, you'll need to reload the deno cache as the setup script won't do this for you. You can do this with `deno cache -r deps.ts` in the `./aggregator` directory.
 
 ### Tests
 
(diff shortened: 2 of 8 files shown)


docs/local_development.md
```diff
@@ -100,13 +100,16 @@ yarn run dev:chrome # or dev:firefox, dev:opera
...
 
 ### Additional troubleshooting tips
 
 - In general, the bundle or submission issues we've encountered have been us misconfiguring the data in the bundle or not configuring the aggregator properly.
 - Be careful using Hardhat accounts 0 and 1 in your code when running a local aggregator. This is because the local aggregator config uses the same key pairs as Hardhat accounts 0 and 1 by default. You can get around this by not using accounts 0 and 1 elsewhere, or changing the default accounts that the aggregator uses locally.
+- When packages are updated in the aggregator, you'll need to reload the deno cache as the setup script won't do this for you. You can do this with `deno cache -r deps.ts` in the `./aggregator` directory.
+- If running Quill against a local hardhat node with (`yarn start-hardhat`) instead of the local geth node (`yarn start`), you'll need to update the local network in `./extension/config.json` to use chainId `31337`. This is because the geth node uses a chainId of `1337` instead. If you're using MetaMask to fund Quill, make sure the localhost network uses chainId `31337` too.
```
JohnGuilding: Instead of making the user manually change the chain id, another option could be to add an explicit hardhat local network and geth local network in the extension config files. Seen a similar pattern with wagmi where they offer config for different local chains (localhost, hardhat, foundry etc.)

docs/local_development.md
```diff
@@ -100,13 +100,16 @@ yarn run dev:chrome # or dev:firefox, dev:opera
...
+- Delete `aggregator.sqlite` in `./aggregator`. This is the local DB which will get regenerated when the aggregator is started.
 - Restart the aggregator and add the "-r" flag to the command e.g `./programs/aggregator.ts -r`.
 - Reset the Quill extension in your browser if you're developing with Quill. You can do this by removing the extension and then re-adding via "Load unpacked" again. Or run `debug.reset();` twice in the background page console.
 
 ### Additional troubleshooting tips
 
 - In general, the bundle or submission issues we've encountered have been us misconfiguring the data in the bundle or not configuring the aggregator properly.
 - Be careful using Hardhat accounts 0 and 1 in your code when running a local aggregator. This is because the local aggregator config uses the same key pairs as Hardhat accounts 0 and 1 by default. You can get around this by not using accounts 0 and 1 elsewhere, or changing the default accounts that the aggregator uses locally.
+- When packages are updated in the aggregator, you'll need to reload the deno cache as the setup script won't do this for you. You can do this with `deno cache -r deps.ts` in the `./aggregator` directory.
```
JohnGuilding, line 111: > you'll need to reload the deno cache as the setup script won't do this for you

I think this statement is correct (reloading the cache being a necessity when packages have changed, rather than anything clever happening), correct me if anyone thinks that's wrong.

extension/source/background/AggregatorController.ts
```diff
@@ -158,4 +175,55 @@ export default class AggregatorController {
...
+    ];
+
+    const feeEstimateBundle = wallet.sign({
+      nonce,
+      actions: [...actionsWithFeePaymentAction],
+    });
+
+    const feeEstimate = await aggregator.estimateFee(feeEstimateBundle);
+    const feeRequired = BigNumber.from(feeEstimate.feeRequired);
+    const safetyPremium = feeRequired.div(5);
```
blakecduncan, line 214: Do you think there would be a reason to change this `safetyPremium` semi frequently?  Just wondering if we should make it an env variable

extension/source/background/AggregatorController.ts
```diff
@@ -158,4 +175,55 @@ export default class AggregatorController {
...
+    const aggregatorUtilitiesAddress = netCfg.addresses.utilities;
+    const aggregatorUtilitiesContract = AggregatorUtilitiesFactory.connect(
+      aggregatorUtilitiesAddress,
+      ethersProvider,
+    );
+
+    const actionsWithFeePaymentAction = [
+      ...actions,
+      {
+        ethValue: 1,
```
blakecduncan, line 198: Possibly a dumb question

Why is this action needed? And why do we set ethValue to 1?

blakecduncan: Oh nvm.  Is this added here just to get an accurate fee estimate?

blakecduncan (approved):
lgtm!  Just one small question

contracts/hardhat.config.ts
```diff
@@ -131,6 +131,7 @@ const config: HardhatUserConfig = {
   },
   networks: {
     hardhat: {
+      chainId: 1337,
```
JohnGuilding, line 134: Decided to enforce hardhat using the same chainId as the geth node. This means we don't have to change config when moving between the two. e.g. If you wanted to run quill against the hardhat node, you'd have to change the config file to use a different chainId.

Deemed changing hardhat chainId over changing the geth chainId mainly because it was easier. `--dev` mode for geth is [not meant to be configured](https://github.com/ethereum/go-ethereum/issues/21120#issuecomment-633646596). If we wanted to change the chainId, looks like we'd have to run a [private network with a custom genesis file](https://ethereum.stackexchange.com/questions/25359/geth-chain-option) which seems like a lot more effort than necessary.

**Edit:**
Another option could be to add an explicit hardhat local network and geth local network in the extension config files. Seen a similar pattern with wagmi where they offer config for different local chains (localhost, hardhat, foundry etc.). This option also seems like more work than is needed compared to just changing the hardhat chainId

extension/source/background/AggregatorController.ts
JohnGuilding: Yeah I think that's a good question. I originally took the value of 5 from [an aggregator test](https://github.com/web3well/bls-wallet/blob/70ca00f0894b301c24692e3c5aa13efebd69c9e1/aggregator/test/AggregationStrategy.test.ts#L76), and used that in the provider. I'm not sure on the answer to the question tbh, and would be interested to get Andrews thoughts on this. I think another useful question would be is 20% to high as a safety margin? I've noticed the new `signWithGasEstimate` method in the contract-updates has a default value of 0, so after seeing that, a margin of 20% here seems very high.

@voltrevo do you see the need for a value like this to be changed ever?  

And is a 20% margin even sensible in this case?

extension/source/background/AggregatorController.ts
JohnGuilding: > why do we set ethValue to 1?

Not a dumb question at all, I had the exact same question before. I had a discussion with @voltrevo about this and it's possible to provide a value of 0 wei, or 1 wei for the `ethValue` when estimating fees. If you follow this thread from [this comment onwards](https://github.com/web3well/bls-wallet/pull/535#discussion_r1131800261), that should provide some useful context.

My understanding is that you need to add the payment action to the estimate fee bundle in order to prevent the [actual transaction costing more than the estimate](https://github.com/web3well/bls-wallet/blob/55ad789c3a86831d4bbadff92a402ae1d017f96c/aggregator/README.md?plain=1#L107-L110). Then it is a choice whether the `ethValue` is 0 or 1. The discussion I've linked above contains some of the tradeoffs for each option.

extension/source/background/AggregatorController.ts
JohnGuilding: Decided to merge btw as wanted to add this change to `merge-in-contract-updates`, think this change would be useful to test quill with, as we want to enforce paying fees as the default behaviour. 

Can add follow ups if needed

voltrevo: Yeah 20% is quite high. I think it's fine for now so that we err on the side of reliability, but we'll definitely want to scrutinize this once compression delivers competitive pricing.

Some thoughts off the top of my head:
- Add this to user configuration
- Aggregator could estimate how much each account has over-paid and adjust accordingly

docs/local_development.md
voltrevo: Nah it's not quite right. The cache logic only says it's ok to use a previous fetch of the same url. If the url is different, it'll refetch it. No cache clearing needed.

The problem is that a lot of urls in the dependency tree have semver-matching urls (e.g. `@^1.0.0`) instead of being pinned (e.g. `@1.0.0`). This can lead to some dependencies being more up-to-date than others, causing conflicts. There's also other ways urls can resolve to different things over time, like updates to esm.sh itself.

I wouldn't go into detail in the docs, instead just something like:

```
- Sometimes there are issues related to the deno cache. You can clear it with `deno cache -r deps.ts test/deps.ts` in the `./aggregator` directory.
```

extension/source/background/AggregatorController.ts
JohnGuilding: Will add the safetyPremium to extension config.

> Aggregator could estimate how much each account has over-paid and adjust accordingly

This sounds useful, worth creating an issue? Or just to be mindful of this when we circle back following compression work?

docs/local_development.md
JohnGuilding: Forgot to reply to this. I see, that makes sense, thanks for pointing this out. Will update in the current work I'm doing

PR #584, opened 2023-04-24
Use compression in aggregator

### LINKED ISSUE #406: Make use of expander contract in aggregator

voltrevo:
# *Dependent PR*

This PR depends on https://github.com/web3well/bls-wallet/pull/586.

After that's merged, toggle the target branch back and forth or push an empty commit to fix the diff. (Preview of [this PR's changes](https://github.com/web3well/bls-wallet/compare/contracts-connector...bw-406-aggregator-compression).)

(If you want, you can also just review+merge this one directly. Github will automerge the previous PRs.)

## What is this PR doing?

Uses compression in aggregator. Reduces the `data` field when sending a single ETH transfer from 836 bytes down to 164 bytes.

(Replicating this test isn't required.)

Use Quill to do an ETH transfer and verify the transaction data is about 324 bytes. Here's what I got on arbitrum goerli:

Screen Shot 2023-04-24 at 3 33 07 pm

[Explorer link](https://goerli.arbiscan.io/tx/0xb5936256c8ee83d4ad58e1f565534cf8f8f731a88b7d5f1d2a3e5da14ac0d35d)

To reduce this further, use `./manualTests/registerWallet.ts` to register the source BLS key and address, as well as the destination address. Verify that reduces the transaction data down to about 164 bytes:

Screen Shot 2023-04-24 at 3 33 17 pm

[Explorer link](https://goerli.arbiscan.io/tx/0x9c24c938c16b08ee16d0b19d82ca711630c7b7781f20961435c1943e8e135eb5)

For comparison I used version `f8a8c490` (`git checkout f8a8c490`) (this is right before the `contract-updates` merge, which naturally changes the contracts so `main` currently doesn't work with the existing deployment). Here's the transaction data from that one:

Screen Shot 2023-04-24 at 3 55 20 pm

[Explorer link](https://goerli.arbiscan.io/tx/0xd4d2bdc98beb7012c6b0b4789301d87a134ecdbe1e20ffba0f657cdda093c52d)

## Wait, wait, still 164 bytes? 😰

Yeah. That's just the data field too. If you include the full transaction data (including the envelope with the aggregator's signature and gas settings etc), then there's another 115 bytes or so (total is about 279 bytes).

Anyway, 164 bytes in the data field is a lot more than we need, and it's because there's still some cleanup to do. Here's the breakdown of those bytes:

```
0x

3a276523
  - method id of run()

0000000000000000000000000000000000000000000000000000000000000020
  - location of byte array for `stream` calldata argument

0000000000000000000000000000000000000000000000000000000000000056
  - `stream` is 86 bytes long
    (==0x56 but solidity abi uses an entire uint256 word for this)

01
  - one operation

00
  - use expander id 0 (fallback expander)

07
  - 0x07 = 0b111 - bit stream:
    - 1: Use registry for sendWallet's public key
    - 1: Include a tx.origin payment
    - 1: Use registry for recvWallet's address

000000
  - id of sender's bls public key

01
  - nonce

0cb730
  - 56,708 gas

02
  - two actions (send ETH and pay tx.origin)

7900
  - 0.0001 ETH

000001
  - id recipient's address

00
  - zero bytes for encodedFunction

3092cbb31e
  - pay 0.0000311872752 ETH to tx.origin

0b018b1989c5c62d43b0a0427f04781f2a4e201ab2c938ad23aa2d0a6c0a284a
03f37bd9fbdac1b34f3046363244c0e47a2a73add7cdfac41c2f134285256dfc
  - BLS signature

00000000000000000000
  - padding bytes from the solidity abi to make a multiple of 32 bytes
```

In summary:
- 78 bytes of ABI overhead
- 1 byte to say that there's one operation
- 21 bytes for the actual operation (send ETH to recipient and pay tx.origin)
- 64 bytes for the signature

We should be able to remove the ABI overhead by introducing another contract whose whole purpose is to call the expander with its raw argument.

Focusing on the positive, very small encodings for the actual operations *have* been achieved, and this means that the *marginal* data cost of each operation is this low number (20 bytes or so). This means you could add 100 transactions for 2100 bytes, so the whole transaction would be 2100+279=2379 bytes, or about 24 bytes per transaction. Then the marginal cost becomes dominant, BUT you need to wait for 100 transactions before you can submit.

Related improvements to work on:
- Fix those 78 bytes of overhead
- Round up to 3 significant figures for gas limit and tx.origin payment (would bring 21 bytes above down to 17)
- Add registration functionality to Quill

Resolves #406.

## Checklist

## Guidelines

### DIFF
--- a/aggregator/manualTests/registerWallet.ts
+++ b/aggregator/manualTests/registerWallet.ts
@@ -1,0 +1,34 @@
+#!/usr/bin/env -S deno run --allow-net --allow-env --allow-read
+
+import { ContractsConnector, ethers } from "../deps.ts";
+import * as env from "../src/env.ts";
+import AdminWallet from "../src/chain/AdminWallet.ts";
+import receiptOf from "./helpers/receiptOf.ts";
+
+const provider = new ethers.providers.JsonRpcProvider(env.RPC_URL);
+
+const adminWallet = AdminWallet(provider);
+
+const connector = await ContractsConnector.create(adminWallet);
+
+const addressRegistry = await connector.AddressRegistry();
+const blsPublicKeyRegistry = await connector.BLSPublicKeyRegistry();
+
+await receiptOf(
+  addressRegistry.register("0xCB1ca1e8DF1055636d7D07c3099c9de3c65CAAB4"),
+);
+
+await receiptOf(
+  blsPublicKeyRegistry.register(
+    // You can get this in Quill by running this in the console of the wallet
+    // page (the page you get by clicking on the extension icon)
+    // JSON.stringify(debug.wallets[0].blsWalletSigner.getPublicKey())
+
+    [
+      "0x0ad7e63a4bbfdad440beda1fe7fdfb77a59f2a6d991700c6cf4c3654a52389a9",
+      "0x0adaa93bdfda0f6b259a80c1af7ccf3451c35c1e175483927a8052bdbf59f801",
+      "0x1f56aa1bb1419c741f0a474e51f33da0ffc81ea870e2e2c440db72539a9efb9e",
+      "0x2f1f7e5d586d6ca5de3c8c198c3be3b998a2b6df7ee8a367a1e58f8b36fd524d",
+    ],
+  ),
+);
--- a/aggregator/src/app/EthereumService.ts
+++ b/aggregator/src/app/EthereumService.ts
@@ -312,6 +312,17 @@ export default class EthereumService {
 
     const compressedBundle = await this.bundleCompressor.compress(bundle);
 
+    const [rawTx, rawCompressedTx] = await Promise.all([
+      this.verificationGateway.populateTransaction.processBundle(bundle).then(
+        (tx) => this.wallet.signTransaction(tx),
+      ),
+      this.bundleCompressor.blsExpanderDelegator.populateTransaction
+        .run(compressedBundle).then((tx) => this.wallet.signTransaction(tx)),
+    ]);
+
+    const txLen = ethers.utils.hexDataLength(rawTx);
+    const compressedTxLen = ethers.utils.hexDataLength(rawCompressedTx);
+
     const processBundleArgs: Parameters<
       (typeof this.bundleCompressor.blsExpanderDelegator)["run"]
     > = [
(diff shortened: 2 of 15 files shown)


aggregator/src/app/EthereumService.ts
```diff
@@ -291,20 +310,34 @@ export default class EthereumService {
...
-        bundle,
-        {
-          nonce: this.NextNonce(),
-          ...await this.GasConfig(),
-        },
-      ];
+    const compressedBundle = await this.bundleCompressor.compress(bundle);
+
+    const [rawTx, rawCompressedTx] = await Promise.all([
+      this.verificationGateway.populateTransaction.processBundle(bundle).then(
```
blakecduncan, line 316: Do we need to populate the transaction with the VG and also with the bundle compressor?  Or do we just do the VG one so that we can log out what the `rawTx` length is?

aggregator/manualTests/registerWallet.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const addressRegistry = await connector.AddressRegistry();
+const blsPublicKeyRegistry = await connector.BLSPublicKeyRegistry();
+
+await receiptOf(
+  addressRegistry.register("0xCB1ca1e8DF1055636d7D07c3099c9de3c65CAAB4"),
+);
+
+await receiptOf(
+  blsPublicKeyRegistry.register(
```
blakecduncan, line 22: Just for my understanding - We register a wallet with the registry because it allows us to compress the transaction even more.  Is this the only way we can currently register a wallet?  Asking just to check if a wallet would get registered on its own after its first transaction.

blakecduncan (approved):
LGTM!

Code all looks good, so gonna merge.  Just had some questions for my clarity

aggregator/src/app/EthereumService.ts
voltrevo: Nah it's only for logging.

aggregator/manualTests/registerWallet.ts
voltrevo: Pretty much yeah. It doesn't happen automatically. We need to do something about that.

Also:

> it allows us to compress the transaction even more

technically correct but I wouldn't phrase it that way. It's not 'even better', it's essential - otherwise your tx will be more expensive, not less.

PR #585, opened 2023-04-24
Fix expander lookup

voltrevo:
## What is this PR doing?

Before this change, I've been assuming that expanders were installed at fixed IDs in the top-level expander (`BLSExpanderDelegator`). That's fine for testing, but exporting to the aggregator I realized it would be bad to rely on that.

Rather than exporting a configuration or scanning the registered expanders to find the appropriate IDs, this change adds an indexed event for expander registration so that those can be used instead. This avoids both problems so that initializing the compressor should be pretty painless (including adding more compressor+expander pairs in future).

`yarn hardhat test`

Contributes to #406.

## Checklist

## Guidelines


blakecduncan (approved):
Nice this make a lot of sense 👍

PR #571, opened 2023-03-29
Audit issue 7 fixes

blakecduncan:
## What is this PR doing?

Issues:
- `BLS_DOMAIN` is just an arbitrary string so it doesn't protect against replay across contracts or chains.
- Ex: The message in `safeSetWallet()` is only signed with the wallet address.  
- Ex: In the `verify()` function, the `chainId` is appended to messages but not the verification gateway address.

Solution:
- Include the chainId and the verification gateway address in the domain.
- Create a seperate domain for `safeSetWallet` and `verify` so that it's not possible for overlap between a bundle message and a proof of possession message.

## How can these changes be manually tested?
- Running all tests and checking Quil/example dapps still work

## Does this PR resolve or contribute to any issues?
https://github.com/web3well/bls-wallet/issues/459

## Checklist
- [ ] I have manually tested these changes
- [ ] Post a link to the PR in the group chat

## Guidelines
- If your PR is not ready, mark it as a draft
- The `resolve conversation` button is for reviewers, not authors
  - (But add a 'done' comment or similar)

PR #588, opened 2023-04-24
Add name and version to bls wallet domains

### LINKED ISSUE #25: Update BLS_DOMAIN to EIP-712

JohnGuilding:
## What is this PR doing?
Adds name and version to BLS Wallet domains. Following offline discussion with @jzaki, we decided that we do not need to implement EIP-712 at the contract level. Instead, the EIP-712 smart contract changes are mainly relevant at the dapp/protocol level, not the contract level for smart contract wallets.

The audit did not specify directly implementing 712, but instead, using it as inspiration and additional information for completing audit issue 7 - #571:

> Other recommendations are to also include a version and a name fields to allow protocol upgrades. Consider using the EIP712 domain as inspiration and additional information.

Although the contracts don't need EIP-712 support, we may want to consider adding support for EIP-712 in Quill in another issue, as it currently does not. 

- run tests

resolves #25 

## Checklist

## Guidelines

### DIFF
--- a/contracts/clients/src/NetworkConfig.ts
+++ b/contracts/clients/src/NetworkConfig.ts
@@ -23,11 +23,15 @@ export type NetworkConfig = {
   auxiliary: {
     chainid: number;
     /**
-     * Domain used for BLS signing
+     * Domain used for signing BLS Proof of Possession messages
      */
-    domain: string;
+    walletDomain: string;
     /**
-     * Starting block contracts began dpeloyment at
+     * Domain used for signing BLS Bundle messages
+     */
+    bundleDomain: string;
+    /**
+     * Starting block contracts began deployment at
      */
     genesisBlock: number;
     /**
@@ -64,7 +68,8 @@ export function validateConfig(cfg: UnvalidatedConfig): NetworkConfig {
     },
     auxiliary: {
       chainid: assertNumber(cfg.auxiliary.chainid),
-      domain: assertString(cfg.auxiliary.domain),
+      walletDomain: assertString(cfg.auxiliary.walletDomain),
+      bundleDomain: assertString(cfg.auxiliary.bundleDomain),
       genesisBlock: assertNumber(cfg.auxiliary.genesisBlock),
       deployedBy: assertString(cfg.auxiliary.deployedBy),
       version: assertString(cfg.auxiliary.version),
--- a/contracts/clients/src/signer/getDomain.ts
+++ b/contracts/clients/src/signer/getDomain.ts
@@ -2,13 +2,15 @@ import { arrayify, solidityPack } from "ethers/lib/utils";
 import { utils } from "ethers";
 
 export default (
+  name: string,
+  version: string,
   chainId: number,
   verificationGatewayAddress: string,
   type: string,
 ): Uint8Array => {
   const encoded = solidityPack(
-    ["uint256", "address", "string"],
-    [chainId, verificationGatewayAddress, type],
+    ["string", "string", "uint256", "address", "string"],
+    [name, version, chainId, verificationGatewayAddress, type],
   );
 
   return arrayify(utils.keccak256(encoded));
--- a/contracts/clients/src/signer/index.ts
+++ b/contracts/clients/src/signer/index.ts
@@ -32,8 +32,23 @@ export async function initBlsWalletSigner({
   // properly initialized for all use cases, not just signing.
   const signerFactory = await signer.BlsSignerFactory.new();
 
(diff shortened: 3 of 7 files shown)


blakecduncan (approved):
I think the changes look good 💯 

I was a little confused by the description though, I don't think I understand the comment - `we decided we do not need to implement EIP-712 at the contract level`.  What would need to change to implement 712 at the contract level and why are we then adding a name and version into the domain?  

Are these changes so that Quill can show the name and version to the user before they sign a message?

jzaki (approved):

jzaki:
> I was a little confused by the description though, I don't think I understand the comment - `we decided we do not need to implement EIP-712 at the contract level`. What would need to change to implement 712 at the contract level and why are we then adding a name and version into the domain?
> 
> Are these changes so that Quill can show the name and version to the user before they sign a message?

The domain does not conform to 712, as it includes "Bundle" and "Wallet" as an extra param. The name and version helps separate domains of version numbers and wallet signer name. So if the gateway contract was updated in situ (it can't be), then the version in the domain would separate messages signed for the previous version to be valid in the new version.
The scope of 712 domains makes sense for domain separation at the dapp level, and for legibility as you suggest. The BLS domain is effectively at a lower level.

blakecduncan:
> 

@jzaki Nice that makes sense 👍.  I was confused because I was curious what the benefit of adding a name and version if we aren't making it 712 compliant.  As the only way to change the VG is to deploy a new one which would change VG address and that would change the domain.  But it sounds like legibility is the main advantage here.

JohnGuilding:
> I was a little confused by the description though, I don't think I understand the comment - we decided we do not need to implement EIP-712 at the contract level. What would need to change to implement 712 at the contract level and why are we then adding a name and version into the domain?

@blakecduncan yeah re-reading I can see how the description is confusing. To build on what James has said, we chose to take inspiration from EIP-712 (for the reasons James mentioned), but we didn't actually implement it. This is because it isn't actually needed at the contract level for a smart contract wallet. 

It would be used in a dapp smart contract as a [domainSeparator](https://eips.ethereum.org/EIPS/eip-712#definition-of-domainseparator), and would be specific to that smart contract. Here's an example from [a uniswap contract](https://github.com/Uniswap/v2-periphery/blob/dda62473e2da448bc9cb8f4514dadda4aeede5f4/contracts/test/DeflatingERC20.sol#L23-L36) of how it could be used. As well as being able to display such information to the user in the wallet, the domain in this case is used to ensure that signed typed data cannot be confused between different protocols/dapps. E.g. there's a possibility different dex's may have the same signed message struct. So the domain separator helps differentiate between such transactions and prevents signature collision.

blakecduncan:
> > I was a little confused by the description though, I don't think I understand the comment - we decided we do not need to implement EIP-712 at the contract level. What would need to change to implement 712 at the contract level and why are we then adding a name and version into the domain?
> 
> @blakecduncan yeah re-reading I can see how the description is confusing. To build on what James has said, we chose to take inspiration from EIP-712 (for the reasons James mentioned), but we didn't actually implement it. This is because it isn't actually needed at the contract level for a smart contract wallet.
> 
> It would be used in a dapp smart contract as a [domainSeparator](https://eips.ethereum.org/EIPS/eip-712#definition-of-domainseparator), and would be specific to that smart contract. Here's an example from [a uniswap contract](https://github.com/Uniswap/v2-periphery/blob/dda62473e2da448bc9cb8f4514dadda4aeede5f4/contracts/test/DeflatingERC20.sol#L23-L36) of how it could be used. As well as being able to display such information to the user in the wallet, the domain in this case is used to ensure that signed typed data cannot be confused between different protocols/dapps. E.g. there's a possibility different dex's may have the same signed message struct. So the domain separator helps differentiate between such transactions and prevents signature collision.

@JohnGuilding Yeah that makes sense.  I think my main point is that we already have a domain separator, it just isn't in the EIP-712 format.  

When I added the domains to address audit issue 7 I decided not to include a name and version because it wasn't needed to address the audit issue and it was added scope to fully implement eip712.  So I think to rephrase my original question, I was curious what the benefit of adding a name and version is if we aren't doing it for EIP-712 compatibility (since we already prevent signature collision).  It sounds like we do get legibility benefits for dApps though

JohnGuilding:
@blakecduncan ok gotcha, I see where your question was coming from now 👍

PR #592, opened 2023-04-28
192 generate determinisitc bundle hashes

### LINKED ISSUE #192: Add ability for bls-wallet-clients to generate a determinisitc hash of a bundle
A la ETH transaction hashes (https://stackoverflow.com/a/57638023), we should add a bundle hashing function to `bls-wallet-clients` so that the hash for a bundle can be determined at will.
- Add bundle hashing function to `bls-wallet-clients`. There may be a general ETH/MetaTransaction spec for this that we should adhere to.
- Replace existing `makeHash` function in aggregator with this.

JohnGuilding:
# What is this PR doing?
Adds functions in aggregator and client module to generate a deterministic hash.

Also ability to look up the bundles in the submitted bundle via a bundle hash

## What characteristics does the deterministic hash have?

The deterministic hash has these characteristics
- Two bundles that are the same generate the same hash
- Two different bundles generate different hashes
- The same bundle produces a different hash on different chains (took this characteristic from 4337 implementation)
- Users and consuming applications can generate the same hash an aggregator would, without needing to connect to one

## How does 4337 hash user operations?
The userOpHash is a hash over the userOp (except signature), entryPoint and chainId. See code [here](https://github.com/eth-infinitism/bundler/blob/ef0751cdcf8e865c2550435c3046a6a0b54ef52c/packages/utils/src/ERC4337Utils.ts#L115-L121) (Note: the typescript bundler implementation simply sets the signature to zero). The 4337 implementation encodes the user operation with a packUserOp function, I followed a similar approach for this PR.

Couldn't find a 4337 reference implementation for hashing aggregated users ops.

## What is included in the bundle hash?
- The hash function for this PR maps over individual sub bundles and sets the signature to zero. 4337 also sets the signature to zero.
- When mapping over the operations, also decided to set the `gas` property to zero, as feel like that could change between bundles.
- The bundle is then hashed, encoded with a chainId, and then hashed again

## Potential gotcha
One issue developers may run into, is they may create a bundle with say an `Action` with a BigNumber `value` property. If this `value` was converted to a hexstring when submitted by aggregator, this would output a different hash. The ambiguity of the type `BigNumberish` & `BytesLike`, and lack of validation means this may confuse developers when trying to use the client module helper method. I ran into this issue when adding the integration test for this PR. I was able to overcome it by using `.toHexString()`.

1. Submit multiple bundles that will get aggregated via `./manualTests/mintNViaAggregator.ts 3 `
2. Get aggregate bundle via postman - GET http://localhost:3000/aggregateBundle/{subBundleHash}

resolves #192 

## Checklist

## Guidelines


aggregator/src/app/BundleService.ts
```diff
@@ -273,6 +326,13 @@ export default class BundleService {
         },
       });
 
+      if (aggregateBundle) {
```
JohnGuilding, line 325: I tried a few different places in the code to set the `aggregateBundleHash` and settled on here. Wanted to update it as close to the submission as possible, while still having to do the update once so wouldn't have to set it in `handleFailedRow` and `submitAggregateBundle`.

JohnGuilding: hmmm after writing that, now questioning whether this is the best place. Do we even need to update the `aggregateBundleHash` if the bundle is going to fail? In which case, does just before the line: `await this.submitAggregateBundle(` work better?

aggregator/src/app/BundleRouter.ts
```diff
@@ -39,5 +39,19 @@ export default function BundleRouter(bundleService: BundleService) {
     },
   );
 
+  router.get(
+    "aggregateBundle/:subBundleHash",
```
JohnGuilding, line 43: Overview of naming in this PR:

Open to any further suggestions :)

`aggregateBundle`: refers to the collection of bundles that were submitted with the bundle hash provided (subBundleHash). Wondering if the naming is confusing as it isn't actually an "aggregated bundle", or a bundle that was aggregated, it is a collection of bundles that were submitted along with the bundle from the hash provided (hope that makes sense).

`aggregateHash`: the hash given to the aggregateBundle that get's submitted on chain. Bit torn on the naming here too, as you can't actually look up the aggregate bundle with this hash as the aggregate bundle is not saved in the db, only the sub bundles (to my knowledge). It is used to group other sub bundles together, see BundleTable.ts.

`subBundle`: refers to the bundles saved in the DB that were submitted by a user/application. Not the aggregate bundle. `subBundleHash` is the bundle hash for said bundle.

contracts/clients/src/helpers/hashBundle.ts
```diff
@@ -0,0 +1,50 @@
+import { BigNumberish, ethers } from "ethers";
+import { Bundle, Operation } from "../signer";
+
+type BundleWithoutSignature = {
+  senderPublicKeys: [BigNumberish, BigNumberish, BigNumberish, BigNumberish];
+  operations: Omit<Operation, "gas">;
+};
+
+export function hashBundle(bundle: Bundle, chainId: number): string {
+  if (bundle.operations.length !== bundle.senderPublicKeys.length) {
```
JohnGuilding, line 18: Thought the room for incorrect arguments was larger in a client module function. Added this basic check but held off on adding anymore, like validating more of the bundle, or chain id etc. Let me know if anyone thinks otherwise

contracts/clients/src/helpers/hashBundle.ts
```diff
@@ -0,0 +1,58 @@
...
+    (entry) => "name" in entry && entry.name === verifyMethodName,
+  )?.inputs[0];
+
+  const validatedBundle = {
+    ...bundle,
+    operations: operationsWithZeroGas,
+  };
+
+  const encodedBundleWithZeroSignature = ethers.utils.defaultAbiCoder.encode(
+    [bundleType as any],
```
JohnGuilding, line 42: I did originally just serialise the bundle with JSON.stringify (as opposed to encoding the bundle with a type), I figured the encoding approach was better for preserving type information. Not sure how important that is here. From my understanding, encoding will lead to a more consistent representation of the data, and be more efficient than JSON.stringify

aggregator/src/app/BundleService.ts
jacque006: If the original bundle fails to be included, then it would just stay in the mempool. If a submitted bundle/userop fails, then it stays as part of the agg bundle in a failed state since it was run. The end user/app would need to re-submit.

aggregator/src/app/BundleRouter.ts
jacque006: Given how things work now I think this is fine. Ideally, we would just have one endpoint you could provide a hash to, and it would give you the one or more userops associated with that, maybe status.

I saw in PR desc that you weren't able to find implementation of ERC-4337 way of doing it. They may not have gotten that far...

jacque006 (approved):
I think we can streamline/make this more consistent in v2. But this works well given the limitations/setup of v1. :+1: 

Should just need to resolve merge conflicts and this can go :shipit:

PR #600, opened 2023-05-04
Add ability to blsWalletSigner to verify multiple addresses

### LINKED ISSUE #392: Bug: blsWalletSigner.verify cannot verify bundles with multiple addresses
`verify(bundle, address)` does not make sense because each operation within the bundle has its own address. This function needs to take an array of addresses instead:
https://github.com/web3well/bls-wallet/blob/bc3d146/contracts/clients/src/signer/verify.ts#L9

This is already causing problematic usage in the aggregator:
https://github.com/web3well/bls-wallet/blob/bc3d146/aggregator/src/app/BundleService.ts#L164

Here we are verifying that the bundle is separately valid for every address, which doesn't make any sense. This can only work if every address is the same.

Admittedly, most normal usage of `verify` is to check the validity of an individual `BlsWalletWrapper`'s `.sign` output. However, verifying a bundle that has been aggregated across `.sign`s from different wallets is a totally valid use-case, so it's important to not bake this assumption into our API.

Example use-cases:
- Debugging/devx: `verify` should be able to verify any valid bundle.
- Atomic txs across addresses: Before `address` was added to `verify`, it was possible to send a bundle to the aggregator with operations from multiple addresses, which forces those operations to be atomic (although sandwiching is possible). This is would still be possible by switching to sending an array of addresses to `verify`.
- devx: I want to understand BLSWallet signatures and the signature of `verify` suggests bundles are associated with a single address, which is wrong. This inhibits understanding and therefore trust/adoption/etc.

JohnGuilding:
## What is this PR doing?
1. Fixes the `blsWalletSigner.verify` bug which means it cannot verify bundles with more than one address 
2. Updates the BundleService to use the updated API

- Run new tests

Resolves #392 

## Checklist

## Guidelines

### DIFF
--- a/aggregator-proxy/package.json
+++ b/aggregator-proxy/package.json
@@ -21,7 +21,7 @@
     "@types/koa__cors": "^3.3.0",
     "@types/koa__router": "^8.0.11",
     "@types/node-fetch": "^2.6.1",
-    "bls-wallet-clients": "0.9.0",
+    "bls-wallet-clients": "0.8.3-a9c0b22",
     "fp-ts": "^2.12.1",
     "io-ts": "^2.2.16",
     "io-ts-reporters": "^2.0.1",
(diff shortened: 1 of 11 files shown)


blakecduncan (approved):
lgtm!

